PR #31, opened 2024-04-22
`ilab` engine proposal

cdoern:
This enhancement introduced a new design for `ilab`. Primarily adding sub-parent commands, new modes of interactions with `ilab`, and clarity on what the source + sink of data are in this system we are building


russellb:
@cdoern This PR will probably have to be re-posted from a public fork. Apologies for the disruption.

PR #48, opened 2024-05-14
ilab model engine proposal

cdoern:
This enhancement introduces a new design for `ilab`. Primarily adding sub-parent commands, new modes of interactions with `ilab`, and clarity on what the source + sink of data are in this system we are building.


docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+
+The starred commands are new, and will be dealt with in another enhancement.
+
+#### Goals for End of May release in CLI
+
+For the next few milestones, it has been identified that the overall structure shift should be in place. We should aim to have the vague hierarchy of commands in place that will exist for the future of `ilab` so that when users become familiar with the CLI, it does not change radically soon after launch.
+
+The primary focus here should be the `ilab model` group. The other ones should not be treated as blockers for launch, but would be nice to have.
+
+Another key change that should be included is a switch to using `/var/lib/models` as a model store and `$HOME/.config/models/ilab.conf` for the new configuration path. Keeping config and models in the CWD does not map well to the importance of these two pieces. This will allow `ilab config` commands to act on a global level per-user. Models will be a system-wide directory as they are with most other applications of this type.
```
hickeyma, line 114: Do we want global config? The thinking from the start is that a user can create multiple independent project s on their system, similar to OperatorSDK.

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+The mental model a user has impacts the way they use the application and whether they view it as a Proof Of Concept.
+
+Here is a diagram of what I mean when I describe this mental model:
+
+```console
+SOURCE                                                                                                                                                                                                                                              SINK
+
+        User has un-commited changes in their taxonomy                          Data is generated and then the user runs                               A model is created, and the user can tag the model               After converting the model, the
+        or specifies `--skill` `--knowledge` in the                 ==>        `ilab model train` using the newly generated data       ==>             or just list it using the new model commands            ==>      user runs the chat procces by specifying `ilab  model run -i` or `ilab model run` followed by `ilab model chat
+        new `ilab data generate`
```
hickeyma, line 133: not `run`, should be `serve` in the AI domain

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+
+```
+
+So in `ilab` the clear source of information is taxonomy, or the new skills/knowledge brought into generate. This data then goes through processes that end up in the sink, which is a running/actionable process which the user can interact with. Providing commands that clearly display these different process is key.  groups for these larger "buckets" of processes makes sense as well due to the likelihood of more commands being needed to properly interact with the newly generated model.
```
hickeyma, line 137: .groups for --> . Groups for

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+|       |____taxonomy
+|_______convert
+|       |_____model
+|_______train
+|       |_____model
+```
+
+This is opposed to the structure in this EP, which might have a duplicate here and there for something like `download`, but in this structure duplicate commands are the norm and will result in a larger and clunkier codebase that is confusing to read.
+
+This first structure looks nice, and `ilab download model` sounds nice. However in terms of implementation, this makes little sense. in `click`, the cli library we use, `ilab` is a "group" that commands are currently "grouped under". In this Model, `ilab` would be a group, `download`, `convert` etc would be a sub-group, and `model` would be a command under ALL of these groups. This would require an implementation of the `model` function in different packages all linking to a different group parent. From the perspective of user contributions and general code readability doesn't make much sense. Part of this design is to encourage user contributions by making the structure of the codebase logical. While these commands make the commands sound better, they make them harder to group, and understand.
```
hickeyma, line 170: add link to `click`

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+
+currently one has to:
+
+1. `ilab init`
+2. `ilab download`
+3. `ilab serve` 
+4. `ilab generate`
+5. `ilab train`
+7. `ilab convert`
+8. `ilab serve --model XXX`
```
hickeyma, line 190: Like above swap and first state current flow and then the nerw proposed improved flow?

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
+# Redesign ilab to be a Full Fledged Model Engine
+
+This documents describes a new structure for ilab, consisting of sub-commands that act as parent or "group" commands. Currently the only group command is `ilab`. If the models `ilab` produces are meant to be the golden standard for open source AI, the tool to use and manage these models needs to be full fledged and extensible, especially to models we have not produced but should be able to run.
+
+open source container tools like podman an docker are commonly viewed as engines. `ilab` should be a model engine, managing the deployment and creation of AI models. To be clear, I will be using tools like podman as a structural analogy here with full understanding they have different implications and use-cases.
```
hickeyma, line 5: Add links to podman and docker. It makes it easier for users who might not know.

hickeyma (changes requested):
Thanks @cdoern for pushing this design and for you perseverance in multiple redesigns. I like this overall and the direction your are proposing.

Some small nits and feedback to address.

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
+# Redesign ilab to be a Full Fledged Model Engine
+
+This documents describes a new structure for ilab, consisting of sub-commands that act as parent or "group" commands. Currently the only group command is `ilab`. If the models `ilab` produces are meant to be the golden standard for open source AI, the tool to use and manage these models needs to be full fledged and extensible, especially to models we have not produced but should be able to run.
```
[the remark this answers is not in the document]
cdoern: they are children to `ilab` but parents to the further children commands. I will specify this in the ep

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+ilab
+|
+|_______model
+|       |
+|       |____convert
+|       |____download
+|       |____train (--convert)
+|       |____serve   (-i)
+|       |____chat
+|       |____inference
```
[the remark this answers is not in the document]
cdoern: @khaledsulayman @alimaredia and others feel inference makes more sense. I tend to agree since inference is a more "ai" term than test (from what I know)

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+|_____train
+```
+
+The main point of this structure is to introduce a group hierarchy and to remove some confusions of the existing `ilab` structure. Some key things missing in `ilab` currently are:
+
+1. `ilab init` does too many things.
+    - We don't want "black box" commands. This creates bad UX and also ties too many processes together in which one might eventually need to be changed.
+2. ambiguity in "what" we are generating, training etc.
+    - The current structure of `ilab` requires an in depth knowledge of the ordering of commands and background knowledge for how our CLI is built. This is not the best way to grow. Commands that are not intuitive and require a deep dive into the docs just to get started might not engage users as we hope to. having an intuitive structure in which people can see easily from the docstring what each commands does and how they might use them, makes more sense. We are generating data, then training a model, running a server to interact with that model, then finally chatting with the model
+3. some commands like `ilab test` and `ilab list` should be renamed to show their true purpose (`ilab model inference`and `ilab taxonomy diff`)
```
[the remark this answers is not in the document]
cdoern: @khaledsulayman @alimaredia WDYT?

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,204 @@
...
+4. we will hit a point where we want to add more commands to `ilab` due to expected integration with the backend, model registries etc. This list will grow to a point where this form of top group organization is not maintainable.
+
+In general, moving to this structure allows for more growth when `ilab` launches publicly and encourages broader usage of the tool for all sorts of open source AI models.
```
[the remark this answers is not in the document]
cdoern: the public launch is outdated, will replace with something about stability.

in terms of all sorts of models I was just trying to indicate ilab shouldn't be tailored just to granite and merlinite.

docs/ilab-model-engine.md
cdoern: I think global model store and config makes sense.

At least global model store IMO, but if you have that it makes sense to have a global config pointing to this model store.

The way we use config today is really limited, most users override 90% of the config per ilab run, so having duplicates all around the system doesn't rly make sense anymore.

docs/ilab-model-engine.md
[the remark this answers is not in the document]
khaledsulayman: what about `evaluate`? I know this is a very minimal test compared to our full-scale model evaluation but I think it would help to a) differentiate from code testing and b) mirror some of the other subcommands which mostly seem to be more minimal versions of the larger operations we run on the backend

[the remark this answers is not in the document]
khaledsulayman: additionally, I do think, minimal as it currently is, evaluation is _really_ what the current `ilab test` is actually doing, and maybe by being more explicit about the correlation between this and the backend version, we can start to think about what is minimal enough within the backend to incorporate into the CLI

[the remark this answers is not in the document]
cdoern: the only issue with `evaluate` is we might overload that later on with actual eval if we expose from the cli

nathan-weinberg:
@cdoern can you fix the Markdown Lint/Spellcheck errors?

docs/ilab-model-engine.md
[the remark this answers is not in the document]
khaledsulayman: Discussed with Charlie in person but just to update @hickeyma I think that's the point. The CLI is really meant to be a laptop version of the backend pipeline we are building, so having the flow of the CLI mirror it is a plus.

If we feel like the functionality currently handled by `ilab test` is still too minimal to be called evaluation, even for running on a laptop, then we can scope out future work to beef it up as desired.

cdoern:
@nathan-weinberg should be set

docs/ilab-model-engine.md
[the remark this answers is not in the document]
cdoern: switched to evaluate

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
+# Redesign ilab to be a Full Fledged Model Engine
+
+This document describes a new structure for ilab, consisting of sub-commands that act as parent or "group" commands for further sub-commands. Currently the only group command is `ilab`. If the models `ilab` produces are meant to be the golden standard for Open source AI, the tool to use and manage these models needs to be full fledged and extensible, especially to models we have not produced but should be able to run.
```
nathan-weinberg, line 3:
```suggestion
This document describes a new structure for `ilab`, consisting of sub-commands that act as parent or "group" commands for further sub-commands. Currently the only group command is `ilab`. If the models `ilab` produces are meant to be the golden standard for open source AI, the tool to use and manage these models needs to be fully-fledged and extensible, especially for models we have not produced but should be able to run.
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+```console
+ilab
+|
+|_______model
+|       |
+|       |____convert
+|       |____download
+|       |____train (--convert)
+|       |____serve   (-i)
```
nathan-weinberg, line 38:
```suggestion
|       |____serve (-i)
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+|       |____diff
+|       |____check
+|       |____download
+```
+
+The main point of this structure is to introduce a group hierarchy and to improve the usability of the existing `ilab` structure. Some key things missing in `ilab` currently are:
+
+1. `ilab init` does too many things.
+    - We don't want "black box" commands. This creates bad UX and also ties too many processes together in which one might eventually need to be changed.
+2. ambiguity in "what" we are generating, training etc.
```
nathan-weinberg, line 61:
```suggestion
2. Ambiguity in "what" we are generating, training etc.
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+    - The current structure of `ilab` requires an in depth knowledge of the ordering of commands and background knowledge for how our CLI is built. This is not the best way to grow. Commands that are not intuitive and require a deep dive into the docs just to get started might not engage users as we hope to. Having an intuitive structure in which people can see easily from the docstring what each commands does and how they might use them, makes more sense. We are generating data, then training a model, running a server to interact with that model, then finally chatting with the model
+3. Some commands like `ilab test` and `ilab list` should be renamed to show their true purpose (`ilab model evaluate`and `ilab taxonomy diff`)
+4. We will hit a point where we want to add more commands to `ilab` due to expected integration with the backend, model registries etc. This list will grow to a point where this form of top group organization is not maintainable.
```
nathan-weinberg, line 64: @cdoern these are great points and excellent justification as to why this change is needed

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+In general, moving to this structure allows for more growth when `ilab` reaches a stable release and encourages broader usage of the tool for all sorts of Open source AI models.
```
nathan-weinberg, line 66:
```suggestion
In general, moving to this structure allows for more growth when `ilab` reaches a stable release and encourages broader usage of the tool for all sorts of open source AI models.
```

docs/ilab-model-engine.md
nathan-weinberg: Agree with Charlie - I think we can start with a global config, but we can always enhance later for users to have options to create separate "projects" (there's some existing work for something like that here https://github.com/instructlab/instructlab/pull/1022)

docs/ilab-model-engine.md
nathan-weinberg: Also, what is `-i` for @cdoern? Assume it's shorthand for `interact` or something of the like based on context clues elsewhere in the document (if I'm missing somewhere it's explicitly defined lmk)

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+The mental model a user has impacts the way they use the application and whether they view it as a Proof Of Concept.
+
+Here is a diagram of what I mean when I describe this mental model:
+
+```console
+SOURCE                                                                                                                                                                                                                                              SINK
+
+        User has un-commited changes in their taxonomy                          Data is generated and then the user runs                               A model is created, and the user can tag the model               After converting the model, the
+        or specifies `--skill` `--knowledge` in the                 ==>        `ilab model train` using the newly generated data       ==>             or just list it using the new model commands            ==>      user runs the chat procces by specifying `ilab  model serve -i` or `ilab model serve` followed by `ilab model chat
```
nathan-weinberg, line 131:
```suggestion
        or specifies `--skill` `--knowledge` in the                 ==>        `ilab model train` using the newly generated data       ==>             or just list it using the new model commands            ==>      user runs the chat procces by specifying `ilab model serve -i` or `ilab model serve` followed by `ilab model chat
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+|_______convert
+|       |_____model
+|_______train
+|       |_____model
+```
+
+This is opposed to the structure in this EP, which might have a duplicate here and there for something like `download`, but in this structure duplicate commands are the norm and will result in a larger and clunkier codebase that is confusing to read.
+
+This first structure looks nice, and `ilab download model` sounds nice. However in terms of implementation, this makes little sense. in [click](https://click.palletsprojects.com/en/8.1.x/), the cli library we use, `ilab` is a "group" that commands are currently "grouped under". In this Model, `ilab` would be a group, `download`, `convert` etc would be a sub-group, and `model` would be a command under ALL of these groups. This would require an implementation of the `model` function in different packages all linking to a different group parent. From the perspective of user contributions and general code readability doesn't make much sense. Part of this design is to encourage user contributions by making the structure of the codebase logical. While these commands make the commands sound better, they make them harder to group, and understand.
```
nathan-weinberg, line 168:
```suggestion
This first structure looks nice, and `ilab download model` sounds nice. However in terms of implementation, this makes little sense. In [click](https://click.palletsprojects.com/en/8.1.x/), the CLI library we use, `ilab` is a "group" that commands are currently "grouped under". In this Model, `ilab` would be a group, `download`, `convert` etc would be a sub-group, and `model` would be a command under ALL of these groups. This would require an implementation of the `model` function in different packages all linking to a different group parent. From the perspective of user contributions and general code readability doesn't make much sense. Part of this design is to encourage user contributions by making the structure of the codebase logical. While these commands make the commands sound better, they make them harder to group, and understand.
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+how to use `ilab` with this new structure:
+
+1. `ilab config init`
+2. `ilab model download`
+3. `ilab data generate`
+4. `ilab model train --convert`
+5. `ilab model serve -i` (or followed by `ilab model chat`)
+
+So, not only we do remove one of the steps (with -i) for most users, but it is actually clear what is happening. For example one may ask: What does `ilab init` do? While `ilab config init` clearly is initializing the config of the cli. A more clear example even is: `ilab generate`. Are we generating a model? A config? No we are generating *data*: `ilab data generate`.
```
nathan-weinberg, line 191:
```suggestion
So, not only we do remove one of the steps (with `-i`) for most users, but it is actually clear what is happening. For example one may ask: What does `ilab init` do? While `ilab config init` clearly is initializing the config of the cli. A more clear example even is: `ilab generate`. Are we generating a model? A config? No we are generating *data*: `ilab data generate`.
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Changes to Existing flow
+
+The current `ilab` commands will still work. Users will be able to type commands like `ilab model train` or `ilab train` for the foreseeable future to ensure feature parity. Eventually, this alias should be removed, and only the sub-commands should probably exist.
```
nathan-weinberg, line 195: We should target an explicit release to do so (outside of this general proposal, in our usual planning meetings/channels)

nathan-weinberg (changes requested):
@cdoern Overall this proposal is very strong and I wholeheartedly agree with the approach - I have some minor requested changes but once they are addressed I will happily add my approval to the pile

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+|       |
+|       |____diff
+|       |____check
+|       |____download
+```
+
+The main point of this structure is to introduce a group hierarchy and to improve the usability of the existing `ilab` structure. Some key things missing in `ilab` currently are:
+
+1. `ilab init` does too many things.
+    - We don't want "black box" commands. This creates bad UX and also ties too many processes together in which one might eventually need to be changed.
```
russellb, line 60: It's not clear that the proposal changes this at all.  `ilab init` is just `ilab config init`. It seems like a distraction.

russellb (changes requested):
posting a few comments before a meeting starts, though I'm not actually finished with reviewing it

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+|       |____diff
+|       |____check
+|       |____download
+```
+
+The starred commands are new, and will be dealt with in another enhancement.
+
+#### Goals for 5/30/2024 milestone release of CLI
+
+For the next few milestones, it has been identified that the overall structure shift should be in place. We should aim to have the base hierarchy of commands in place that will exist for the future of `ilab` so that when users become familiar with the CLI, it does not change radically soon after launch.
```
russellb, line 109:
```suggestion
For the next few milestones, it has been identified that the overall structure shift should be in place. We should aim to have the base hierarchy of commands in place that will exist for the future of `ilab` so that when users become familiar with the CLI, further drastic changes are not needed.
```

since the project has already launched

docs/ilab-model-engine.md
russellb: I would prefer to remove this from this proposal. Changing the CLI command structure is not dependent on making this configuration management change.

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+This means we will now have commands that have a parent group `ilab` followed by a sub-command/group terminated by a grandchild command: `ilab model list` for example. This workflow not only adds purposeful organization to this project, but will encourage contributors to add commands as they use `ilab`, and file bugs in specific parts of the codebase that need work as opposed to the current flow which is hard to keep track of from a user's point of view.
+
+### Necessity of a input -> configuration -> result mental model
+
+The commands uncovered by adding this structure resemble the typical architecture for an "engine". `ilab` is not complete without the mechanisms to create, list, delete, and inspect the models. Models, as container images do in container engines, act as the configuration for the end result: the chat process. The interesting thing about this analogy, is that there needs to be a pre-cursor to configuration as well. There is the container image, the container, and the Containerfile. The Containerfile is the raw user input that leads to an image.
```
russellb, line 119:
```suggestion
The commands uncovered by adding this structure resemble the typical architecture for an "engine". `ilab` is not complete without the mechanisms to create, list, delete, and inspect the models. Models, as container images do in container engines, act as the configuration for the end result: the chat process. The interesting thing about this analogy, is that there needs to be a pre-cursor to configuration as well. There is the container image, the container, and the `Containerfile`. The `Containerfile` is the raw user input that leads to an image.
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+
+        User has un-commited changes in their taxonomy                          Data is generated and then the user runs                               A model is created, and the user can tag the model               After converting the model, the
+        or specifies `--skill` `--knowledge` in the                 ==>        `ilab model train` using the newly generated data       ==>             or just list it using the new model commands            ==>      user runs the chat procces by specifying `ilab  model serve -i` or `ilab model serve` followed by `ilab model chat
+        new `ilab data generate`
+
+```
+
+So in `ilab` the clear source of information is taxonomy, or the new skills/knowledge brought into generate. This data then goes through processes that end up in the sink, which is a running/actionable process which the user can interact with. Providing commands that clearly display these different process is key. Groups for these larger "buckets" of processes makes sense as well due to the likelihood of more commands being needed to properly interact with the newly generated model.
```
russellb, line 136:
```suggestion
So in `ilab`, the clear source of information is the taxonomy, or the new skills/knowledge brought into generate. This data then goes through processes that end up in the sink, which is a running/actionable process which the user can interact with. Providing commands that clearly display these different processes is key. Groups for these larger "buckets" of processes make sense as well due to the likelihood of more commands being needed to properly interact with the newly generated model.
```

docs/ilab-model-engine.md
russellb, line 191:
```suggestion
So, not only do we remove one of the steps (with `-i`) for most users, but it is actually clear what is happening. For example, one may ask: "What does `ilab init` do?" while `ilab config init` clearly initializes the CLI's config. A more clear example is `ilab generate`. Are we generating a model? A config? No, we are generating *data*: `ilab data generate`.
```

docs/ilab-model-engine.md
nathan-weinberg: Unless you mean 1.0 launch @cdoern?

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+|_____list
+|_____serve
+|_____test
+|_____train
+```
+
+vs. after:
+
+```console
+ilab
```
romfreiman, line 31: I would say we should have those on an API resource level with OpenAPI and then have swagger to generate the CLI according. Otherwise, it would be very hard to make it api driven in the future,

docs/ilab-model-engine.md
n1hility: There are some thorny details with changing the data store, so agree that ideally its split out into a separate proposal.

Ex:
/var/lib/models is normally not writable, so I assume maybe you were thinking XDG_DATA_HOME/ilab/models (~/.local/share/ilab/models) and XDG_CONFIG_HOME/ilab (~/.config/ilab)

Other questions are what happens if someone has two working areas, do all checkpoints end up in the model directory or do they end up in the workspace.

Also relatedly, huggingface recently abandoned this approach with a user cache directory for model, so something to think about. I agree with the goal that we probably want an easy way to share downloaded models.

n1hility:
Overall I think this is a good proposal. It offers a good UX with lots of future expansion potential

docs/ilab-model-engine.md
leseb: It's a detail but using `/var/lib/ilab/models` is preferred.

docs/ilab-model-engine.md
cdoern: yep, I think I specify it implicitly below but will add a note.

docs/ilab-model-engine.md
cdoern: thanks!

docs/ilab-model-engine.md
cdoern: Good point, this was an old reasoning anyway, will remove

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+|       |____check
+|       |____download
+```
+
+The main point of this structure is to introduce a group hierarchy and to improve the usability of the existing `ilab` structure. Some key things missing in `ilab` currently are:
+
+1. `ilab init` does too many things.
+    - We don't want "black box" commands. This creates bad UX and also ties too many processes together in which one might eventually need to be changed.
+2. ambiguity in "what" we are generating, training etc.
+    - The current structure of `ilab` requires an in depth knowledge of the ordering of commands and background knowledge for how our CLI is built. This is not the best way to grow. Commands that are not intuitive and require a deep dive into the docs just to get started might not engage users as we hope to. Having an intuitive structure in which people can see easily from the docstring what each commands does and how they might use them, makes more sense. We are generating data, then training a model, running a server to interact with that model, then finally chatting with the model
```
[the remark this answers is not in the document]
cdoern: I think this should be kept. The point of this redesign is to make it more clear what is being acted on.

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,201 @@
...
+3. Some commands like `ilab test` and `ilab list` should be renamed to show their true purpose (`ilab model evaluate`and `ilab taxonomy diff`)
+4. We will hit a point where we want to add more commands to `ilab` due to expected integration with the backend, model registries etc. This list will grow to a point where this form of top group organization is not maintainable.
+
+In general, moving to this structure allows for more growth when `ilab` reaches a stable release and encourages broader usage of the tool for all sorts of Open source AI models.
+
+Outside the scope of this EP, but still important in the near future, the following structure is preferred:
```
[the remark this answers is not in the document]
cdoern: sure

docs/ilab-model-engine.md
cdoern: no, this is an old sentence (this EP has been around for a while :) )

cdoern:
@russellb (GH wont let me quote reply)
> I would prefer to remove this from this proposal. Changing the CLI command structure is not dependent on making this configuration management change.


I think this EP is meant to cover the cmd redesign sure but also the way ilab is thought about and generally used. That is why initially had the title as "model engine". Sure,  we are redesigning the cmds, but we are also redesigning how we think about using ilab, its scope, etc. 

A lot of your comments seem to be removing some parts of that (the future structure, the ilab init stuff, etc). I think this is something I'd really like to leave, as well as the original title, and the future structure to be discussed in another EP as all of these things fit more with the mental model I was hoping to get across with this EP. 

Pushing up a new set of changes but hoping to roll back some of them if possible! lmk what you think.

docs/ilab-model-engine.md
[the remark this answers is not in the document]
nathan-weinberg: Also in favor of keeping this passage, with Russell's text suggestions

docs/ilab-model-engine.md
nathan-weinberg: @cdoern spacing here still needs to be fixed

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,160 @@
...
+|       |____init
+|       |
+|_______tax
+|       |
+|       |____diff
+|       |____check
+|       |____download
+```
+
+Note: -i will serve and start a chat process as opposed to the current flow which implicitly starts servers in certain commands
```
nathan-weinberg, line 57:
```suggestion
Note: `-i` will serve and start a chat process as opposed to the current flow which implicitly starts servers in certain commands
```

nathan-weinberg (approved):
I have a few outstanding comments but I agree with this document as a whole and personally am ready for us to move forward with implementation, so approving from my side

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,160 @@
...
+8. `ilab chat`
+
+how to use `ilab` with this new structure:
+
+1. `ilab config init`
+2. `ilab model download`
+3. `ilab data generate`
+4. `ilab model train --convert`
+5. `ilab model serve -i` (or followed by `ilab model chat`)
+
```
markstur, line 149: So if I apply the enhancements like -i and ensure a server for generate (and chaining) it would be:
`ilab init download generate train --convert serve -i`.  (maybe -i should be --chat)

So, removal of steps is separate from the redesign. I guess the argument here is that adding more words to the commands adds clarity.  Not sure that's really good thing but...

I do, however, agree with the sub-module cli design so we're good.

docs/ilab-model-engine.md
markstur: we'll have to revisit having 2 ways of doing everything for sure.  I guess it is comforting to suggest we a dual path for transition.

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,160 @@
...
+## Changes to Existing flow
+
+The current `ilab` commands will still work. Users will be able to type commands like `ilab model train` or `ilab train` for the foreseeable future to ensure feature parity. Eventually, this alias should be removed, and only the sub-commands should probably exist.
+
+## Known Issues
+
+### click
+
+click doesn't like the setup we currently have in `ilab`. So, adding sub-parent commands won't be as easy as creating a new `click.group`. We will need to make different libraries, each of which is a click group most likely.
```
markstur, line 160: If click is bad at this, is argparse better?

Also the "different libraries" comment is not clear to me and makes me concerned about how restructuring for click might be different than restructuring for future API or for general good structure.

markstur (commented):
I went through and commented again on some of things that don't really jive with my way of thinking, but these are mostly just comments and not change requests.  I know we are near the end of this.  :)

But it is easier to comment critiques than bother noting the parts I agree with.  To be clear... the argument might not be immediately compelling to me, but I do think we will benefit from the sub-command design eventually.  So it's probably better to do it soon instead of using a bunch of flags (generate --data?) or commands like generate-data as work-arounds for subcommands.  <-- that was me attempting to agree with you. :)

docs/ilab-model-engine.md
markstur: I'm also concerned we will want to restructure the CLI again once we have an API spec

.spellcheck-en-custom.txt
```diff
@@ -120,3 +120,12 @@ XT
...
 YAML
+UX
+CWD
+EP
+codebase
+config
+docstring
+podman
+CLI
+CLI's
```
russellb, line 131: This isn't something to block on, but if you have to update the PR anyway, it'd be good to run `make spellcheck-sort`.

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,160 @@
...
+
+This document describes a new structure for `ilab`, consisting of sub-commands that act as parent or "group" commands for further sub-commands. Currently the only group command is `ilab`. If the models `ilab` produces are meant to be the golden standard for open source AI, the tool to use and manage these models needs to be fully-fledged and extensible, especially for models we have not produced but should be able to run.
+
+open source container tools like [podman](https://podman.io/) and [docker](https://www.docker.com/) are commonly viewed as engines. `ilab` should be a model engine, managing the deployment and creation of AI models. To be clear, I will be using tools like podman as a structural analogy here with full understanding they have different implications and use-cases.
+
+## Key Component
+
+### InstructLab Structure Redesign
+
+Here are a few outlines. The first two that represent the state of `ilab` before/after this enhancement and the third describes `ilab` after I introduce new key commands to this structure. I am including the second outline for clarity purposes, but this EP only covers the contents of the first structure.
```
russellb, line 11:
```suggestion
Here are two outlines. They represent the state of `ilab` before/after this enhancement.
```

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,160 @@
...
+|       |
+|_______data
+|       |
+|       |____generate
+|       |
+|_______config
+|       |
+|       |____init
+|       |
+|_______tax
```
russellb, line 50: Minor nit -- would prefer `taxonomy` instead of the shorthand `tax`

docs/ilab-model-engine.md
russellb: Agree that this is a separate feature

docs/ilab-model-engine.md
```diff
@@ -0,0 +1,160 @@
...
+
+The main point of this structure is to introduce a resource based hierarchy and to improve the usability of the existing `ilab` structure. At the top level of commands, we are not thinking about verbs but resources commonly managed in `ilab`, under them fall some actions. Some key things missing in `ilab` currently are:
+
+1. Ambiguity in "what" we are generating, training etc.
+    - The current structure of `ilab` requires an in-depth knowledge of the ordering of commands and background knowledge for how our CLI is built. This is not the best way to grow. Commands that are not intuitive and require a deep dive into the docs just to get started might not engage users as we hope to. Having an intuitive structure in which people can see easily from the docstring what each command does and how they might use them, makes more sense. We are generating data, then training a model, running a server to interact with that model, then finally chatting with the model
+2. Some commands like `ilab test` and `ilab list` should be renamed to show their true purpose (`ilab model evaluate`and `ilab taxonomy diff`)
```
russellb, line 63:
```suggestion
2. Some commands like `ilab test` and `ilab list` should be renamed to show their true purpose (`ilab model evaluate` and `ilab taxonomy diff`)
```

russellb (changes requested):
My biggest remaining concern is about dropping the discussion of the data and config store.

I would include the `-i` feature in a similar category of a new feature that seems separate from this. I would prefer removing that, too.

The other comments are just trivial things I saw while reading through again.

docs/ilab-model-engine.md
leseb: @cdoern I second that using groups might not be straightforward and the current `lab.py` might require a bit of remodeling but I think the statement "adding sub-parent commands won't be as easy as creating a new `click.group`" is a bit misleading. Implementing something like `ilab foo bar --baz` is possible :). I'm not sure why we need to make different libraries. Thanks!

docs/ilab-model-engine.md
cdoern: sure, I can remove this

cdoern:
Ok, I removed the new flags I want to add and the bits about the global model store/config. 

cc @russellb @n1hility @markstur PTAL :pray:

russellb (approved):
lgtm, thank you @cdoern !

PR #50, opened 2024-05-15
Introduce the Taxonomy Caching Service

derekhiggins:
Describes concept of a service that is intended to preserve Taxonomy knowledge documents.

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -7,8 +7,10 @@ arge
 arXiv
 backend
 backends
+BASEURL
 benchmarking
 Bhandwaldar
+CDN
 CLI
 cli
 Colab
@@ -35,6 +37,7 @@ GiB
 Gmail
 gpu
 hipBLAS
+https
 ilab
 impactful
 Inferencing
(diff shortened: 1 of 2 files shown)


hickeyma (commented):
Thanks @derekhiggins for pushing this proposal.

Reading the proposal, it would seem that your primary goal is to preserve a copy of knowledge document(s) that are referenced in a knowledge contribution to the taxonomy repository. If this is the case, then to me it is more a backup/replication service than a cache service. I think the use case is valuable regardless.

I would like to see raised in the proposal:

- Potential solutions with pros/cons. Would syncing between the source repo and a InstructLab knowledge documents repo be a runner?
- Data regulation issue with storing the data

nathan-weinberg:
@derekhiggins what's going on with this?

derekhiggins:
> @derekhiggins what's going on with this?

Nothing currently, It was an idea that I thought at the time would be necessary, I'm no longer close enough to the instructlab project to know if it is still relevant. If it isn't we can close it, if it is I can refresh it and help address any outstanding questions but I now longer have the time to work on it that I had when I create this PR.

PR #52, opened 2024-05-17
Introduce InstructLab Profiles Managed via `ilab profile...` to Run Key Commands at Different Fidelity Levels

cdoern:
This enhancement discusses "more intensive" training and data generation techniques as well as a new Data Mixing command. This is all built off of the command redesign. The goal here is to produce higher fidelity models using the CLI.

### DIFF
--- a/docs/ilab-profile.md
+++ b/docs/ilab-profile.md
@@ -1,0 +1,392 @@
+# Introduce Functionality to Utilize the InstructLab backend: `ilab` Profiles, and Command Adaptations
+
+This document describes adding different data generation, mixing, and training configurations to the `ilab` CLI to enable higher-fidelity training using the backend code. By higher-fidelity, we mean models that perform better, were trained on better hardware, off of larger data sets, and used more intensive training techniques.
+
+Users will be able to set a pre-defined profile type in their config.yaml that would enable sane defaults for higher fidelity training, generation, and evaluation. These defaults will funnel into existing and new flags for the CLI.
+
+## `ilab` Profiles
+
+### Building off of the InstructLab Structure Redesign
+
+Commands like: `ilab model train` will still be a single command offered to users. However, new "profiles" offered by the CLI will enable users to toggle what mode they are running `ilab model train` in.
+
+Eventually, we will want to enable `ilab profile init` and `ilab profile set` commands that allow users to initialize a custom profile based on one of the pre-defined ones. This is not targeted for an upcoming release, though.
+
+### Immediate Goals and Core Principles
+
+For the near future, there will be upper level profiles that can be initialized via `ilab config init`.
+
+Profiles should at first be static and not exposed to users. Internally, a profile would look something like:
+
+```yaml
+profile:
+   train:
+       gpus: 0-4
+       epochs: 15
+       effective_batch_size: 48
+       save_samples: 2
+       learning_rate: 1
+       warmup_steps: 1000
+       deepspeed:
+           enabled: True
+           deepspeed_offload: ["cpu"]
+           quantize_dtype: ["nf4", "fp8"]
+           deepspeed_config: foobar.json
+       
+```
+
+These can be rendered as files or in code.
+
+#### Where would this be stored?
+
+Click has the ability to associate objects with the context passed around the command structure. The profile could be:
+
+1. a ctx.obj additionally to the `ctx.obj.config` object. So this would be referenced as `ctx.obj.profile`.
+2. the profile could live inside of the config obj, moving some of the existing generation options into a sub object called `profile`. so this would look something like `ctx.obj.config.profile` which would have sub-elements like `.training`, `.evaluation`, `.generation`. 
+
+#### All new options will be flags with variable levels of expertise
+
+Options that are introduced in new commands should exist in the profile AND as command line arguments via click. **Click has a way to hide arguments using an `active` option one can implement in a custom class**. So some of the new more "expert" level flags will either be hidden unless a specific profile is enabled.
+
+#### Is this just for training?
+
+No, adding user profiles for generation and evaluation would be ideal. The forthcoming backend libraries are taking an approach where it will be easiest to have pre-baked "config" objects that will feed into them. We need to be careful about how we introduce new concepts like this. Adding lower level intensive configurations to ilab is confusing to community users.
+
+**PLEASE NOTE:**
+
+**For immediate releases I would introduce the idea of a profile, a way to set these via `ilab config init` and hardcoded profiles that plug into key `ilab` commands namely training, generation, and evaluation.**
+
+### Reasoning
+
+The profile settings will be used as arguments for most if not all libraries being introduced to the `ilab` backend.
+
+Plugging into hardware acceleration and multi-phase training is the logical next step for `ilab`. Ensuring we do this in a clean way that does not overload our current commands is also crucial. Many of the processes in the backend are confusing so we want to abstract some of the steps away from users while also giving them a reasonable amount of choice in configuring these new processes. However, maintaining the current laptop story is important to users without hardware access. Splitting these two paths into separate profiles clearly defined for users maintains the integrity of each.
+
+#### Single Command Approach Across Fidelity Levels 
+
+Having a single entrypoint in terms of each command is a safe bet, since making different training, eval, and generation commands for each technique we introduce will scale poorly. `ilab model train`, for example, after setting a profile will account for most of the use cases `ilab` targets. The same goes for `ilab data generate` and `ilab model evaluate`.
+
+## Representation of the Profile in the config.yaml
+
+Adding onto the current config.yaml, adding a profile would be as simple as:
+
+```
+chat:
+  context: default
+  greedy_mode: false
+  logs_dir: data/chatlogs
+  max_tokens: null
+  model: models/merlinite-7b-lab-Q4_K_M.gguf
+  session: null
+  vi_mode: false
+  visible_overflow: true
+general:
+  log_level: INFO
+profile: multi-gpu
+serve:
+  gpu_layers: -1
+  host_port: 127.0.0.1:8000
+  max_ctx_size: 4096
+  model_path: models/merlinite-7b-lab-Q4_K_M.gguf
+``` 
+Notice I am proposing that if a profile is specified, the generation defaults move inside of the profile. This will be discussed more below.
+
+Also, there should most likely be a default **cpu** profile always set if the user does not choose a different one.
+
+## Proposal for Default Profiles
+
+1. **CPU Only, Laptop Profile**
+    - low instruction number for generation
+    - gpus: -1 for training, eval, generation
+    - no deepspeed config or accelerator config for training
+    - low epoch number for training
+    - single phase training, no eval support given that there are no GPUs.
+2. **Single GPU, Generate, Train, Eval profile -- CUDA**
+    - mid level instruction number for generation, higher than the default 10 (100-500)
+    - eventual deepspeed support with LoRA and QLoRA in training
+    - 10+ epochs for training
+    - eval on checkpoints if hardware permits (depends on vRam)
+3. **Multi GPU, Generate, Train, Eval profile -- CUDA**
+    - high instruction count, 500 or so for generation
+    - deepspeed support without quantization or LoRA
+    - evaluation on checkpoints after a phase of training
+
+**Profile 3 would most likely have sub-profiles for different GPU support.**
+
+These initial pre-baked profiles will allow for most of the dynamic range we are looking for in upcoming releases.
+
+## Workflows using profiles
+
+1. `ilab config init --profile cpu_only`
+2. `ilab model generate`
+3. `ilab model train`
+
+to use multi-gpu:
+
+1. `ilab config init --profile multi_gpu`
+2. `ilab model generate`
+3. `ilab model train --model-dir=./phase00/model --data-dir=./phase00/data`
+4. `ilab model evaluate --benchmarks=mmlu --input-dir=./phase10/model --output-dir=./phase10/data`
+
+the above multi-gpu is a good example of where the profiles come into play. If the user had not set their profile, they would need to specify flags like:
+
+`ilab model generate --num-instructions=1000 --num-cpus=15`
+`ilab model train --devide=cuda --gpus=0:4 --epochs=15 --config=training_config.json --accelerator=deepspeed --deepspeed_config=ds_config.json --model-dir=./phase00/model --data-dir=./phase00/data --model-repo=instructlab/granite-7b-lab`
+`ilab model evaluate --benchmarks=mmlu --gpus=1-2 --data-dir=/path-to-gen/mixed --tasks=["mmlu.algebra"...] --input-dir=./phase10/model --output-dir=./phase10/data`
+...
+
+Adding the CPU only profile will be really beneficial to `ilab` users as it will eliminate confusion over some common use cases.
+
+Most of the time choosing these flags is not clear. For expert users this is where `ilab profile init` would come in for future release. They would have a custom profile with bespoke options. However, for most other users a pre-baked single GPU profile will be very useful. They could also for now just override with flags.
+
+## Adaptations to Existing Commands as well as New Commands for Profiles and the Backend Code
+
+`ilab` is gaining new functionality to plug into the backend pipeline. Exposing this in a way that preserved the current UX but also enables power users is a core design principle.
+
+### Implementation specifics for `ilab model train`
+
+`ilab model train` will maintain all of its current flags which will represent overrides for the values stored in the training section of a user profile. This means all of the existing flags will be ported to a default config with sane values set for all that are applicable.
+
+The existing code which runs the HF trainer will be ported to another library which can also run multi-phase training. This library will be shelled out to using torchrun.
+
+#### The Importance of a User Profile: Future Mixing and Matching of Training Techniques
+
+Trying to find a middle ground between server use cases and laptop use case is important. Exposing an eventual way to mix and match different training techniques makes sense for the following use cases:
+
+1. Developer with a Gaming PC:
+   * Transformers+Pytorch support QLoRA && FSDP. While deepspeed might be a more "server-rack" use-case, having multi-phase training in the CLI for anyone with a consumer GPU makes sense.
+2. Someone interested in ML, has a Homelab, or *anything with 2 GPUs*
+   * Transformers+Pytorch supports DeepSpeed on a single system spreading the training over the GPUs. Any professional or hobbyist that has 2 GPUs will be looking for this option
+3. The laptop use case
+   * Maintaining QLoRA as the performant training mode for the laptop is crucial as most environments cannot handle the full models. However, unlocking some better results by using FSDP+QLoRA could improve local results and get people more interested in InstructLab.
+
+The transformers library supports the following combinations:
+
+1. LoRA + BitsNBytes Quantized Model + DeepSpeed (QLoRA + DeepSpeed)
+   - someone with a consumer GPU could get much better training results from using this combo as opposed to torch without deespeed and inter-checkpoint eval
+2. LoRA + FSDP
+3. LoRA + DeepSpeed
+   - This enables native multi-GPU support out of the box. Model Loading, trainer setup, and deepspeed initialization are all handled by transformers. I have tested this and it works great on something like 2x4090 (3090 too) or 2xA10
+
+**NOTE**: Changing the way we use the transformers library is not in scope for upcoming releases. This document serves as a proposal for upcoming and future enhancements to training, generation, and evaluation. the notes above regarding combining LoRA, DS, FSDP, etc are meant to guide us in future training enhancements and to describe how there are too many options to add as conflicting flags.
+
+Given these requirements a full profile internally for training would look like:
+
+```yaml
+profile:
+   train:
+       gpus: 0-4
+       epochs: 15
+       effective_batch_size: 48
+       save_samples: 2
+       learning_rate: 1
+       warmup_steps: 1000
+       deepspeed:
+           enabled: True
+           deepspeed_offload: ["cpu"]
+           quantize_dtype: ["nf4", "fp8"]
+           deepspeed_config: foobar.json
+       
+```
+
+all available options are:
+
+```yaml
+model_path: str
+data_path: str
+output_dir: str
+
+num_gpus: int
+max_seq_len: int
+max_batch_len: int
+num_epochs: int
+effective_batch_size: int
+save_samples: int
+learning_rate: float
+warmup_steps: int
+
+deepspeed:
+    enabled: bool
+    deepspeed_config: str
+    ds_offload_strat: Enum["cpu", "nvme", None]
+    cpu_offload_optimizer: bool
+    cpu_offload_params: bool
+    quantize_dtype: Enum["nf4", "fp8", None]
+    
+lora: 
+    enabled: bool
+    lora_rank: int
+    lora_alpha: float
+    lora_dropout: float
+    target_modules: list
+```
+
+These options will be in the default profiles AND will be available as flags in the CLI. This may seem like a lot of flags to be adding, but as mentioned above some of these will be hidden behind "expertise" levels.
+
+NOTE:
+
+**Many of these are contingent upon one another, the CLI will validate a user profile when it is set. The CLI will error out if a profile is invalid.**
+
+### Implementation Specifics for `ilab data generate`
+
+**The below flags are an assumption and I need confirmation from the SDG team**
+
+`ilab data generate` configuration should also be placed into top level profiles. The end result will look like:
+
+```yaml
+profile:
+   train:
+       gpus: 0-4
+       epochs: 15
+       accelerator: deepspeed
+   generate:
+       gpus: 1-2
+       num_instructions: 1000
+       taxonomy_path: /path/to/large/taxonomy
+       num_grounded_questions: 10
+       num_samples: 10
+       config: path_to_more_sdg.yaml
+       ...
+```
+
+Some of the new flags and profile options are:
+
+- --gpus: str, defined which GPUs to use for generation
+- --num-grounded-questions: int
+- --num-samples: int
+- --config: str, points to a config which outlines internal SDG options outside of the CLI
+
+#### This is currently _partially_ in config.yaml... why would we change that?!
+
+A lot of this exists in the current `config.yaml` structure. As compared to the other entities in config.yaml, the generation settings stand out as CLI options placed into the config for convenience. In this case convenience is not the same as clarity of use. `ilab init` populates these for most users and then they leave them be.
+
+Opting to move generation options into the profile is preferred and more verbose in the sense that if someone is setting a profile, they are aware of the settings being enabled. Of course, readers may be wary of moving them entirely out of the config.yaml. I will outline the alternative of embedding the profile **into** the config.yaml below.
+
+#### Following the above design principles
+
+A new concept is an additional `config` option within the generation profile which points to optional configuration yaml that contains more data generation settings, this can live as a file reference in the profile. At first, this will be hardcoded to a single default. Eventually if a user wants to override things like the default prompts, we should allow custom configs. 
+
+Mixing of the generated data before training will occur as part of generation. If we choose to expose options for data mixing to the users, these would be some of the options we need:
+
+- --num-util-proc=int (?)
+- --output-dir=str (defaults to generated/mixed)
+- --knowledge-recipes=[]str (path to yaml)
+- --skill-recipes=[]str (path to yaml)
+
+These can be introduced to either the profile where applicable or just as flags if they are highly dynamic. Reminder, everything introduced in these new commands will also be CLI flags.
+
+**So, a generation profile would have the following entries**:
+
+```yaml
+    generate:
+       model_family: mixtral
+       gpus: 1-2
+       chunk_word_count: 1000
+       num_instructions: 1000
+       model_endpoint_url: 127.0.0.1:8000
+       taxonomy_path: /path/to/large/taxonomy
+       num_grounded_questions: 10
+       num_samples: 10
+       config: path_to_more_sdg.yaml
+```
+
+**If mixing is to be exposed that process would require a profile like**:
+
+```yaml 
+    generate:
+       model_family: mixtral
+       gpus: 1-2
+       chunk_word_count: 1000
+       num_instructions: 1000
+       model_endpoint_url: 127.0.0.1:8000
+       taxonomy_path: /path/to/large/taxonomy
+       num_grounded_questions: 10
+       num_samples: 10
+       config: path_to_more_sdg.yaml
+       mix:
+           num-util-proc: 10
+           output-dir: generated/mixed
+           knowledge-recipes: k_recipes.yaml
+           sill-recipes: s_recipes.yaml
+```
+
+You will notice some of the currently existing flags have been added to the profile like chunk_word_count and endpoint_url. Any flag that is relatively constant or can be viewed to have a sane default should live in the profile. Remember, all of these can be overriden via flags by expert users.
+
+### `ilab model evaluate`
+
+InstructLab will run benchmarks for model performance in upcoming releases. These benchmarks are useful for inter-checkpoint evaluation and full scale evaluation at the end of multi-phase training.
+
+As of writing the two benchmarks are MMLU Bench and MT Bench. You might only want to run certain benchmarks depending on the particular changes made to the model. Ex: MMLU is useful for knowledge additions.
+
+MMLU bench needs the following options:
+- --model: str, default is granite (?)
+- --tasks: []str, default is {"mmlu_pr"}. This is the list of MMLU tasks to run.
+- --few-shots (int)
+
+MT bench needs the following options:
+- --api-key (?)
+- --model: str, default is granite (?)
+- --endpoint: str, vLLM compatible endpoint to use for the benchmark
+
+Some other flags would be:
+- --output-dir: str, determines where the best checkpoint is put for the next phase of training
+- --input-dir: str, takes the directory of the model/checkpoint to evaluate.
+- --benchmarks: []str, this will let the user specify which benchmarks in their profile to run
+- --prune: bool, this will let the user prune the other checkpoints not evaluated as "best" when running eval in between training phases
+- --gpus: number of GPUs to use. Internally called batch-size for MMLU bench
+
+
+All of these options will be in the profile but can also be overriden by flags if needed. 
+
+Here is an example profile specifying settings for mmlu benchmarking:
+
+```yaml
+profile:
+   train:
+       gpus: 0-4
+       epochs: 1
+       ...
+   generate:
+       gpus: 1-2
+       num_instructions: 1000
+       taxonomy_path: /path/to/large/taxonomy
+       num_grounded_questions: 10
+       num_samples: 10
+       ...
+   evaluate:
+       prune: True
+       input-dir: /models/foobar
+       outpur-dir: /bench/output
+       gpus: 3-4
+       mmlu:
+           few-shots: 1
+           model: granite  
+           tasks: "mmlu-pr"
+       ...
+```
+
+a user could then simply run something like `ilab model evaluate --input-dir=foobar --benchmarks=mmlu`
+
+checkpoint evaluation and full evaluation are different but ultimately run similar benchmarks just on different types of data. So a single command here makes sense. The profile will simply outline the config for the supported benchmarks.
+
+The command will output a dictionary of scores from evaluation as well as pairs of questions and answers depending on the benchmark run. The scores will be shown to the user, and the data will all live in the `output-dir`.
+
+Training depends on a specific directory format to pick up a checkpoint to resume from after evaluation is run. The specifics of that will be coordinated by the training library writers and the evaluation library writers, then implemented by the CLI.
+
+**A full profile for evaluation would have the following structure**:
+
+```yaml
+       gpus: 3-4
+       mmlu:
+           tasks: "mmlu-pr"
+           few-shots: 1
+           model: granite
+           data-dir: generated/mixed
+           
+       mt:
+           model: granite
+           endpoint: 127.0.0.1:8080
+           
+```
+
+**Question: how will the library know which form of eval to run?**


russellb:
Added a `hold` because this is blocked on the design in #48 and the corresponding code in github.com/instructlab/instructlab/pull/990 (and maybe others)

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,176 @@
...
+ilab
+|
+|_______model
+|       |
+|       |____convert
+|       |____download
+|       |____train (--convert)
+|       |    |
+|       |    |______lofi *
+|       |    |______hifi *
```
khaledsulayman, line 66: I think lofi and hifi probably make sense w.r.t. the models themselves. However, currently the size of the model is not the only differentiator between the correlating CLI and backend components. Many of the full-size workflows run a bunch of stuff that the CLI does not currently, and might not ever on a laptop.

Initially, we joked about specifying —turbo to run the backend version of a command, and using the laptop version by default. Whatever it is, I think the name should indicate that it’s not just using a higher fidelity model, but actually just a bigger workload in general.

russellb, line 66: I'm not a fan of any variation of "low" and "high" in the commands. I think it's too limiting. In practice, I expect we'll need more of a sliding scale over time and not a binary set of choices.

An alternative would be to expose more directly the algorithm we're talking about, and algorithm-specific tunables. 

* `ilab train qlora ...`
* `ilab train deepspeed ...`
* ...

something like that.

It's possible to also have an `ilab train auto` where we try to make some automatic decisions, or maybe even a set of t-shirt size type options geared at a range of resource profiles, but I think it may be premature to rush into those while things still seem to be evolving so much.

russellb (changes requested):

docs/lofi-hifi-backends.md
khaledsulayman: to add to this, I think that if a user tries to run `ilab train` by itself, they should be shown a help message outlining the choices and outlining hardware limitations and other considerations of each, e.g.
```
usage: ilab train <training_algo>

Available Training Options:
  qlora: Requires n GPUs, m GB RAM | suitable for running on laptop...
  fsdp: Requires n GPUs, m GB RAM | suitable for running on RHEL AI/larger infra...
```

@RobotSail also brought up the brilliant suggestion of having it interactively help the user choose, which I am very much on board with.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+### Reasoning
+
+
+Plugging into hardware acceleration and multi-phase training is the logical next step for ilab. Ensuring we do this in a clean way that does not overload our current commands is also crucial. Many of the processes in the backend are confusing so we want to abstract some of the steps away from users while also giving them a reasonable amount of choice in configuring these new processes. However, maintaining the current laptop story is important to users without hardware access. Splitting these two paths into separate commands maintains the integrity of each.
+
+
+
+### ilab model train integrated
```
jaideepr97, line 104: @cdoern what is the rationale behind `integrated` vs `qlora`? I think integrated introduces more ambiguity than just calling out the exact algorithm being used

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+This command would take something like the following arguments
+
+
+--gpus=str , describes the amount of GPUs (of what is available) to use for this process. This comes in the form of: 0-1, 8, etc.
+--quantize=bool, enabled Qlora which basically loads the model in a quantized form so it can fit on a consumer GPU
+--optimizer=str (deepspeed, fsdp) describes the optimizer framework to use during training
+# --ditributed or --containerized, triggers a containerized version of the train workflow with hardcoded dependencies.
```
jaideepr97, line 113: my understanding based on recent discussions is that we don't want the CLI spinning up and managing any containers

hickeyma (changes requested):
I think it would be better to phrase this as extending the ilab commands to be configurable to use different technologies. The title doesn't clearly define what the design doc is about.

docs/lofi-hifi-backends.md
cdoern: yep, sorry

docs/lofi-hifi-backends.md
cdoern: @RobotSail had some opinions over qlora vs phased. I think the gist is: one is a training technique the other is an algorithm. The names need to both be training techniques

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
+# Introduce Commands that Run Jobs with Different Fidelity levels for Key ilab functions
+
+This document describes adding different data generation, mixing, and training backends to ilab to enable higher fidelity training using the backend code.
+
+Currently all training is done via qlora or the like. Adding the following commands will enable higher fidelity training and introduce commands such as data mixing.
```
nathan-weinberg, line 5:
```suggestion
Currently all training is done via qlora or the like.
```
nit: can we be a bit more specific here

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+|       |____diff
+|       |____check
+|       |____download
+```
+
+
+The starred commands are the new ones under the redesigned structure. We are opting to add integrated and phased as commands under train and lofi/higi under generate rather than flags given their completely different functionality and backend structure.
+
+
+These commands would connect to the instructlab backend, which will be in the form of libraries. The lower fidelity commands if accepted will be the equivalent of what currently exists for generate and train. qlora, with pytorch, mlx, etc.
```
nathan-weinberg, line 87:
```suggestion
These commands would connect to the planned InstructLab backend libraries. The lower fidelity commands will be the equivalent of what currently exists for generate and train in the existing CLI code - QLoRA, with pytorch, mlx, etc.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+The Higher Fidelity versions would validate the existence of hardware that can properly run the generation, mixing, and training backends. At least for training, the existing infrastructure simply shells out to various python scripts, libraries, etc. So, as long as we combine this backend code into a place that can be imported into ilab without breaking other dependencies, this should be more of a structural change than a functional one. We know the backend code works on an isolated system, we just need to make it pluggable.
```
nathan-weinberg, line 90:
```suggestion
The Higher Fidelity versions would validate the existence of hardware that can properly run the generation, mixing, and training backends. At least for training, the existing infrastructure simply shells out to various python scripts, libraries, etc. So, as long as we combine this backend code into a place that can be imported into `ilab` without breaking other dependencies, this should be more of a structural change than a functional one. We know the backend code works on an isolated system, we just need to make it pluggable.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+High Fidelity can run locally on someone's laptop or desktop and even utilize deepspeed if they have GPUs. In a more powerful system, the user can also run it in a container, utilize deepspeed and potentially even distribute the workload accorss machines using torch.distributed.
```
nathan-weinberg, line 92:
```suggestion
High Fidelity can run locally on someone's laptop or desktop and even utilize deepspeed if they have GPUs. In a more powerful system, the user can also run it in a container, utilize deepspeed and potentially even distribute the workload across machines using `torch.distributed`.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+A key component here too is checkpoint evaluation. We want users to be able to have an understanding of model checkpoints starting with the ability to run evaluation on them in between phased training runs. The output of `ilab checkpoint evaluate` will tell the user what to point `ilab model train phased` arguments towards.
+
+
+### Reasoning
+
+
+Plugging into hardware acceleration and multi-phase training is the logical next step for ilab. Ensuring we do this in a clean way that does not overload our current commands is also crucial. Many of the processes in the backend are confusing so we want to abstract some of the steps away from users while also giving them a reasonable amount of choice in configuring these new processes. However, maintaining the current laptop story is important to users without hardware access. Splitting these two paths into separate commands maintains the integrity of each.
```
nathan-weinberg, line 100:
```suggestion
Plugging into hardware acceleration and multi-phase training is the logical next step for `ilab`. Ensuring we do this in a clean way that does not overload our current commands is also crucial. Many of the processes in the backend are confusing so we want to abstract some of the steps away from users while also giving them a reasonable amount of choice in configuring these new processes. However, maintaining the current laptop story is important to users without hardware access. Splitting these two paths into separate commands maintains the integrity of each.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+    * Transformers+Pytorch support Qlora&&FSDP. While deepspeed might be a more "server-rack" use-case, having multi-phase training in the CLI for anyone with a consumer GPU makes sense.
+2. Someone interested in ML, has a Homelab, or *anything with 2 GPUs*
+    * Trandformers+Pytorch supports Deepspeed on a single system spreading the training over the GPUs. Any professional or Hobbyist that has 2 GPUs will be looking for this option
+3. The laptop usecase
+    * Maintianing Qlora as the performant training mode for the laptop is crucial as most people cannot handle the full models. However, unlocking some better results by using FSDP+Qlora could improve local results and get people more interested in InstructLab.
+
+The above usecases create a spectrum of possibility of what users can do with ilab! adding `ilab model train lofi/hifi` each with different options for optimizers: Adam (default, Qlora), FSDP (Qlora OR Non-Qlora train), and Deepspeed (Non-Qlora, multi GPU) increase the amount of situations where ilab is viable. Adding other options like --multi-phase, --learning-rate, etc give the user granular control over this new multi-phased training approach.
```
nathan-weinberg, line 137:
```suggestion
The above usecases create a spectrum of possibility of what users can do with ilab! adding `ilab model train lofi/hifi` each with different options for optimizers: Adam (default, QLoRA), FSDP (QLoRA OR Non-QLoRA train), and Deepspeed (Non-QLoRA, multi GPU) increase the amount of situations where `ilab` is viable. Adding other options like `--multi-phase`, `--learning-rate`, etc., give the user granular control over this new multi-phased training approach.
```
`lofi`/`hifi` references need to be removed (probably should change the PR title as well)

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+### ilab model train phased
+
+This command would take roughly the following arguments
+
+--optimizer=str (deepspeed, fsdp) describes the optimizer framework to use during training
+--device=str , accelerator to use cpu,cuda,rocm,mlx,mps
+--model-dir=path , dir where the model to be used in this phase is located
+--data-dir=path , dir where the data for this phase is located
+
+Note there is no Lora in this command and there is no quantization. The `ilab model train integrated` command will use the transformers library with pytorch. This is because those have awesome plugins for deepspeed and fsdp WITH lora and qlora. Those are absolute necessities for community usecase. However, they will be mostly unused in the "High Fidelity" usecase.
```
nathan-weinberg, line 149:
```suggestion
Note there is no LoRA in this command and there is no quantization. The `ilab model train integrated` command will use the transformers library with pytorch. This is because those have awesome plugins for deepspeed and fsdp WITH LoRA and QLoRA. Those are absolute necessities for community usecase. However, they will be mostly unused in the "High Fidelity" usecase.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+`ilab model train phased --device=cuda --model-dir=./phase00/model --data-dir=./phase00/data`
+`ilab checkpoint evaluate ./phase05/checkpoints --output-dir=./phase10`
+`ilab model train phased --device=cuda --model-dir=./phase10/model --data-dir=./phase10/data`
+....
+
+Basically they would run phased with an eval in between. The Eval looks at the checkpoints output by the previous phase and outputs a model dir in the next phase's working directory. 
+
+## Alternatives
+
+The other alternative is to keep the same train and generate commands and instead add a --backend or --hifi flag to trigger the high fidelity code. The issue here is that ilab train is already overloaded with pytorch, mlx, etc. Adding more switches and dials into the main train code will make it hard to maintain.
```
nathan-weinberg, line 207:
```suggestion
The other alternative is to keep the same train and generate commands and instead add a `--backend` or `-hifi` flag to trigger the high fidelity code. The issue here is that ilab train is already overloaded with pytorch, mlx, etc. Adding more switches and dials into the main train code will make it hard to maintain.
```

nathan-weinberg (changes requested):

russellb:
dropping `hold` because #48 merged

markstur (commented):
My initial impression was concern about using sub-sub-commands instead of flags.  I think that is explained well.  So... it makes sense to introduce a new level of commands when they affect what sets of flags are relevant.  ✅ 

But my concern now (given some doubts about click) is do we have some PoC to reassure that we have good ability to implement this w/ usability.  E.g. click support, usable default help.  It would just be nice to hear we have that confidence.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
+# Introduce Commands that Run Jobs with Different Fidelity levels for Key ilab functions
+
+This document describes adding different data generation, mixing, and training backends to ilab to enable higher fidelity training using the backend code.
```
russellb, line 3:
```suggestion
This document describes adding different data generation, mixing, and training backends to `ilab` to enable higher fidelity training using the backend code.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+Currently all training is done via qlora or the like. Adding the following commands will enable higher fidelity training and introduce commands such as data mixing.
+
+## Key Component
+
+### Building off of the InstructLab Structure Redesign
+
+After github.com/instructlab/instructlab/pull/990 is merged, ilab will use a parent -> child command structure. This proposal operates under that new structure.
```
russellb, line 11:
```suggestion
After <https://github.com/instructlab/instructlab/pull/990> is merged, `ilab` will use a parent -> child command structure. This proposal operates under that new structure.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+```console
+ilab
+|
+|_______model
+|       |
+|       |____convert
+|       |____download
+|       |____train (--convert)
+|       |____serve   (-i)
```
russellb, line 23:
```suggestion
|       |____serve
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+ilab
+|
+|_______model
+|       |
+|       |____convert
+|       |____download
+|       |____train (--convert)
+|       |    |
+|       |    |______integrated *
+|       |    |______phased *
```
russellb, line 54: I don't understand what `integrated` means here. It seems to imply that other types would not be integrated. ... but integrated with what?

Some other ideas:

- `simple` and `full`
- `fast` and `full`

I know, naming is hard ...

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+|       |
+|       |____serve   (-i)
```
russellb, line 56:
```suggestion
|       |____serve
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+|       |____chat
+|       |____inference
+|       |
+|_______data
+|       |____generate
+|       |    |
+|       |    |______lofi * (name pending)
+|       |    |______hifi * (name pending)
```
russellb, line 64: seems like we should be able to land on consistent names since the pair of options is similar in spirit to the train case.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+|       |____init
+|_______tax
+|       |
+|       |____diff
+|       |____check
+|       |____download
+```
+
+
+The starred commands are the new ones under the redesigned structure. We are opting to add integrated and phased as commands under train and lofi/higi under generate rather than flags given their completely different functionality and backend structure.
```
russellb, line 84:
```suggestion
The starred commands are the new ones under the redesigned structure. We are opting to add integrated and phased as commands under train and lofi/hifi under generate rather than flags given their completely different functionality and backend structure.
```

docs/lofi-hifi-backends.md
russellb: You don't connect to libraries, so how about:

```suggestion
These commands would make use of new InstructLab libraries. The lower fidelity commands will be the equivalent of what currently exists for generate and train in the existing CLI code - QLoRA, with pytorch, mlx, etc. 
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+|       |____chat
+|       |____inference
+|       |
+|_______data
+|       |____generate
+|       |    |
+|       |    |______lofi * (name pending)
+|       |    |______hifi * (name pending)
+|       |
+|       |____mix *
```
russellb, line 66: Do you have more detail about why `data mix` is separate from `data generate` ? I expected it to be included in the `generate` step.

Is there a clear case for running mixing separately?

@oindrillac @aakankshaduggal

docs/lofi-hifi-backends.md
russellb, line 90: This assumes a bit of knowledge of code that most people can't see. I'm not sure that referencing that code is important. Could you focus on the proposal?  Does it come down to -- "we propose integrating additional training code into the instructlab repository that will be run directly as part of the `ilab` process" ?

docs/lofi-hifi-backends.md
russellb, line 92: > High Fidelity can run

How about calling out the exact command like: `ilab train phased` can run ...

> In a more powerful system, the user can also run it in a container,

Why is a more powerful system required to be able to run `ilab` in a container? That seems doable on any level of system.

> even distribute the workload accorss machines using torch.distributed

Is the proposal that this is automatic? Or are you suggesting there is an optional action a user could take?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
+# Introduce Commands that Run Jobs with Different Fidelity levels for Key ilab functions
+
+This document describes adding different data generation, mixing, and training backends to ilab to enable higher fidelity training using the backend code.
+
+Currently all training is done via qlora or the like. Adding the following commands will enable higher fidelity training and introduce commands such as data mixing.
+
```
russellb, line 6: A major part of this proposal is integrating phased training, though it doesn't really explain anywhere what that means. It would be helpful in the introduction to recap the current state of training support and then introduce the major new training capabilities proposed here. You can refer back to the original Lab paper for a more detailed description.

https://arxiv.org/html/2403.01081v3

The same issue applies to data generation. The doc doesn't really explain why `lofi/hifi` commands are necessary. An overview of the current state, plus an overview of what's proposed to be added, along with why it needs to be a separate command, would be helpful.

If you don't want to get into all of that for the data generation side, you could focus this proposal on training only.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+### ilab model train integrated
+
+
+This command would take something like the following arguments
+
+
+--gpus=str , describes the amount of GPUs (of what is available) to use for this process. This comes in the form of: 0-1, 8, etc.
```
russellb, line 110: How about a default to use all?  `-1` perhaps

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+--quantize=bool, enabled Qlora which basically loads the model in a quantized form so it can fit on a consumer GPU
```
russellb, line 111:
```suggestion
--quantize=bool, enable Qlora, which loads the model in a quantized form so it can fit on a consumer GPU
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+--output-dir=str, where to put the model after training
+--num-epochs=int, the amount of epochs to run in this phase of training
+--device=str, the accelerator to use: cuda,rocm,cpu,mlx,mps
+++ many of the current train flags probably
+
+#### Implementation Specifics
+
+The Transformers library (which is currently what we use for training along with torch) has options for fsdp, deepspeed, and many of the related arguments currently in use elsewhere in the project. My idea here is to try and use these as best as we can in the CLI rather than make our own custom versions that we then have to maintain.
+
+There might be a usecase for kicking off some custom deepspeed code for more in depth inter-checkpoint eval. That would be the point of the `phased` command. The phased train command is meant to be run in conjunction with `ilab checkpoint evaluate` which will give the user the best checkpoint to run the next phase on.
```
russellb, line 126:
```suggestion
There might be a use case for kicking off some custom deepspeed code for a more in-depth inter-checkpoint eval. That would be the point of the `phased` command. The phased train command is meant to be run in conjunction with `ilab checkpoint evaluate`, which will give the user the best checkpoint to run the next phase on. 
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+--device=str, the accelerator to use: cuda,rocm,cpu,mlx,mps
+++ many of the current train flags probably
+
+#### Implementation Specifics
+
+The Transformers library (which is currently what we use for training along with torch) has options for fsdp, deepspeed, and many of the related arguments currently in use elsewhere in the project. My idea here is to try and use these as best as we can in the CLI rather than make our own custom versions that we then have to maintain.
+
+There might be a usecase for kicking off some custom deepspeed code for more in depth inter-checkpoint eval. That would be the point of the `phased` command. The phased train command is meant to be run in conjunction with `ilab checkpoint evaluate` which will give the user the best checkpoint to run the next phase on. 
+
+Keep in mind though, this is the community CLI! I feel as though we should try to find a middle ground between server usecases and community usecases. Having the default path in the InstructLab CLI be torch+transformers makes sense for the following usecases:
```
russellb, line 128: Community and Server are not separate things. Community is inclusive of all use cases.

```suggestion
 I feel as though we should try to find a middle ground between server use cases and laptop use cases. Having the default path in the InstructLab CLI be torch+transformers makes sense for the following use cases:
```

docs/lofi-hifi-backends.md
russellb, line 149: similar comment about the use of "community" here -- that's not a use case differentiator

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+
+### ilab model train phased
+
+This command would take roughly the following arguments
+
+--optimizer=str (deepspeed, fsdp) describes the optimizer framework to use during training
+--device=str , accelerator to use cpu,cuda,rocm,mlx,mps
+--model-dir=path , dir where the model to be used in this phase is located
+--data-dir=path , dir where the data for this phase is located
```
russellb, line 147: There are a lot more options in the previous train command. There are some that still seem relevant like input-dir, model-name, output-dir

docs/lofi-hifi-backends.md
russellb, line 6: Another thing that would be helpful up top is to define "high fidelity" more clearly up front. It's used to describe a use case, but that use case isn't really clear from reading the doc.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+### ilab data mix
+
+This command would take something like the following arguments
+
+--num-util-proc=int
+--output-dir=str (defaults to generated/mixed)
+--knowledge-recipes=[]str (path to yaml)
+--skill-recipes=[]str (path to yaml)
+
+* Do we need an `ilab recipe` cmd? *
```
russellb, line 174: What does this mean?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+The big advantage here is more "fine tuned" training than currently exists in the CLI because of deepspeed (or fsdp). The user could even set this up for multi GPU or multi system support with future ilab enhancements.
+
+### ilab model train phased
+
+A user would run something like (assuming phase00 has run) on a GPU enabled server.
+
+`ilab model train phased --device=cuda --model-dir=./phase00/model --data-dir=./phase00/data`
+`ilab checkpoint evaluate ./phase05/checkpoints --output-dir=./phase10`
+`ilab model train phased --device=cuda --model-dir=./phase10/model --data-dir=./phase10/data`
+....
```
russellb, line 201: This is the only place in the doc where `phase00`, `phase05`, and `phase10` are mentioned. I think these could use a more detailed definition of their meaning.

russellb (changes requested):

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,212 @@
...
+
+There might be a usecase for kicking off some custom deepspeed code for more in depth inter-checkpoint eval. That would be the point of the `phased` command. The phased train command is meant to be run in conjunction with `ilab checkpoint evaluate` which will give the user the best checkpoint to run the next phase on. 
+
+Keep in mind though, this is the community CLI! I feel as though we should try to find a middle ground between server usecases and community usecases. Having the default path in the InstructLab CLI be torch+transformers makes sense for the following usecases:
+
+1. Developer with a Gaming PC:
+    * Transformers+Pytorch support Qlora&&FSDP. While deepspeed might be a more "server-rack" use-case, having multi-phase training in the CLI for anyone with a consumer GPU makes sense.
+2. Someone interested in ML, has a Homelab, or *anything with 2 GPUs*
```
hickeyma, line 132: What about the basic consumer laptop with GPU like Mac M-Series?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+```console
+ilab
+|
+|_______model
+|       |
+|       |____convert
+|       |____download
+|       |____train (--convert)
+|       |    |
+|       |    |______peft *
```
leseb, line 55: Ditto on the positional argument.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+`ilab train phased` would utlize much of the backend training code. Primarily doing the following:
+
+- running a single phase of training
+- outputting checkpoint dirs for evaluation to be run on
+- deepspeed support
+- fsdp support.
+
+`ilab train phased` will be used in conjunction with `ilab checkpoint evaluate`. This command will run full scale inter-checkpoint evaluation on the given directory. An output dir will then hold then best checkpoint and all necessary data to run another `ilab train phased` command on.
+
+Much of the code for `ilab train phased` and `ilab train peft` will live in the CLI, and hopefully will be shared. However, plans to move much of this code to an external training library should eventually happen. This is most likely not in scope for the first version of this work.
```
leseb, line 104: Is it still the case? Looks like the training repo is around the corner  https://github.com/instructlab/dev-docs/pull/85

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+### Reasoning
+
+Plugging into hardware acceleration and multi-phase training is the logical next step for ilab. Ensuring we do this in a clean way that does not overload our current commands is also crucial. Many of the processes in the backend are confusing so we want to abstract some of the steps away from users while also giving them a reasonable amount of choice in configuring these new processes. However, maintaining the current laptop story is important to users without hardware access. Splitting these two paths into separate commands maintains the integrity of each.
+
+Not only maintaining the current laptop story but enhancing it. Qlora by itself is great, but having LoRA based training with the ability to run ZeRO phaes 1-3 or full FSDP accleration given the right hardware will improve the experience for people on 1 or 2 GPU systems (which are quite common).
```
leseb, line 110:
```suggestion
Not only maintaining the current laptop story but enhancing it. Qlora by itself is great, but having LoRA based training with the ability to run ZeRO phases 1-3 or full FSDP acceleration given the right hardware will improve the experience for people on 1 or 2 GPU systems.
```

leseb: How do we know that "people on 1 or 2 GPU systems" is common? If we can't easily prove it, I'd suggest to remove it. It's not adding much to your point that the laptop story is important to maintain and enhance.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+--input-dir=str, describes where the generated+mixed data lives to be trained on.
+--model-name=str, name of the model to be output
+--output-dir=str, where to put the model after training
+--num-epochs=int, the amount of epochs to run in this phase of training
+--device=str, the device type to load the model onto: cuda,rocm,cpu,mlx,mps
+++ many of the current train flags probably
+
+#### Implementation Specifics
+
+The Transformers library (which is currently what we use for training along with torch) has options for fsdp, deepspeed, and many of the related arguments currently in use elsewhere in the project. My idea here is to try and use these as best as we can in the CLI rather than make our own custom versions that we then have to maintain.
```
leseb, line 135:
```suggestion
The Transformers library (which is currently what we use for training along with torch) has options for fsdp, deepspeed, and many of the related arguments currently in use elsewhere in the project. The goal is to try and use these as best as we can in the CLI rather than make our own custom versions that we then have to maintain.
```

docs/lofi-hifi-backends.md
leseb: I think it's best to avoid using `I`, so this will feel more like a general document.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+Keep in mind though, this is the community CLI! I feel as though we should try to find a middle ground between server usecases and community usecases. Having the default path in the InstructLab CLI be torch+transformers makes sense for the following usecases:
+
+1. Developer with a Gaming PC:
+    * Transformers+Pytorch support Qlora&&FSDP. While deepspeed might be a more "server-rack" use-case, having multi-phase training in the CLI for anyone with a consumer GPU makes sense.
+2. Someone interested in ML, has a Homelab, or *anything with 2 GPUs*
+    * Trandformers+Pytorch supports Deepspeed on a single system spreading the training over the GPUs. Any professional or Hobbyist that has 2 GPUs will be looking for this option
+3. The laptop usecase
+    * Maintianing Qlora as the performant training mode for the laptop is crucial as most people cannot handle the full models. However, unlocking some better results by using FSDP+Qlora could improve local results and get people more interested in InstructLab.
```
leseb, line 146:
```suggestion
    * Maintaining Qlora as the performant training mode for the laptop is crucial as most people cannot handle the full models. However, unlocking some better results by using FSDP+Qlora could improve local results and get people more interested in InstructLab.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+The transformers library supports the following combinations:
+
+1. LoRA + BitsNBytes Quantized Model + Deepspeed (QLoRA + DeepSpeed)
+    - someone with a consumer GPU could get much better training results from using this combo as opposed to torch without deespeed and inter-checkpoint eval
+2. Lora + FSDP (untested by me)
```
leseb, line 152:
```suggestion
2. Lora + FSDP
```

If the transformers library claim the support, let's keep it that way.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+3. LoRA + DeepSpeed
+    - This enables native multi-GPU support out of the box. Model Loading, trainer setup, and deepspeed initialization are all handeled by transformers. I have tested this and it works great on something like 2x4090 (3090 too) or 2xA10
```
leseb, line 154: Can you expend on what "works great" exactly?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+However we also still want to support something like LoRA + no acceleration besdies using something like CUDA, MPS, or MLX. All of the above would be triggered by flags like `--accelerator`, `--device`, etc.
```
leseb, line 156:
```suggestion
However we also still want to support something like LoRA + no acceleration besides using something like CUDA, MPS, or MLX. All of the above would be triggered by flags like `--accelerator`, `--device`, etc.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+There might be a usecase for kicking off some custom deepspeed code for more in depth inter-checkpoint eval. That would be the point of the `phased` command. The phased train command is meant to be run in conjunction with `ilab checkpoint evaluate` which will give the user the best checkpoint to run the next phase on. 
+
+Keep in mind though, this is the community CLI! I feel as though we should try to find a middle ground between server usecases and community usecases. Having the default path in the InstructLab CLI be torch+transformers makes sense for the following usecases:
+
+1. Developer with a Gaming PC:
+    * Transformers+Pytorch support Qlora&&FSDP. While deepspeed might be a more "server-rack" use-case, having multi-phase training in the CLI for anyone with a consumer GPU makes sense.
+2. Someone interested in ML, has a Homelab, or *anything with 2 GPUs*
+    * Trandformers+Pytorch supports Deepspeed on a single system spreading the training over the GPUs. Any professional or Hobbyist that has 2 GPUs will be looking for this option
```
leseb, line 144:
```suggestion
    * Transformers+Pytorch supports Deepspeed on a single system spreading the training over the GPUs. Any professional or Hobbyist that has 2 GPUs will be looking for this option
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+transformers lets you put "auto" for any field that has a corresponding `TrainingArguments` option. For example the dtype of the model can be passed into the DS config using `bf16: auto` and `fp16: auto`. 
+
+for `ilab model train peft` the DS json config will:
+- use ZeRO phase 2
+- offload to CPU, since NVME offloading is not guaranteed.
+- use a train batch size of 48
+- Set FP16 and BF16 to auto, whichever one the model's dtype defaults to will be auto populated into the config.
+- optimizer is AdamW with the learning rate and all other params set to auto, these are most commonly set by transformers, especially if you pass in one of pytorch's optimizers to the training arguments.
+- scheduler is WarmupLR with all params set to auto and handeled by transformers.
```
leseb, line 168:
```suggestion
- scheduler is WarmupLR with all params set to auto and handled by transformers.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+
+Using Phase 3 is unecessary here as that deals with larger systems with more data spread over more GPUs, it is also typically slower than Phase 2 due to the sharding of data.
+
+The key pieces of information here is the cpu offloading, the utilization of phase 2, and defaulting to "auto" for most other settings that are not directly related to performance uplifts for single or dual GPU support.
+
+
+### ilab model train phased
+
+This command would take roughly the following arguments
```
leseb, line 177:
```suggestion
This command would take the following arguments
```

It's the design phase, so let's be more precise.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+--model-dir=path , dir where the model to be used in this phase is located
+--data-dir=path , dir where the data for this phase is located
+
+Note there is no LoRA in this command and there is no quantization. The `ilab model train peft` command will use the transformers library with pytorch. This is because those have awesome plugins for deepspeed and fsdp WITH lora and qlora. Those are absolute necessities for community usecase. However, they will be mostly unused in the "High Fidelity" usecase.
+
+`ilab model train phased` will opt to use torchrun under the hood, stepping away from the transformers library a bit. Though, research in future versions will most likely lead to a convergance between the two since transformers paired with torch can do most of the tasks we manually shell out to torchrun and torch.distriuted for. 
+
+The DeepSpeed config currently used for `ilab model train phased` will have more hardcoded values, until flags are added in later versions allowing users to configure things like the trianing batch size, offloading optimizer etc. with PEFT most of that is handeled using the `auto` keyword that transformers can funnel defaults into.
```
leseb, line 188:
```suggestion
The DeepSpeed config currently used for `ilab model train phased` will have more hardcoded values, until flags are added in later versions allowing users to configure things like the training batch size, offloading optimizer etc. with PEFT most of that is handled using the `auto` keyword that transformers can funnel defaults into.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+`ilab model train peft --device=cuda --accelerator=deepspeed --num-epochs=5`
+
+This would 
+1. load the model in its full form to the GPUs vram (multiple GPUs in this case)
+2. setup the transformers trainer with this model, and with a hardcoded deepspeed config ilab would come with
+3. train for 5 epochs using AdamW as the optim and deepspeed on top of that.
+4. give you a model in safetensors format or GGUF format since this is not a bitsnbytes model.
+
+
+The big advantage here is faster and higher fidelity training than currently exists in the CLI because of deepspeed (or fsdp). The user could even set this up for multi GPU or multi system support with future ilab enhancements.
```
leseb, line 245: The benefits from this approach look similar to the ones when using `--quantize`, can we add more on the gains?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,279 @@
...
+Basically they would run phased with an eval in between. The Eval looks at the checkpoints output by the previous phase and outputs a model dir in the next phase's working directory. 
+
+
+### ilab model evaluate full AND ilab model evaluate checkpoint
+
+This command will take something like the following flags:
+
+--benchmarks [mmlu, mt, pr-mmlu, pr-bench (mt-pr)] default: (mmlu)
+    - you can only run certain benchmarks depending on what type of evaluation you are doing. 
+Note: We could have `ilab model evaluate` as a single command and take flags that depend on eachother like `--checkpoint-dir` and `--benchmarks` but in general, with the new CLI design we are trying to get out of the habit of flags that depend on eachother.
```
leseb, line 266:
```suggestion
Note: We could have `ilab model evaluate` as a single command and take flags that depend on each other like `--checkpoint-dir` and `--benchmarks` but in general, with the new CLI design we are trying to get out of the habit of flags that depend on each other.
```

leseb (changes requested):
How does this proposal fit with https://github.com/instructlab/dev-docs/pull/85? Should the new code go into the new training lib? If yes, this should be mentioned. Same question for the newly added sdg repo?

xukai92 (commented):
i'm a bit concerned about the proposed way to have a positional arg to switch between the current CLI behavior (LoRA) via `peft` and the intended backend via `phased`.
the main reason is that they are orthogonal concepts that could be potentially use together plus `peft` is a more loaded concept (i.e. LoRA is just one example of PEFT).
- for the first, it's possible to have LoRA used together in the LAB phased training
- for the second, we would like to support more than LoRA in the near future. also the term `phased` is a bit ambiguous (not only LAB uses phased training)
perhaps a simple way to move forward on this is to make them options, i.e.
```
ilab model train --peft lora # default being none
ilab model train --method lab # default being sft
```
besides, again from what discussed briefly in the kick-off call, i would think the full backend being a series of calls like below but with actual functions (that `ilab model train --method lab` would be an alias for)
```
ilab model train --data phase-00
ilab model evaluate mmlu --checkpoint-dir ckpt/phase-00
ilab model train --data phase-05 --checkpoint ckpt/phase-00/best
ilab model evaluate mmlu --checkpoint-dir ckpt/phase-05
ilab model train --data phase-10  --checkpoint ckpt/phase-05/best
ilab model evaluate mtb --checkpoint-dir ckpt/phase-10
```
and the best model would be located at `ckpt/phase-10/best`

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: leaving this one open even though this part of the doc is gone, I was under the assumption that mixing would be its own command if not then I am fine with removing all mentions of it in the doc

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

docs/lofi-hifi-backends.md
cdoern: no longer relevant due to doc overhaul, resolving

though, I think in order to auto assign a profile in the future we will need to somewhere check hardware the user has?

docs/lofi-hifi-backends.md
cdoern: its a str, so `all` probably

cdoern: there is some standard notation for this sort of GPU referencing, let me look for that

docs/lofi-hifi-backends.md
cdoern: removed, this was an old thought

cdoern:
Ok, I resolves a bunch of outdated comments since this EP just got a major overhaul.

After talking with teams designing new instructlab libraries the only way I can see to reconcile ALL of them while also maintaining the current CLI usecases is to provide top level "profiles". The team has kicked this idea around before from having a "prius" profile to a "F1 racecar" profile. For immediate purposes, I would create a way to set a profile out of three-ish hardcoded ones the training team (@RobotSail @aldopareja @JamesKunstle @Maxusmusti) will provide me and then in the future we can setup an interactive way to make a custom profile based off of these defaults.

cdoern:
This may seem like a pretty 180 degree shift. However, since the main point of this doc was ilab train commands, the training team is designing the library as a single entrypoint which takes a config file of options. With this being the case, the majority of this doc became outdated.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
+# Introduce Instructlab profiles managed via `ilab profile...` to run different key commands at different Fidelity levels
+
+This document describes adding different data generation, mixing, and training backends to `ilab` to enable higher fidelity training using the backend code. By higher fidelity we mean models that preform better, were trained on better hardware, off of larger data sets and using more intensive training techniques.
```
russellb, line 3:
```suggestion
This document describes adding different data generation, mixing, and training backends to `ilab` to enable higher-fidelity training using the backend code. By higher fidelity, we mean models that perform better, were trained on better hardware, off of larger data sets, and used more intensive training techniques.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+Rather than augmenting the existing commands to have conflicting and confusing flags that the team will be forced to maintain, deprecate, and add onto, I am proposing we open the door to these new workflows via instructlab "profiles" and the `ilab profile` command group.
```
russellb, line 5:
```suggestion
Rather than augmenting the existing commands to have conflicting and confusing flags that the team will be forced to maintain, deprecate, and add onto, I am proposing we open the door to these new workflows via instructlab "profiles" and the `ilab profile` command group.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
+# Introduce Instructlab profiles managed via `ilab profile...` to run different key commands at different Fidelity levels
```
russellb, line 1: I think it's worth changing the filename to reflect the latest approach.

docs/lofi-hifi-backends.md
russellb: Can we scope this down to training only?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+This document describes adding different data generation, mixing, and training backends to `ilab` to enable higher fidelity training using the backend code. By higher fidelity we mean models that preform better, were trained on better hardware, off of larger data sets and using more intensive training techniques.
+
+Rather than augmenting the existing commands to have conflicting and confusing flags that the team will be forced to maintain, deprecate, and add onto, I am proposing we open the door to these new workflows via instructlab "profiles" and the `ilab profile` command group. 
+
+Currently all training is done via QLoRA or the like. Adding the following commands will enable higher fidelity training and introduce commands such as data mixing.
```
russellb, line 7: > or the like

Is it just QLoRA?  Is it more?  If it's more, can we name it?

I'd like to request dropping the data mixing part and let's just focus on training.  Containing the scope will make it easier to get this done.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+This document focuses primarily on training, specifically different configuration types or "profiles" for `ilab model train`.
```
russellb, line 9: Ah ha - here you clarify the doc really is intended to be focused on training. I think removing earlier references to other topics would help.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+## Key Component
+
+### Building off of the InstructLab Structure Redesign
+
+After <https://github.com/instructlab/instructlab/pull/990> is merged, ilab will use a parent -> child command structure. This proposal operates under that new structure.
```
russellb, line 15: Just a style thing - when referring to a literal command, wrap it in backticks

```suggestion
After <https://github.com/instructlab/instructlab/pull/990> is merged, `ilab` will use a parent -> child command structure. This proposal operates under that new structure.
```

docs/lofi-hifi-backends.md
russellb: This first line reads more like a summary than a title. I would shorten it to something like:
```suggestion
# InstructLab Training Profiles
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+`ilab model train` will still be a single command offered to users. However, new "profiles" offered by the CLI will enable users to toggle what mode they are running instructlab in. These profiles will initially just have global options for training. Rather than adding dozens of flags to the CLI, storing them in "pre-baked" configs is a good first step.
```
russellb, line 17:
```suggestion
`ilab model train` will still be a single command offered to users. However, new "profiles" offered by the CLI will enable users to toggle what mode they are running `ilab model train` in. These profiles will initially just have global options for training. Rather than adding dozens of flags to the CLI, storing them in "pre-baked" configs is a good first step.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+Eventually, we will want to enable a `ilab profile init` command that will allow users to initialize a custom profile based off of one of the pre-baked ones. This is not targeted for an upcoming release though.
```
russellb, line 19: This seems more like a configuration thing managed under `ilab config` if it will have a CLI interface for editing.

russellb:
```suggestion
Eventually, we will want to enable an `ilab profile init` command that allows users to initialize a custom profile based on one of the pre-baked ones. This is not targeted for an upcoming release, though.
```

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+### Building off of the InstructLab Structure Redesign
+
+After <https://github.com/instructlab/instructlab/pull/990> is merged, ilab will use a parent -> child command structure. This proposal operates under that new structure.
+
+`ilab model train` will still be a single command offered to users. However, new "profiles" offered by the CLI will enable users to toggle what mode they are running instructlab in. These profiles will initially just have global options for training. Rather than adding dozens of flags to the CLI, storing them in "pre-baked" configs is a good first step.
+
+Eventually, we will want to enable a `ilab profile init` command that will allow users to initialize a custom profile based off of one of the pre-baked ones. This is not targeted for an upcoming release though. 
+
+### Immediate Goals
```
russellb, line 21: Could you add a new section called something like "UX Overview" that walks through what the experience would look like to a user?

I know the commands are scattered throughout the doc, but there's a lot of discussion around each one. A focused look at the UX (a tl;dr of the commands) would help.

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+For the near future, a single upper level profile that can be initialized via `ilab profile set <profile_name>`
```
russellb, line 23: This reads like a sentence that hasn't been completed

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+profile:
+    train:
+        gpus: 0-4
+        epochs: 15
+        accelerator: deepspeed
+        ds_config: foobar.json
+```
+Eventually this profile would have settings for generation, eval, etc. But for immediate goals, hardcoded training settings is the MVP. Rather than having a `--config` option at the `ilab model train` level, storing the profile at the global level allows us to expand this idea to other ilab commands in the future. We need to be careful about how we introduce new concepts like this.
+
+For immediate releases I would introduce the idea of a profile, a command to set a specific profile, and hardcoded profiles that plug into key ilab commands namely training.
```
russellb, line 37: This idea that a profile is not specific to training seems like a critical detail of this proposal.

My expectation was something very focused on training.

A global profile for everything is just the existing configuration file. You can have multiple configuration files if needed. I'm having a hard time deciding how this would end up being significantly different.

I think you're proposing a set of options that are supported in the config file but NOT via command line arguments. Is that right?

Honestly I think I'd rather just have all the command line arguments PLUS an extra one `--profile` that's shorthand for a set of defaults.

russellb (changes requested):
My biggest request after a first look is to focus the proposal on training. That is going to make it easier to get to a proposal that can get consensus.

docs/lofi-hifi-backends.md
cdoern: also, this is the design doc for friday's design freeze so it needs to consider all commands being added

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+This command would take something like the following arguments
+
+--num-util-proc=int
+--output-dir=str (defaults to generated/mixed)
+--knowledge-recipes=[]str (path to yaml)
+--skill-recipes=[]str (path to yaml)
+
+Similar to data generation and model training, performance related flags should be baked into a profile if any exist.
+
+### ilab model evaluate full AND ilab model evaluate checkpoint
```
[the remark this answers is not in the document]
danmcp: There would be a difference between checkpoint and basic model eval from an input and output perspective though.  Checkpoint eval takes a directory of models and outputs a winner.  Model eval takes a single model and outputs the scores.  Perhaps that would be a better way to focus the experience.  eval and compare vs eval.

danmcp (changes requested):

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,193 @@
...
+
+### `ilab model evaluate full` AND `ilab model evaluate checkpoint`
+
+instructlab will run MMLU bench and MT bench in upcoming releases. These benchmarks are useful for inter-checkpoint evaluation and full scale evaluation at the end of multi-phase training.
+
+MMLU bench needs the following options:
+- --model
+- --tasks
+- --few-shots (int)
+- --batch-size
```
danmcp, line 135: Should this be tuned based off #gpus in config rather than a separate param?

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+ilab model train will maintain all of its current flags which will represent overrides for the values stored in the training section of a user profile. This means all of the existing flags will be ported to a default config with sane values set for all that are applicable. Some flags that change frequently like `--data-dir` and `--model-dir` for phased training will be added to the training command.
+
+The existing code which runs the HF trainer will be ported to another library which can also run multi-phase training. This library will be shelled out to using torchrun.
+
+some new entries in the config file that will not be exposed as user level flags are:
+
+- accelerator=str (deepspeed, fsdp) describes the optimizer framework to use during training
+- gpus=str describes the amount of GPUs (of what is available) to use for this process. This comes in the form of: 0-1, 8, etc.
+- ds_config=str desceibes a path to a .json file configuring deepspeed.
```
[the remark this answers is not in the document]
cdoern: sure!

docs/lofi-hifi-backends.md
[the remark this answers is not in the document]
cdoern: I switched this to a single cmd that just takes bench info

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+--knowledge-recipes=[]str (path to yaml)
+--skill-recipes=[]str (path to yaml)
+
+Similar to data generation and model training, performance related flags should be baked into a profile if any exist.
+
+### ilab model evaluate full AND ilab model evaluate checkpoint
+
+This command will take something like the following flags:
+
+--benchmarks [mmlu, mt, pr-mmlu, pr-bench (mt-pr)] default: (mmlu)
```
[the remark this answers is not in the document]
cdoern: good point, no default it is

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+    - you can only run certain benchmarks depending on what type of evaluation you are doing. 
+Note: We could have `ilab model evaluate` as a single command and take flags that depend on each other like `--checkpoint-dir` and `--benchmarks` but in general, with the new CLI design we are trying to get out of the habit of flags that depend on each other.
+--output-dir: str, determines where the best checkpoint is put for the next phase of training
+--input-dir: str, takes the directory of the model/checkpoint to evaluate.
```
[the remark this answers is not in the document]
cdoern: unsure, @alimaredia @alinaryan @nathan-weinberg what is the format of passing a model vs checkpoint dir in the evaluation library? I have never run eval

docs/lofi-hifi-backends.md
```diff
@@ -0,0 +1,207 @@
...
+
+For immediate releases I would introduce the idea of a profile, a command to set a specific profile, and hardcoded profiles that plug into key ilab commands namely training.
+
+### Reasoning
+
+The underlying training, eval, and generation libraries will handle the specifics based off of the config provided via the profile. For example is a user passes CPU training, no Deepspeed/FSDP etc then the training library will run the equivalent of "linux_train" that currently exists, outputting a model ready to be used. If the user has 4 GPUS, Deepspeed enabled and 15 epochs, the training library might give you a series of checkpoints.
+
+`ilab checkpoint evaluate` will be used in conjunction with `ilab model train` when the user is running multi-phase training. This command will run full scale inter-checkpoint evaluation on the given directory. An output dir will then hold then best checkpoint and all necessary data to run another `ilab train phased` command on.
```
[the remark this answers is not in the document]
cdoern: I think we agreed on no orchestration, but I could be convinced to add a `--eval` flag to training.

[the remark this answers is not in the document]
danmcp: I am not sure what would make that be considered orchestration.  I wouldn't have said it's simply something that takes multiple steps.  Training is already multiple steps per phase.  Eval of a checkpoint dir is also multiple steps.  Generally I think of orchestration as involving more complex workflows or coordinating parallel processes.  The question I have here is what's the desired input/output flow from a user perspective.  Specifically, what's the reason(s) a user would like a separation between the train and eval of each phase?

docs/lofi-hifi-backends.md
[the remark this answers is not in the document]
danmcp: The current design:

https://github.com/instructlab/eval/pull/6/files

Is eval would take a single model.  So for a checkpoint the caller would need to look through the checkpoints and pass each model 1x1.

docs/lofi-hifi-backends.md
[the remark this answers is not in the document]
danmcp: One more potential thing to think about here. If we ever want to point ilab at a remote API to perform these actions, would that change the answer for what the desired user experience would be?

docs/ilab-profile.md
```diff
@@ -0,0 +1,392 @@
+# Introduce Functionality to Utilize the InstructLab backend: `ilab` Profiles, and Command Adaptations
+
+This document describes adding different data generation, mixing, and training configurations to the `ilab` CLI to enable higher-fidelity training using the backend code. By higher-fidelity, we mean models that perform better, were trained on better hardware, off of larger data sets, and used more intensive training techniques.
```
hickeyma, line 3:
```suggestion
InstructLab is a workflow for a model alignment technique for adding contributions directly to Large language Models (LLMs) using multi-phased alignment training. The Instruct Lab CLI (`ilab`) currently hard codes the different capabilities (e.g. data generation, training etc.) of the workflow.

This document describes extending the CLI to make the different capabilities of the workflow to be configurable. This would enable users to configure the CLI based on:
   - Hardware specification of the user. For example, numbers of GPUs, memory etc.
   - Using capabilities which are distributed for the workflow.
```

docs/ilab-profile.md
```diff
@@ -0,0 +1,392 @@
...
+
+Users will be able to set a pre-defined profile type in their config.yaml that would enable sane defaults for higher fidelity training, generation, and evaluation. These defaults will funnel into existing and new flags for the CLI.
```
hickeyma, line 5: Should this be in the what you are trying to do section? This is implementation and more "how" want to do it.

docs/ilab-profile.md
```diff
@@ -0,0 +1,392 @@
...
+
+## `ilab` Profiles
```
hickeyma, line 7:
```suggestion
## Proposed Design
```

hickeyma: Now introduce the profiles concept here, giving an overview like from above.

docs/ilab-profile.md
```diff
@@ -0,0 +1,392 @@
...
+Eventually, we will want to enable `ilab profile init` and `ilab profile set` commands that allow users to initialize a custom profile based on one of the pre-defined ones. This is not targeted for an upcoming release, though.
+
+### Immediate Goals and Core Principles
+
+For the near future, there will be upper level profiles that can be initialized via `ilab config init`.
+
+Profiles should at first be static and not exposed to users. Internally, a profile would look something like:
+
+```yaml
+profile:
```
hickeyma, line 22: Profile should have a `name` attribute to identify it

docs/ilab-profile.md
```diff
@@ -0,0 +1,392 @@
...
+    - mid level instruction number for generation, higher than the default 10 (100-500)
+    - eventual deepspeed support with LoRA and QLoRA in training
+    - 10+ epochs for training
+    - eval on checkpoints if hardware permits (depends on vRam)
+3. **Multi GPU, Generate, Train, Eval profile -- CUDA**
+    - high instruction count, 500 or so for generation
+    - deepspeed support without quantization or LoRA
+    - evaluation on checkpoints after a phase of training
+
+**Profile 3 would most likely have sub-profiles for different GPU support.**
```
hickeyma, line 114: Where is the Mac Metal profile covered?

docs/ilab-profile.md
```diff
@@ -0,0 +1,392 @@
...
+
+No, adding user profiles for generation and evaluation would be ideal. The forthcoming backend libraries are taking an approach where it will be easiest to have pre-baked "config" objects that will feed into them. We need to be careful about how we introduce new concepts like this. Adding lower level intensive configurations to ilab is confusing to community users.
+
+**PLEASE NOTE:**
+
+**For immediate releases I would introduce the idea of a profile, a way to set these via `ilab config init` and hardcoded profiles that plug into key `ilab` commands namely training, generation, and evaluation.**
+
+### Reasoning
+
+The profile settings will be used as arguments for most if not all libraries being introduced to the `ilab` backend.
```
hickeyma, line 61:
```suggestion
The profile settings will be used as arguments for most if not all capabilities that are part of the workflow.
```

hickeyma (changes requested):
Thanks @cdoern for working hard on this and the improvements. I like the direction is is going in now.

I have some feedback inline. It is mostly around:

- What are we trying to solve rather than how
- It is capability of the workflow rather than just getting things from the backend (even if implementation from the backend)
- Mention how you can add customized profile

PR #55, opened 2024-05-20
Knowledge past wikipedia

jjasghar:
We need a plan to get more places then wikipedia submissions. This is the first step to formulate it.


russellb (approved):
lgtm as an official place to list the status of each source

lhawthorn (approved):
LGTM, thank you for the thorough scoping of the problem space and excellent write up.

I can help us track down the right contact from a legal perspective, too.

PR #60, opened 2024-05-22
added link to place where community can comment on data sources

lhawthorn:
This will eventually need to be more streamlined than an open issue, but for now an open issue suffices.

Also clarified wording in the document.

### DIFF
--- a/docs/knowledge-submissions-past-wikipedia.md
+++ b/docs/knowledge-submissions-past-wikipedia.md
@@ -42,7 +42,7 @@ Status:
 ### Next steps
 
 1. We have to find the correct legal person to find a way to be the correct point person for this project.
-1. Collect suggested places from the community and add them to the above table
+1. Collect [suggested data source domains from the community](https://github.com/instructlab/dev-docs/issues/59) and add them to the above table.
 1. Work with our legal team to get approvals and denials.
 1. Inform the triage team and triagers of the new locations we can or can not accept.
 


docs/knowledge-submissions-past-wikipedia.md
```diff
@@ -42,7 +42,7 @@ Status:
 ### Next steps
 
 1. We have to find the correct legal person to find a way to be the correct point person for this project.
-1. Collect suggested places from the community and add them to the above table
+1. Collect [suggested data source domains from the community](https://github.com/instructlab/dev-docs/issues/59) and add them to the above table.
```
bjhargrave, line 45: This implies the issue will be open "forever" as a receiving site for suggestions. Perhaps a github discussion might be better for this role?

lhawthorn: We keep going back and forth on what we plan to use GitHub discussion for vs. issues. I think perhaps the real place to collect this info is in a thread on the community mailing list. Thoughts?

hickeyma (changes requested):
@lhawthorn The current steps in [Process Steps](https://github.com/instructlab/dev-docs/blob/main/docs/knowledge-submissions-past-wikipedia.md#process-steps) seem to cover how to handle data source updates. Do you think that this PR as now documented?

lhawthorn:
Yes, I believe this PR is now out of date. I am closing it. Thank you, Martin.

PR #63, opened 2024-05-22
New repository proposal: sdg

russellb:
This document includes a proposal for a new repository which contains
a Python library focused on Synthetic Data Generation (SDG). This is a
discreet area of functionality that can evolve on its own with its own
group of contributors and maintainers.

Signed-off-by: Russell Bryant <rbryant@redhat.com>

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -103,6 +103,7 @@ safetensors
 Salawu
 SDG
 Sigstore
+sdg
 sexualized
 SHA
 Shivchander
--- a/docs/sdg-repo.md
+++ b/docs/sdg-repo.md
@@ -1,0 +1,41 @@
+# New Repository Proposal: sdg
+
+## Summary
+
+This document proposes a new repository under the `instructlab` GitHub organization:
+
+- `instructlab/sdg`
+
+## Background
+
+The `instructlab/instructlab` repository includes a basic implementation of
+Synthetic Data Generation (SDG). This implementation does not implement the full
+approach as described by the [LAB paper](https://arxiv.org/abs/2403.01081).
+
+We desire to build out a more complete implementation of SDG that is more in
+line with the LAB methodology. We propose a new repository to house this code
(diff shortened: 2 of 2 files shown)


oindrillac (approved):
Thanks!

khaledsulayman (approved):

nathan-weinberg (approved):

nathan-weinberg:
@russellb Can we add a bit about creating a new SDG Triager/Maintainer team, in line with the scheme we have for the CLI teams? These team will be responsible for triaging, maintenance, releases, etc.

Alternatively this can be a proposal to the `community` repo `CONTRIBUTORING_ROLES.md` doc once the new repo is created.

markstur:
I wonder if anyone can come up with a less cryptic name than SDG, but I don't have a good suggestion (data-gen or synth-data... etc don't thrill me)

markstur (approved):
make it so

lhawthorn:
Dumb question - why don't we just call it synthetic-data-generation? It does what it says on the tin.

nathan-weinberg:
Honestly I'm not opposed to @lhawthorn's suggestion or something like `synthetic-datagen`

russellb:
I would prefer a rough mapping to the Python package included in the repo. This proposal was instructlab/sdg for the instructlab-sdg package. For the other name suggestions, a PyPI check to see if it can also be the Python package name would be good. It would also be helpful to know if the name suggestions are brainstorming or a “-1” for the proposed name. 

Naming is hard.

hickeyma (commented):
Nice proposal, thanks @russellb.

One thing should we also include the basic implementation (which is a modified version of self-instruct (https://arxiv.org/abs/2212.10560)) currently in the CLI to be moved to the repository also? That way we extract all SDG implementation from the CLI and we have the community supported implementations in 1 repository.

bjhargrave:
> big +1 to moving away from git submodules for the schema - a package on pypi makes sense to me



> +1 to this as well

OK, I'll start some work towards this end.

russellb:
https://github.com/instructlab/sdg

PR #67, opened 2024-05-29
Incident Management Doc

jjasghar:
This is the first pass at created an incident management process for InstructLab. We will have problems and this is the beginning of the way to communicate them out.

Thanks to @coderanger for the inspiration for this.

### DIFF
--- a/docs/incident-management.md
+++ b/docs/incident-management.md
@@ -1,0 +1,124 @@
+# Incident Management for InstructLab
+
+## Attribution
+
+Before we start, I would like to mention that this document and process is inspired
+by and influenced heavily from [here][gist] originally authored by [coderanger][ranger].
+
+## Problem Statement
+
+With the multiple systems that come together for the open source project InstructLab
+to work, we need a catalogue of major issues and problems so we can track and find
+critical bugs in the open. Building a formalized incident management process is
+the first step in building this knowledge database.
+
+## Proposed Solution
+
+Create a light weight but formalized Incident Management process. Whomever calls
+for the Incident is the initial "Incident Commander" [IC] and becomes the hub of
+the information, starts the conversation in the shared Slack channel
+and has a shared timeline (google doc or the like).
+After "a reasonable" amount of time they can hand off the IC
+to someone else, and so on till the Incident is resolved.
+
+### What counts as an incident?
+
+Anything with community visible negative consequences. In most cases this
+will be an outage or downtime event but some non-outage incidents include severe
+performance degradations and security events.
+
+## When should I file an incident report?
+
+Preferably after the incident is resolved. Or at least don't consider it even remotely
+a priority to do the paperwork in the middle of an ongoing issue. But if possible do
+think about recording timeline information that isn't in Slack or the Mailing list which
+might otherwise be lost, so that you will have them for the report later.
+
+## Incident Template
+
+## [Incident Date and Title]
+
+Incident Commander: [Name of IC]
+
+## Summary
+
+[A quick one or two sentence description of the issue from a high-level view.]
+
+## Duration
+
+[The amount of time of the user impact. This does not include any time spent after restoration of service but before the formal end of the incident.]
+
+## User Impact
+
+[Describe the impact to end users, such as which features or services were unavailable.]
+
+## Timeline
+
+[A timeline of events from the start of the incident until the Incident Commander declares it over. All times should be in UTC.]
+
+* 01:23 - Incident start.
+* 23:34 - Incident concluded.
+
+## Proximate Trigger
+
+[The most direct trigger of incident. Often times this will be a human error such as a code bug that was missed in review or an operational mistake. Our process is blameless and we document our mistakes so that we can learn from them. But try to not turn this into a personal callout, even of yourself.]
+
+## Root Cause
+
+[Root causes are the underlying deeper problem that lead to the incident. For example, if the proximate trigger was a bug missed in review then a root cause might be missing static analysis tooling in CI that could have caught it, or missing code review guidelines. Root causes should never be human error, they are systemic issues that create the conditions for human error to become a problem. Also root causes are sometimes slippery, you can trace the chain of events back infinitely far if you try hard enough but putting "Root Cause: 13 billion years ago a singularity expanded into the Big Bang" is not productive. Look for root causes that help guide our future path rather than documenting every contributing factor for its own sake.]
+
+## Detection
+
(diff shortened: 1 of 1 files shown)


docs/incident-management.md
```diff
@@ -0,0 +1,123 @@
+# Incident Management for InstructLab
+
+## Attribution
+Before we start, I would like to mention that this document and process is inspired
+by and influenced heavily from [here][gist] originally authored by [coderanger][ranger].
```
leseb, line 5: Can't we just add this person as a Co-author to this contribution instead of mentioning this?

docs/incident-management.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Problem Statement
+
+With the multiple systems that come together for the open source project InstructLab
+to work, we need a catalogue of major issues and problems so we can track and find
+critical bugs in the open. Bulding a formalized incident management process is
```
leseb, line 11: s/Bulding/Building/

docs/incident-management.md
```diff
@@ -0,0 +1,123 @@
...
+Create a light weight but formalized Incident Management process. Whomever calls
+for the Incident is the initial "Incident Commander" [IC] and becomes the hub of
+the information, starts the conversation in the shared Slack channel
+and has a shared timeline (google doc or the like).
+After "a reasonable" amount of time they can hand off the IC
+to someone else, and so on till the Incident is resolved.
+
+### What counts as an incident?
+
+Anything with customer or community visible negative consequences. In most cases this
```
leseb, line 25: Customer? This sounds a bit strange given this is an open source project. "Community" seems sufficient.

leseb (commented):
This is the first time I've seen something so formalized for an open source project. It feels very company-oriented, but I might be missing something. Another point is the operational nature of the Incident Commander (IC) role as described. Currently, InstructLab, at least in terms of the CLI, doesn't have any long-running services, so I'm not sure if we would experience outages in the same way. However, I do agree that having a process for post-mortem analysis is valuable for addressing issues. I have mixed feelings about this but want to hear from others as well :). Thanks for the write-up!

jjasghar:
> This is the first time I've seen something so formalized for an open source project. It feels very company-oriented, but I might be missing something. Another point is the operational nature of the Incident Commander (IC) role as described. Currently, InstructLab, at least in terms of the CLI, doesn't have any long-running services, so I'm not sure if we would experience outages in the same way. However, I do agree that having a process for post-mortem analysis is valuable for addressing issues. I have mixed feelings about this but want to hear from others as well :). Thanks for the write-up!

Ah, see, it turns out we do have a _bunch_ of back-end services required for us to run this project. We have a team of backend engineers ensuring the model training and tuning is done. This requires communication between two different groups of people, and we have no formal way to communicate. (hence this process) 
We had a "major" incident, just this last week and when I asked for a some information about it, we noticed we didn't have a plan for these types of situations.

docs/incident-management.md
jjasghar: No idea, I figured that there would be _major_ changes to this document as we go forward. This was just a start and figured it would be worth just "tipping the hat to them."

docs/incident-management.md
```diff
@@ -0,0 +1,123 @@
...
+will be an outage or downtime event but some non-outage incidents include severe
+performance degradations and security events.
```
russellb, line 27: I'm having a hard time understanding what would qualify as an incident under this proposal.  Could you come up with some concrete examples?

It seems to be oriented toward running a service where some level of availability is to be expected. That's not really true here. There's stuff running, but not in service to a public group of some kind.

Before commenting further, it would help if we had a sample set of scenarios to use as context for discussion.

russellb (changes requested):

docs/incident-management.md
russellb: I just nonticed I'm duplicating some discussion between you and @leseb - sorry about that.

> Ah, see, it turns out we do have a bunch of back-end services required for us to run this project. We have a team > of backend engineers ensuring the model training and tuning is done. This requires communication between two > different groups of people, and we have no formal way to communicate. (hence this process)
> We had a "major" incident, just this last week and when I asked for a some information about it, we noticed we didn't have a plan for these types of situations.

but none of this is visible to the public right now, so I'm not sure a public process like this makes sense.

docs/incident-management.md
mairin: agreed

docs/incident-management.md
lhawthorn: I am personally a huge fan of deeply explicit acknowledgements for many reasons, not the least of which is showing people we actually pay attention to the state of the art and best practices other devs/projects are using. JMHO.

docs/incident-management.md
lhawthorn: +1 on community is sufficient. Customers are community members, too.

docs/incident-management.md
lhawthorn: @russellb I take your comment to mean "once our backend services like model training are publicly visible, we should have a process for dealing with incidents"

I agree with you!

However, as we did have a hiccup that caused @jjasghar to open this PR, should we at least have some documented process for dealing with said hiccups amongst the maintainer team? I am sure we have one that is informal and ad hoc, but I'd like to see it written down (for my ignorant self).

I think @jjasghar can maybe just give me a summary of the what happened and how to remediate/deal in this issue. The deep specifics are unlikely to be interesting or useful for GitHub history purposes. :D

russellb: I don't know anything about the example in question. It would probably help to speak about it more concretely instead of in the hypothetical.

hickeyma (commented):
I am in agreement with @leseb and @russellb. Some of my thoughts:

- This seems like a process when you have service(s) that you support. Will the open source community be providing such support?
- The backend services I would think will remain for the most part to run the workflow to perform model alignment and tuning of a new model to be published. Do we envision exposing this externally?
- Processing of taxononmy PRs should be treated like any open source contribution: push, CI/CD, review/re-commit cycle and then merge or reject. Any capabilities run (e.g. SDG, training) for the PR should be treated like CI/CD in all projects - raise a bug if the gate is broken.
- What was the issue(s) that occurred that instigated this proposal? This is important in the context if the proposal is required or not.

nathan-weinberg:
@jjasghar what's the status of this doc?

jjasghar:
Haven't touched this since we created it. When we start hosting some level of service for our community we will need some standard procedure, but I think we can have this sit around for the time being.

PR #69, opened 2024-05-31
Propose a backend working group

russellb:
There is a broad group of people interested in collaborating on the
software needed to scale out the workflow in a clustered environment.
As a first step toward formalizing this work, this document describes
a working group that will collaborate on documents in this repository
to defined high level architecture and design for the related
components to build out this updated backend infrastructure.

Signed-off-by: Russell Bryant <rbryant@redhat.com>

PR #71, opened 2024-05-31
Api definition guidelines

gabe-l-hart:
## Description

This PR introduces a framework for managing `OpenAPI` definitions for `InstructLab APIs` and `Platform APIs`.


russellb:
This looks pretty good to me, thanks!

My main request is that we merge #69 first (should be able to next week) and then we treat this as one of the first artifacts coming out of that working group.

The only change I'd suggest then is to move the `doc/api-definition-guidelines.md` file under `docs/backend/`, but that requires #69 to be in first.

There is a bit of additional context I'd like to capture, but I'm OK not blocking this PR on it. For example, this is written with an architecture in mind about an InstrutLab API and supporting platform APIs.  I'd like to write a high level backend direction doc that would introduce these concepts. I have existing content for this. If feedback on that doc changes anything substantially, we can always adjust these docs too.  It's all a moving target.

russellb:
The `hold` label is just for waiting on #69.

gabe-l-hart:
That sounds good. I have an internal platform ADR about the decision to make REST APIs as the common service layer that we can adapt for the general backend architecture perspective.

russellb:
#69 merged, so I'll drop the `hold`

russellb (dismissed):
I'm happy with this.

I'm still a bit surprised that nobody else has reviewed it though. Maybe we aren't making it visible in enough of the right places?

README.md
```diff
@@ -43,3 +43,7 @@ easily updated in the future as needed. Some options include:
 * [Mermaid](https://github.com/mermaid-js/mermaid#readme)
 * [Excalidraw](https://excalidraw.com/)
 ** Be sure to leave "Embed Scene" turned on when exporting the PNG.
+
+### API Specifications
+
+API definitions use [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/).
```
markstur, line 49: suggest link to the sub-dir api-definitions

```suggestion
[API definitions](./api-definitions) use [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/).
```

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,20 @@
...
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/). For structural and style guidelines, see [api-definitions](../../api-definitions/README.md).
+
+## Where will service API definitions live?
+
+Service API definitions will live in [api-definitions](../../api-definitions) within this repo.
```
leseb, line 20: If we put the API definition in here, we might have to rename this repository, it does not look like `dev-docs` anymore.
I'd personally be in favor of putting the definitions in a dedicated repo or as close to the code as possible (same repo as the code) for a few reasons:

* versioning, branching, and releasing a new API
* CI around validation and generation
* potential build and publishing of SDK (that's typically another repo too)

README.md
```diff
@@ -29,7 +29,7 @@ The rules for merging depend on the type of change in question and its scope of
 
 ## Formatting Guidelines
 
-Design documents should be placed in `docs/`.
+Design documents should be placed in [docs/](./docs). API Specifications should be placed in [api-definitions/](./api-definitions).
```
hickeyma, line 32: Does `api-specifications` as directory name read better than `api-definitions`?

api-definitions/README.md
```diff
@@ -0,0 +1,23 @@
+# API Definitions
```
hickeyma, line 1: Is "API Specification" a better fit?

api-definitions/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+The common abstract interface for the service APIs is defined by [REST APIs](https://www.redhat.com/en/topics/api/what-is-a-rest-api) using [OpenAPI](https://www.openapis.org/). This project is the central home for these API definitions.
+
+## Style Guidelines
+
+* Use `kebab-case` for path elements
+* Use `snake_case` for properties
+* Use `UpperCamelCase` for internal reusable schema names
```
hickeyma, line 9: Would it be helpful to provide links to the case definitions?

docs/backend/api-definitions-guidelines.md
hickeyma: Hmm, interesting @leseb. Should the APIs be defined in a specification repo? Its not REST but OpenTelemetry defines its [specifications](https://github.com/open-telemetry/opentelemetry-specification) in one place.

hickeyma (changes requested):
This is looking good, thanks for pushing this @gabe-l-hart. Some small nits inline and questions around where specifications should reside.

README.md
gabe-l-hart: I think either word works here, so no strong opinion at all

api-definitions/README.md
gabe-l-hart: I was hoping the example usage in backticks would be sufficient. Do you think we need further explanation than that? External links seem overkill to me, but I could expand on the definition. Maybe one point to make is that for path elements in particular, we need to use [ascii](https://www.w3schools.com/tags/ref_urlencode.ASP) to avoid percent encoding.

gabe-l-hart: Updated with more detailed guidelines for each

README.md
gabe-l-hart: Thanks! @markstur can you edit your comment to have `../api-definitions` now that I put this doc under `docs/backend`? Then I can "Sign off and commit" with your name on it

docs/backend/api-definitions-guidelines.md
gabe-l-hart: @leseb Thanks for raising this. I too would prefer a dedicated repo, but thought it might be best to start here and move later. Depending on how quickly this ramps up, it may make more sense to just go straight to the end solution though?

The corresponding thought for me that is still unresolved is whether these API specs should all live under `instructlab` or not. The `platform` layer is going to intentionally NOT be `instructlab` specific so that these APIs can be used independently, but at the moment we don't have a "definition of the canonical AI service platform" name, so I don't know _where_ they should live. Having this question still open is part of why I figured a move was coming eventually and we might as well start small.

Definitely no strong opinions here for how we start, so I'm open to anything!

gabe-l-hart: W.r.t. the relationship between the API specs and the code, I agree that there are really three possible options:

1. The API lives with the implementation
  * Pro: This is the easiest to navigate from a project-centric view
  * Con: It's much harder to keep commonality across different services/projects and it's much harder to treat the API as an abstract definition if we decide to reimplement it (or maintain multiple implementations for different deployment scenarios)
3. The API lives in a central place and is used by-reference in the implementation
  * Pro: Single source of truth, easy to see platform-centric view
  * Con: This is the hardest to navigate from a project perspective (submodules, fetch scripts, yuck!) 
5. The API lives is a central place as the "source of truth," and a copy (possibly with additional features) lives by-value with the implementation
  * Pro: This gives the best of both from a user perspective (easy to see project view, easy to see platform view)
  * Con: There is no longer a single source-of-truth and the two copies must be kept in sync

Personally, I've tended towards (2) in the past, but the logistical headaches of the `submodule` approach have been a nightmare. We also attempted to solve (2) with the source-of-truth living [in code](https://github.com/caikit/caikit/tree/main/caikit/interfaces), but the `code -> API spec` auto-generation turned out to be just as big a nightmare and folks continuously complained that it was impossible to actually _see_ the API.

For "move fast and break things," (1) is clearly the best approach, but I think the phase we're at right now is to try to back off from this perspective and create some commonality that is divorced from the individual project implementations.

I'm currently leaning towards (3) for these reasons. This would mean that the individual projects _could_ (should?) keep their own copies of their respective APIs, but that the "platform definition" would live in the central location and the onus would be on the project owner to keep the project in sync with the platform (i.e. contribute changes to the platform to support changes at the project level).

sauagarwa: The other option is we can start with 1, since it will be changing a lot initially, once we stabilize we can look into option 3.

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,20 @@
...
+* `Platform APIs`: These are APIs that are ignorant of `InstructLab` and provide generic AI platform capabilities (e.g., Fine Tuning, SDG, Eval)
+* `InstructLab APIs`: These are the APIs that reflect the user-facing functionality of `InstructLab` itself. They are aware of the end-to-end `InstructLab` workflow.
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/). For structural and style guidelines, see [api-definitions](../../api-definitions/README.md).
+
+## Where will service API definitions live?
```
booxter, line 18: How does the project guarantee that (manually written?) specs correspond to implementations? Wonder if these should be automatically generated by the REST framework used and not manually maintained separately from implementations.

Is there compliance testing planned for these APIs?

docs/backend/api-definitions-guidelines.md
booxter: > (2) with the source-of-truth living [in code](https://github.com/caikit/caikit/tree/main/caikit/interfaces), but the code -> API spec auto-generation turned out to be just as big a nightmare and folks continuously complained that it was impossible to actually see the API.

You could generate api definitions and store them along side the implementation, for user consumption. In one of my pet projects, I [use](https://github.com/booxter/letsrolld/blob/a7c0fca2196eff33308f565520da5c3f00b5ddd0/src/letsrolld/webapi/app.py#L233) [flask-restful-swagger-3](https://pypi.org/project/flask-restful-swagger-3/) that is used to enforce compliance in runtime as well as to generate swagger definitions (that I then store in the same repo - the correspondence between the two is enforced by github checks.)

relyt0925:
This looks good to me!

docs/backend/api-definitions-guidelines.md
leseb: I think the point of using openapi to describe the API is also to use side-tooling to auto-generate clients and SDKs, even server-side.

docs/backend/api-definitions-guidelines.md
leseb: The choice also depends on the programming language's dependency management capabilities. For example, with Go, option 2 is straightforward and avoids the hassle of submodules. You can have one repository as the source of truth for the API definition. Then, a separate repository hosts the generated client API code. The main Go project can use this client library as a module dependency. A similar approach could work with Python, where you produce a package on PyPI, thus avoiding submodules.

Option 3 with Git is essentially a re-sync, which is fairly easy to manage. However, this approach mimics what a submodule would do, requiring extra effort for synchronization. If we can devise an efficient synchronization strategy, this option might be viable.

I still prefer option 2 despite its management overhead because it provides a clear separation of concerns and an obvious location for the source of truth.

Option 1 is impractical since we will likely have multiple components consuming the API.

gabe-l-hart: @leseb Interesting point about generating packages off of the central repo. One of the things I've always liked about both OpenAPI and Protobuf is that the specification does not mandate an implementation language, so the server is free to implement the spec in any language they choose. I think we could accomplish this with your suggestion of a central repo if that repo's CI produced and published packages for all of the "well supported" languages (probably `python` + `go` to start, possibly with `rust` per [the other thread in Russell's proposal](https://github.com/instructlab/dev-docs/pull/74#discussion_r1630126956).

If we were to go this route, I think it would be a strong push to NOT hold these API definitions in the `dev-docs` repo so that we weren't mixing CI concerns between building docs and building generated code packages.

docs/backend/api-definitions-guidelines.md
gabe-l-hart: Yes, agreed. The goal of the OpenAPI specs is to leverage codegen and/or validation as much as possible in the server implementation language of choice. Per [the conversation below](https://github.com/instructlab/dev-docs/pull/71/files#r1630747990), I think this is a good direction to consider where we have an isolated repo with the specs that creates distributed packages for our supported language.

README.md
gabe-l-hart: Given my lack of opinion on this, I thought I would look to see if the internet had a strong opinion, and it turns out that `OpenAPI` itself does ([here](https://swagger.io/resources/articles/difference-between-api-documentation-specification/))!

**TLDR;** What we're discussing here is an `API Definition` (aimed at being consumed by machines) not an `API Specification` (aimed at being consumed by humans).

gabe-l-hart: >An API definition is similar to an API specification in that it provides an understanding of how an API is organized and how the API functions. But the API definition is aimed at machine consumption instead of human consumption of APIs. An API definition provides information about how the API functions, how it links with other APIs, and the expected results in a machine-readable format. It focuses on defining the API and outlining the structure of the API.

api-definitions/README.md
gabe-l-hart: See above comment about `API Definition` vs `API Specification` from Open API docs

README.md
gabe-l-hart: Given the conversation below about likely moving the definitions themselves to a different repo, I think I'm going to not add this suggestion for now.

docs/backend/api-definitions-guidelines.md
gabe-l-hart: I've. now refactored this PR to propose putting the definitions themselves into their own repo. Unfortunately, this makes it a little hard to find the conversation that led to the change!

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+## Style Guidelines
+
+* Use `kebab-case` for path elements
+  * All characters must be in the [ascii](https://www.ascii-code.com/) character set to avoid percent encoding in URIs
+  * All letters must be lowercase
+  * Words are separated by the `-` (dash) character
+* Use `snake_case` for properties
+  * All characters must be in the [utf-8](https://www.w3schools.com/charsets/ref_html_utf8.asp) character set for simple `json` encoding
+  * Words are separated by the `_` (underscore) character
+* Use `UpperCamelCase` for internal reusable schema names
```
leseb, line 42:
```suggestion
* Use `CamelCase` for internal reusable schema names
```

Is there a difference between `UpperCamelCase`, I mean isn't the "camel" doing the upper?

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/). For structural and style guidelines, see [api-definitions](../../api-definitions/README.md).
+
+## Where will service API definitions live?
+
+Service API definitions will live a new repository github.com/instructlab/service-api-definitions. This repo will have two primary responsibilities:
+
+1. House the static service API definitions
+2. Build and publish any language-specific generated packages for consumption by service implementation projects (see below)
```
leseb, line 23: When I read "build and publish," I understand it to mean that the generated files will be hosted in the service-api-definitions repository. I’m not sure this is the best approach.

I can see two scenarios:

The current approach, where the generated code is hosted in the service-api-definitions repository.
An alternative where we still build and publish, but to a dedicated repository. This option provides a clear separation of concerns and prevents the API spec repository from becoming bloated.

Did I understand correctly? If so, my preference would be option 2. Thanks!

leseb (changes requested):
Thank you for your diligence in revisiting all the comments. We're making great progress, and I appreciate your efforts!

docs/backend/api-definitions-guidelines.md
hickeyma: One repository as source of truth for the specifications would be my preference. The `dev-docs` is not a suitable locations as said.

README.md
hickeyma: @leseb probably does. This doc is a definition and the repo is a specification?! :)

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+There are two primary classes of service APIs needed to support InstructLab:
+
+* `Platform APIs`: These are APIs that are ignorant of `InstructLab` and provide generic AI platform capabilities (e.g., Fine Tuning, SDG, Eval)
+* `InstructLab APIs`: These are the APIs that reflect the user-facing functionality of `InstructLab` itself. They are aware of the end-to-end `InstructLab` workflow.
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/). For structural and style guidelines, see [api-definitions](../../api-definitions/README.md).
```
hickeyma, line 16:
```suggestion
Service APIs will be defined using the [OpenAPI Sepcification](https://www.openapis.org/) (OAS) written in [YAML](https://yaml.org/). For structural and style guidelines, see [api-definitions](../../api-definitions/README.md).
```

README.md
hickeyma, line 49:
```suggestion
API definitions use the [OpenAPI Sepcification](https://www.openapis.org/) (OAS) written in [YAML](https://yaml.org/).
```

docs/backend/api-definitions-guidelines.md
hickeyma: Where is see [api-definitions](../../api-definitions/README.md) ?

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Where will service API definitions live?
+
+Service API definitions will live a new repository github.com/instructlab/service-api-definitions. This repo will have two primary responsibilities:
```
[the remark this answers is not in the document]
hickeyma: I would prefer specification over definitions

hickeyma, line 20:
```suggestion
Service API specifications will live a new repository `instructlab/service-api-specification`. This repo will have two primary responsibilities:
```

docs/backend/api-definitions-guidelines.md
hickeyma: Maybe For structural and style guidelines, see [api-definitions](../../api-definitions/README.md) --> See [Style Guidelines](#style-guidelines) for guidance.

hickeyma (changes requested):
Thanks for the updates @gabe-l-hart. Some more comments provided.

docs/backend/api-definitions-guidelines.md
gabe-l-hart: Shoot, I just missed cleaning this one up. Good catch.

docs/backend/api-definitions-guidelines.md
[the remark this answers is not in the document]
gabe-l-hart: @hickeyma, can you elaborate on that preference in light of the definition vs specification article? My thought is that this new repo would _explicitly_ house `definitions` (for machine consumption). Are you thinking that it would also house `specifications` (for human consumption)?

docs/backend/api-definitions-guidelines.md
gabe-l-hart: Probably just a poor choice of words. It's probably language specific, and I'm less familiar with modern `go` package management (when I was writing it a number of years ago, it was still "grab the latest from the tip of master in the repo). For `python`, I was imagining the CI generating the python stubs (build) and putting them in a package that gets published to PyPI (publish).

docs/backend/api-definitions-guidelines.md
gabe-l-hart: Same question as above. Why the preference for `specification` vs `definition` if the explicit goal of the repo is to own machine-consumable `definitions`. Are you thinking that the `yaml` format is the human-readable `specification` vs the generated language-specific packages are the machine-readable `definitions`?

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+1. House the static service API definitions
+2. Build and publish any language-specific generated packages for consumption by service implementation projects (see below)
+
+## How will service implementations reference shared APIs?
+
+When a project chooses to implement one or more service APIs, there are three acceptable methods for doing so, listed in order of preference:
+
+1. Consume a supported language-specific package. The `service-api-definitions` repo will build consumable packages with generated code for supported languages. This is the preferred method of consumption as it avoids repository references and code duplication.
+2. For languages without a supported package, the `service-api-definitions` repo may be held as a [git submodule](https://www.git-scm.com/book/en/v2/Git-Tools-Submodules).
+3. It is also acceptable for an implementation to copy the relevant API definitions to the local project repository. Any changes made in the central repository will need to be sync'ed by the project owners, and any new APIs added in the project will not be considered usable until they have been integrated into the central API definitions.
```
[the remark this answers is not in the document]
gabe-l-hart: I agree with your preference. This is probably me being overly protective of folks who may feel that this proposal is imposing too much structure on individual project owners who just want to write code and move fast. Votes in favor of just eliminating this option?

docs/backend/api-definitions-guidelines.md
gabe-l-hart: Yeah, fair point. I wanted to make it explicit that it's `UpperCamel` vs `lowerCamel` (e.g. `java` classes vs variables/methods)

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+* Use `UpperCamelCase` for internal reusable schema names
+  * These are internal names, so the character set is not limited
+  * Words are capitalized and concatenated with no separator
+
+## API Layout
+
+* There will be two main portions of the APIs:
+  * `instructlab.yaml`: This defines the user-facing `InstructLab` REST API
+  * `platform.yaml`: This defines the platform-level APIs used by the `InstructLab` workflow.
+* Each platform `Capability` should own its own fully-functional sub-API file that can be used by individual capability service implementations
```
booxter, line 51: what is "platform capability"? is it defined elsewhere? maybe an example could clarify it.

leseb (commented):
I think we are getting close, thanks for keeping up Gabe!

docs/backend/api-definitions-guidelines.md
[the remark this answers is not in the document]
gabe-l-hart: Ah, good point. I think you're right that the `kubernetes` community uses `spec` to mean what OpenAPI calls `Definition`, though I suppose it's also a little muddy with `CustomResourceDefinition` which is the thing that defines the `schema` for a "thing" and then `spec` being the canonical name for the field within that `schema` where a user would create an instance of that "thing." I that vein, it still feels like `Definition` is the consistent word where the `schema` is defined.

That said, I'm a big fan of putting a `Cares About` number (1-10) on this kind of opinion and then just taking the highest one.

**CA: 2**

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/).
+
+## Where will service API definitions live?
+
+Service API definitions will live a new repository github.com/instructlab/service-api-definitions. This repo will have two primary responsibilities:
```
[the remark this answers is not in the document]
gabe-l-hart: Nice! I do like that. It's explicit, shorter, and gets us out of the `specification`/`definition` discussion

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. House the static service API definitions
+2. Build and publish any language-specific generated packages for consumption by service implementation projects (see below)
+
+## How will service implementations reference shared APIs?
+
+When a project chooses to implement one or more service APIs, there are three acceptable methods for doing so, listed in order of preference:
+
+1. Consume a supported language-specific package. The `service-api-definitions` repo will build consumable packages with generated code for supported languages. This is the preferred method of consumption as it avoids repository references and code duplication.
```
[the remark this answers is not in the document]
gabe-l-hart: Good question. I'm coming from the `python` POV where `build` would be creating the `wheel` and pushing to PyPI. I haven't used `go` extensively since the repo-management was all centered around pulling the tip of `master`, so I'm not clear if there's any equivalent way to have statically prebuilt packages with `go.mod`. I can rephrase to be a little more generic.

docs/backend/api-definitions-guidelines.md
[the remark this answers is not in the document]
gabe-l-hart: I'm going to make this change and we can iterate on it in comments from there

docs/backend/api-definitions-guidelines.md
gabe-l-hart: I somewhat defined it above under [What parts of InstructLab need service APIs](https://github.com/instructlab/dev-docs/pull/71/files#diff-5d6426ec3d00a810ecc49e2fc141ac72545910a7caaaaccbd0ab42315324c96eR9). It's admittedly a bit of a "definition by example." Does this feel sufficient?

booxter: Is `capability` here just a `component` of sorts? The `capability` term threw me off a bit (I think of capabilities as properties of things; not as things themselves.)

booxter:
Directionally, this looks fine. The initial guidelines are quite abstract and perhaps will be clarified / filled with essence once actual APIs are proposed and actual bindings are generated. The main point is - the guidelines clarified that bindings will be auto-generated and hosted elsewhere; and there will be a separate repo for schema itself. This should be enough to merge it IMO.

docs/backend/api-definitions-guidelines.md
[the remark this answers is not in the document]
booxter: Go module URIs are git repos. I am not aware of package index for go modules. AFAIU some automation in the golang module repo will have to track changes to schema and generate commits to update bindings module. This should be doable. The comment of mine here is quite minor - the `build` term is not applicable to some platforms. But the essence - auto-generation of bindings - is the same.

docs/backend/api-definitions-guidelines.md
gabe-l-hart: Yeah, I'm using this word a bit loosely, though we've been working on a more rigorous definition outside of the InstructLab scope.

My other hat outside of IL has been focusing on trying to put shape around "the AI platform." To this end, I've wanted a way to separate _what_ the platform can do from _how_ it does it (interface vs implementation). I've been using the word `capability` to mean a logical grouping of interface definitions that holistically accomplish some user-focused AI need (e.g. "fine tune a model" would include "submit fine tuning," "poll tuning job," "cancel tuning job," etc...). Each `capability` can be implemented using any number of projects, and may have multiple implementations depending on use-case (e.g. multiple SDGs for different form-factors).

In this sense the `capability` is a property of the `platform`. I didn't really want to lead with trying to define abstract platform terminology in this PR, but if there's appetite, I'd be happy to try to put something together to formalize the verbiage.

docs/backend/api-definitions-guidelines.md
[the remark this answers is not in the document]
gabe-l-hart: Makes sense (I'm _very_ outdated on `go`). I updated the language to try to fuzz the operation beyond strictly `build`.

docs/backend/api-definitions-guidelines.md
booxter: Thanks Gabe for quick answer. Terms are important. But no need to expand this PR. I was just confused. I think I understand where you come from, even if I may not be too thrilled about particular word choices.

docs/backend/api-definitions-guidelines.md
leseb: Resolving this thread since we 're all on the same page!

docs/backend/api-definitions-guidelines.md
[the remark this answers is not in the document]
leseb: Looks like we went with `instructlab/openai`. I'm fine with that.

leseb (approved):
I'm happy with the current content. The "how we sync generated APIs" many not yet be fully formed or may change a little as we begin to deep dive more into the implementation. However, I think we all agreed on:

* Having a new repo for the source of truth (the API)
* Generate client/SDK/server code using that API def onto new repository:
  * it could be a pull from a client repo
  * it can be a build/push from the API def repo

Thanks for all your efforts Gabe!

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/).
+
+## Where will service API definitions live?
+
+Service API definitions will live a new repository github.com/instructlab/openapi. This repo will have two primary responsibilities:
```
nathan-weinberg, line 20:
```suggestion
Service API definitions will live a new repository: `instructlab/openapi`. This repo will have two primary responsibilities:
```

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+* Use `kebab-case` for path elements
+  * All characters must be in the [ascii](https://www.ascii-code.com/) character set to avoid percent encoding in URIs
+  * All letters must be lowercase
+  * Words are separated by the `-` (dash) character
+* Use `snake_case` for properties
+  * All characters must be in the [utf-8](https://www.w3schools.com/charsets/ref_html_utf8.asp) character set for simple `json` encoding
+  * Words are separated by the `_` (underscore) character
+* Use `UpperCamelCase` for internal reusable schema names
+  * These are internal names, so the character set is not limited
+  * Words are capitalized and concatenated with no separator
```
nathan-weinberg, line 43: We should enforce this via linting

nathan-weinberg (approved):
one nit but otherwise LGTM

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+There are two primary classes of service APIs needed to support InstructLab:
+
+* `Platform APIs`: These are APIs that are ignorant of `InstructLab` and provide generic AI platform capabilities (e.g., Fine Tuning, SDG, Eval)
+* `InstructLab APIs`: These are the APIs that reflect the user-facing functionality of `InstructLab` itself. They are aware of the end-to-end `InstructLab` workflow.
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/).
```
hickeyma, line 16:
```suggestion
Service APIs will be defined using the [OpenAPI Specification (OAS)](https://www.openapis.org/) written in [YAML](https://yaml.org/).
```

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+## Where will service API definitions live?
+
+Service API definitions will live a new repository: `instructlab/openapi`. This repo will have two primary responsibilities:
+
+1. House the static service API definitions
+2. Build and publish any language-specific generated packages for consumption by service implementation projects (see below)
+
+## How will service implementations reference shared APIs?
+
+When a project chooses to implement one or more service APIs, there are three acceptable methods for doing so, listed in order of preference:
```
hickeyma, line 27:
```suggestion
When a project chooses to implement one or more service APIs, there are two acceptable methods for doing so, listed in order of preference:
```

docs/backend/api-definitions-guidelines.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `InstructLab APIs` are essential for hosting `InstructLab` as a service in a repeatable way. The `Platform APIs` are critical for component reuse and extensibility (e.g., new SDG algorithms for new taxonomy data types), but they are not strictly required for hosting `InstructLab` as a service.
+
+## How will service APIs be defined?
+
+Service APIs will be defined using [OpenAPI](https://www.openapis.org/) format in [YAML](https://yaml.org/).
+
+## Where will service API definitions live?
+
+Service API definitions will live a new repository: `instructlab/openapi`. This repo will have two primary responsibilities:
```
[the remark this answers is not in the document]
gabe-l-hart: 🤦 good catch

docs/backend/api-definitions-guidelines.md
gabe-l-hart:
```suggestion
Service APIs will be defined using the [OpenAPI (OAS)](https://www.openapis.org/) format written in [YAML](https://yaml.org/).
```

hickeyma (approved):
This looks fine as it gives guidelines and direction with API definitions. I think the "how" will be worked out as we generate and sync specific APIs. We are taking the right approach using OpenAPI specifications and adding the specifications to a new repo for source of truth.

Thanks @gabe-l-hart for working on this.

hickeyma:
@markstur @russellb The guideline looks good to go. Do you have any further comments or obejctions to merging and moving forward?

markstur (approved):
lgtm

PR #74, opened 2024-06-04
backend: Add a future backend overview

russellb:
Add a high-level proposed overview of the future of InstructLab
backend services. There are a few key things covered here:

- defining guiding principles to assist in our decision-making

- providing a high-level overview of a services-oriented architecture
  for backend services - a set of REST APIs providing platform-level
  services and an InstructLab workflow on top.

- List key technology decisions that are helpful to call out early before
  proceeding more deeply into implementation.

Signed-off-by: Russell Bryant <rbryant@redhat.com>


russellb:
@mergifyio rebase

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,209 @@
...
+## Key Technology Decisions
+
+### Kubernetes
+
+We consider Kubernetes an acceptable dependency for the Backend Service. While
+it may be technically possible to run most of the code elsewhere, the project
+will focus its energy on running this collection of components in a
+Kubernetes-based environment.
+
+This also applies to when these components are run on a single server. In that
```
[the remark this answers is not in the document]
russellb: I would prefer putting that off into a future, dedicated proposal. The simplest way to get started is to focus on clustered operation, as that is where the "real" backend runs anyway. That is what we're aiming to replace as quick as possible.

russellb:
I moved this to draft, because I wrote many weeks ago originally. I want to update some of the details with some evolved opinions.

docs/backend/backend-overview.md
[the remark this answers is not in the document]
rhatdan: Sure, I see this as version 1.1 or 2.0 of the system. The key idea is we need to get to the point that `ilab command` on linux talks to the RESTAPI for launching commands. On RHEL AI, we can back the RESTAPI by launching podman containers or talking to podman services running on the system, on OpenShift AI systems, the RESTAPI can launch kubernetes workloads.

When we get a plugin for Podman Desktop, it could launch containers in the local podman machine or talk directly to the RESTAPI on RHEL AI or OpenShift AI.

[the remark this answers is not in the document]
russellb: Yeah, I understand the proposal. I just prefer to keep it out of scope for now.

russellb:
> I moved this to draft, because I wrote many weeks ago originally. I want to update some of the details with some evolved opinions.

I made the tweaks I had in mind. This is ready for review.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+or down as needed.
+
+We need to drive down the minimum requirements while also ensuring we can scale
+up to much larger environments with more resources available and make full use
+of those resources.
+
+As a minimum target, we should be able to run some configuration of the full
+backend pipeline on a single machine. We can use these as minimum requirements:
+
+- Linux server with 4x H100 GPUs
```
rhatdan, line 74: This seems rather particular, and probably should say example. Do we have any idea on non Nvidia Platforms?

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+design services that are focused on their needs.
+
+We should consider the persona of an InstructLab Model Developer and what
+actions and information they care about. That should drive the service
+interface provided by our Backend Service. Some examples of this point:
+
+- An InstructLab Model Developer cares about defining inputs to our pipeline,
+  launching that pipeline, following its status, and consuming its outputs.
+
+- An InstructLab Model Developer does not care about the exact details of how
```
rhatdan, line 88: In RHEL AI systems there will be no Kubernetes, as of now.

russellb: This is an upstream doc. RHEL AI is out of scope.

The doc is also focused on what we need to build to have an open and reproducible full backend that can run on a full cluster. The tools used to produce the full model builds out of the project are not open and not reusable in their current form anyway. Evolving and replacing that is the primary goal here. It just so happens that we could scale down the same tools to a single server, as well.

docs/backend/backend-overview.md
russellb: I got this example target many, many weeks ago. It could definitely use some expanding. All of this is currently being figured out as we expand the `ilab` capabilities with capabilities more in line with what the current backend uses.

I can update the wording to make it clear this is one acceptable example, but there will be more.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+  They don't care about the underlying infrastructure details. They care about
+  data sets, model training, model evaluation, and all of those types of related
+  details associated with the pipeline implemented by the Backend Service.
+  - **Sub-category A**: A Model Developer who is trying to add domain-specific
+  knowledge to the model and is very interested in how they have improved (or
+  not) the model in that specific domain
+  - **Sub-category B**: A Model Developer who is interested in improving an
+  overall model benchmark
+
+## Guiding Principles for the Future
```
leseb, line 37: What bugs me about this section is that the principles mix general software best practices with project-specific desires. For instance, 'scalability' is a general software best practice, whereas 'Model Independence' is specific to our project.

I hope that makes sense.

leseb (changes requested):
Thanks for putting together this design!

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+of concerns should be maintained in a future service-oriented architecture.
+The diagram under [Service-Oriented](#service-oriented) shows a high-level diagram
+with a single "InstructLab Service API." Implementation of each sub-section of
+the functionality should be done in such a way that the core of it is reusable.
+In some cases that may be via libraries. In others, it may make sense to build a
+microservice that can be used in other contexts.
+
+Within InstructLab, we could have a top-level application that implements the
+entire backend pipeline workflow and does so by making use of each of these
+libraries and/or services.
```
gabe-l-hart, line 120: I like this statement. Do you think it's adding something explicit about supporting different cluster/infra configurations?

```suggestion
libraries and/or services. The InstructLab application will be responsible for supporting all cluster configurations behind a common API, in service of the Scalable principle.
```

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+### Small Iterations
+
+While we may cast a vision for the future, we desire to get there via a series
+of small iterations. We have a working backend pipeline already and will
+continue to run it every week. In parallel, we can work to evolve it through a
+series of small iterations that evolve our processes over time to use new
+service-based interfaces, gradually evolving the execution process that occurs
+each week.
+
+We do not want a new effort happening in parallel to what is already working
```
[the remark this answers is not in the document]
gabe-l-hart: I agree with this. Is it worth making this point explicit to say that the architectural changes to support the backend service implementation must be sufficiently modular that they can be adopted in the product piece-by-piece? I think this is what the `Small Iterations` implies, but it could also just mean "build the backend one piece at a time" versus "make the backend consumable on piece at a time."

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+today.
+
+### Minimize Mandatory External Dependencies
+
+Major external dependencies should be made optional where feasible to do so.
+The more of these we require, the more complicated it will be to reuse this
+system in other environments.
+
+In other cases, it may be appropriate to provide an abstraction for the
+dependency so that alternatives can be put in place. Alternate Object Storage
```
gabe-l-hart, line 143: It might be worth mentioning that the abstraction should be extensible by the `Backend Service Owner` (i.e. at `deploy time`).

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+
+Most people will not have direct access to a large cluster to test changes
+to this system. We should create development environments and minimal
+configurations as much as possible to lower the barrier to doing useful
+development and testing with these components.
+
+### Model Independence
+
+We desire to keep the backend flexible such that serving and training alternate
+models in the future is not more work than absolutely necessary.
```
gabe-l-hart, line 156: This may be too much in the weeds, but it's my experience that there are two places where model independence/model agnostic tends to fall down:

1. Parameters/data formats for tuning processes
2. Engineered prompts for inference operations

I think we'll have a number of (2) in `InstructLab` (SDG and Eval come to mind), so I think we'll want to be able to understand exactly where these exist and how a user would customize them for a different choice of teacher/critic model.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+### Make Development Accessible
+
+Most people will not have direct access to a large cluster to test changes
+to this system. We should create development environments and minimal
+configurations as much as possible to lower the barrier to doing useful
+development and testing with these components.
+
+### Model Independence
+
+We desire to keep the backend flexible such that serving and training alternate
```
gabe-l-hart, line 155: This seems to define the "Model" in `Model Independence` as the model being trained. I think we'd also want to extend that to models used during the training process (teacher and critic) which may not be the same as the model being trained (and likely won't be).

docs/backend/backend-overview.md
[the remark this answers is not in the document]
gabe-l-hart: @russellb is the goal of keeping non-k8s REST APIs out of scope to make inclusion of single-node k8s the intentional direction for RHEL AI? That seems like a large discussion that may be bigger than this review.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+Python is the primary language used in the current backend. It is also the
+de facto choice for AI/ML developers. We will continue using Python for core
+functionality, especially when it comes to interfacing with AI/ML libraries like
+PyTorch.
+
+Another significant element of the Backend Service is automating workloads on
+Kubernetes. The most popular language in this context is Go. Kubernetes itself
+is in Go, and the Go client libraries are the most mature.
+
+To best support the full scope of the project, we will use both Python and Go,
```
gabe-l-hart, line 191: Per our earlier conversation, I think it may be worth including `rust` in the list for service implementations. Alternatively, it may be worth noting that this is not an exclusive list, but instead an indication of supported preference.

docs/backend/backend-overview.md
hickeyma: I would rather say that we are still working out what the minimum GPUs and VRAM is. A minimum specification should be based on research and definite e2e (model tuned and improved) testing.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+
+We should consider the persona of an InstructLab Model Developer and what
+actions and information they care about. That should drive the service
+interface provided by our Backend Service. Some examples of this point:
+
+- An InstructLab Model Developer cares about defining inputs to our pipeline,
+  launching that pipeline, following its status, and consuming its outputs.
+
+- An InstructLab Model Developer does not care about the exact details of how
+  these processes are running on a cluster. They should not require any access
```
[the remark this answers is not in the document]
hickeyma: I agree. I think we should be open to the idea that the backend can be deployed in more than a k8s cluster e.g. bare metal.

docs/backend/backend-overview.md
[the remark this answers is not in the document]
hickeyma: Its ok to keep it out of scope as long as the backend can be deployed independent of containers and k8s cluster. It can be flexible.

hickeyma (commented):
Great to have this doc strated, thanks @russellb for the work on this. Some comments inline.

docs/backend/backend-overview.md
russellb: Yeah, I remember the conversation, but I couldn't come up with a concrete example of something in scope that fit the criteria we discussed for where it would make sense.  I'm open to it, but I just want to provide more concrete justification. I feel like I can provide those for Python and Go.

To be clear I'm focused on InstructLab specific services with these comments on implementation details.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+- Observable
+- Reusable
+- Small Iterations
+- Minimize Mandatory External Dependencies
+- Make Development Accessible
+- Model Independence
+
+### Scalable
+
+The initial iteration of the Backend Service targeted a particular set of clusters
```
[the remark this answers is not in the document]
russellb: Yeah -- this was referring to an older implementation, not the one proposed in this doc that would work differently.  apologies if that didn't come through correctly.

An earlier iteration of this doc had a whole lot more detail about the "old" system, but I ripped it all out since it's not really so relevant (nobody can see the details of it, anyway)

docs/backend/backend-overview.md
russellb: Yeah, I like that. I'll drop the specific reference and just say -- "we should really work hard to capture requirements for various configurations and use cases"

docs/backend/backend-overview.md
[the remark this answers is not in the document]
russellb: I can tweak the wording here to be a bit more high-level and still get the intent of the point across - thanks for the feedback

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+  Administrator.  Ensuring the Backend Service is configured and running
+  correctly on the cluster is the concern of the Backend Service Owner.
+
+If we map this back to the InstructLab project itself, you can imagine that the
+Backend Maintainers team serves as Backend Service Owners. There may be another
+team, the taxonomy maintainers, that represent InstructLab Model Developers,
+focused on evaluating taxonomy contributions and running batches of them through
+the backend pipeline.
+
+![High-Level Service Diagram](./images/backend-service-high-level.png)
```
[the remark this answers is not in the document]
russellb: You're right, they should be outside, thanks!

docs/backend/backend-overview.md
gabe-l-hart: Yep, I think limiting to `python` and `go` for the time being is fine as long as it's not phrased as an exclusive list. Maybe we could add a sentence along the lines of:

> Additional languages may be considered for addition to the supported list as needed through additional proposal documents in this repo.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Observable
+
+The Backend Service must be observable, both by the Backend Service Owners and
+by InstructLab Model Developers.
+
+### Reusable
+
+The Backend Service was envisioned from the beginning as a pipeline of distinct
+steps (data generation, training, evaluation, and publishing). This separation
```
nathan-weinberg, line 110:
```suggestion
steps (synthetic data generation, training, evaluation, and publishing). This separation
```

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+entire backend pipeline workflow and does so by making use of each of these
+libraries and/or services.
+
+![Service Breakdown](./images/backend-service-breakdown.png)
+
+### Small Iterations
+
+While we may cast a vision for the future, we desire to get there via a series
+of small iterations. We have a working backend pipeline already and will
+continue to run it every week. In parallel, we can work to evolve it through a
```
nathan-weinberg, line 128: This sentence should be removed
```suggestion
of small iterations. In parallel, we can work to evolve it through a
```

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+series of small iterations that evolve our processes over time to use new
+service-based interfaces, gradually evolving the execution process that occurs
+each week.
+
+We do not want a new effort happening in parallel to what is already working
+today.
+
+### Minimize Mandatory External Dependencies
```
nathan-weinberg, line 136: One other worth calling out here is that some external dependencies we are bringing "in-house" as forks - for example, https://github.com/instructlab/GPTDolomite

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+development and testing with these components.
+
+### Model Independence
+
+We desire to keep the backend flexible such that serving and training alternate
+models in the future is not more work than absolutely necessary.
+
+## Key Technology Decisions
+
+### Kubernetes
```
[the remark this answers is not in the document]
nathan-weinberg: This is not meant to be a pointed question but I do want to raise, why Kubernetes specifically?  It should definitely be a supported platform, but we seem singularly focused on running it there when there is a such a broad spectrum of orchestration platforms.

docs/backend/backend-overview.md
nathan-weinberg: I like the above wording by Gabe

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+Another significant element of the Backend Service is automating workloads on
+Kubernetes. The most popular language in this context is Go. Kubernetes itself
+is in Go, and the Go client libraries are the most mature.
+
+To best support the full scope of the project, we will use both Python and Go,
+depending on the specific needs of the component in question.
+
+## An Example Service - Data Generation
+
+The Data Generation step is the first piece of the backend pipeline. To help
```
nathan-weinberg, line 196:
```suggestion
## An Example Service - Synthetic Data Generation

The Synthetic Data Generation (SDG) step is the first piece of the backend pipeline. To help
```

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+The following personas will be referenced throughout the document:
+
+- **Cluster Administrator**: The person responsible for managing a Kubernetes cluster.
+  This person cares about all hardware and its associated hardware and
+  cluster-wide software resources.
+
+- **Backend Service Owner**: The person responsible for managing the Backend Service
+  on top of the cluster provided by the Cluster Administrator.
+
+- **InstructLab Model Developer**: This person is a user of the Backend Service.
```
franciscojavierarceo, line 27: Would you bucket a Data Scientist, ML Engineer, or AI Engineer in this group? I think so but I wanted to make sure I understood the intent here.

docs/backend/backend-overview.md
```diff
@@ -0,0 +1,210 @@
...
+to this system. We should create development environments and minimal
+configurations as much as possible to lower the barrier to doing useful
+development and testing with these components.
+
+### Model Independence
+
+We desire to keep the backend flexible such that serving and training alternate
+models in the future is not more work than absolutely necessary.
+
+## Key Technology Decisions
```
franciscojavierarceo, line 158: Kubeflow is a strategic priority for OpenShift AI. Do you think it would be among one of the acceptable dependencies? 

For larger scale fine tuning and orchestrating more intensive pipelines, the [Training Operator](https://www.kubeflow.org/docs/components/training/overview/) and [KFP](https://www.kubeflow.org/docs/components/pipelines/overview/) feel like they could be key technology decisions. 

Any thoughts here @gabe-l-hart?

gabe-l-hart: This is definitely interesting. I believe at this point for kubernetes-based backend deployments (which notably does not include RHEL AI), the Training Operator will almost certainly be a key piece of technology. KFP may also play a role in data prep if/when the SDG workflows expand to include code-based customization versus just config-based (the current implementation doesn't allow this, but it may happen in the future).

I haven't taken a pass over this doc in a while and I know it was originally written before the decision was final to not include a lightweight kubernetes distro in RHEL AI (see discussion below). @russellb do you plan to update the doc based on the state of kube as a prereq now that we're further along in the RHEL AI process?

russellb: I don't plan to update this right now as I have some other priorities.

russellb:
This is a bit stale, and I do not plan to update it soon, so I will close it. I'll reopen later if I plan to push it further.

PR #80, opened 2024-06-06
Initial SDG REST API definitions

gabe-l-hart:
## Related PRs

This PR depends on #71 being merged

## Description

This PR adds the first platform API definition for `Synthetic Data Generation`


api-definitions/common/file-pointer.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+schemas:
+  LocalPath:
+    required: ['path']
+    properties:
+      path:
+        type: string
+        description: The name of the file on disk
+
+  S3Path:
```
leseb, line 14: Why does it have to be S3? How about we make this API standard agnostic? `ObjectStoragePath` looks sufficient (or something along those lines) and will comply with S3 as well as AzureBlob for instance. We might need to introduce a `provider` property that specifies the type of object storage.

Or perhaps you can point me to where it was decided to use S3 only? Thanks!

api-definitions/common/job-status.yaml
```diff
@@ -0,0 +1,29 @@
...
+        type: string
+        description: Unique identifier for a single job
+      status:
+        type: string
+        enum:
+          - QUEUED
+          - RUNNING
+          - COMPLETED
+          - CANCELED
+          - ERRORED
```
leseb, line 22: Should we map with whatever Kubernetes says on Job'status?

api-definitions/common/s3-connection.yaml
```diff
@@ -0,0 +1,51 @@
...
+        type: string
+        description: The private Secret Key
+
+  IAMCredentials:
+    type: object
+    properties:
+      # TODO: What else goes here?
+      apikey:
+        type: string
+        description: The IAM apikey
```
leseb, line 23: I believe both HMAC and IAM use the same concept of access_key and secret_key. If so, how about we consolidate both properties into one?

Something like?

```yaml
  Credentials:
    type: object
    description: Credentials for accessing the service
    properties:
      access_key:
        type: string
        description: The public access key
      private_key:
        type: string
        description: The private key
```

?

leseb (changes requested):
Thanks for drafting more API specs :) I believe the object storage piece should be more agnostic to suit any object storage system (cloud GCP/Azure/AWS/IBM). Overall this looks nice even though as commented on another PR that I don't feel this repo is the right place for putting the API specs. Thanks!

hickeyma (commented):
@gabe-l-hart AIs its a draft, is this ready for review or is mopre to be done?

gabe-l-hart:
Thanks for digging in on this @hickeyma @leseb! I have this in draft until we merge #71 since it will likely need refactoring/moving once that conversation is done. I'll dig into the comments on the content of the specs later today (I agree that the connection stuff probably needs some serious refining).

api-definitions/common/file-pointer.yaml
gabe-l-hart: Interesting. This is probably one of those places where I have a mental word for a thing that I've never questioned. I was under the impression that `S3` had been adopted as the abstract API of which `Amazon S3` was the canonical, but not only, implementation. For example [IBM COS](https://cloud.ibm.com/apidocs/cos/cos-compatibility) documents itself as `S3 compatible`. I'm absolutely open to using a more generic name for this if `S3` is not the right generic label.

api-definitions/common/s3-connection.yaml
gabe-l-hart: I know at least for [IBM IAM](https://github.com/IBM/ibm-cos-sdk-python?tab=readme-ov-file#quick-start), there are different credentials (`apikey` + token endpoint + short-lived `token`). I think the AWS definition of IAM is different though, so maybe `IAM` is not a unique identifier.

The general question here would be whether we want to force `HMAC` and use that as the only supported access mechanism or try to keep this `oneOf` with an extensive list of provider-specific credential blob types.

api-definitions/common/job-status.yaml
relyt0925: Do we feel we need to model anything for when a job goes through a "temporary failure" and let's say goes through a retry? Or would we jusft go from FAILED to QUEUED again? Or would we just consider that "process" as another job entirely. 

Just thinking through how we would like to look at modeling what could happen let's say when a job hits a transient failure (let's say due to part of it running on bad infrastructure that is then replaced) and then a retry of that is scheduled.

api-definitions/platform/synthetic-data-generations.yaml
```diff
@@ -0,0 +1,104 @@
...
+    post:
+      summary: Initialize a Synthetic Data Generation job
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/SDGJobBody'
+      responses:
+        "200":
```
relyt0925, line 17: should this be 201 since it's async?

api-definitions/platform/synthetic-data-generations.yaml
```diff
@@ -0,0 +1,104 @@
...
+        data_json_schema:
+          type: object
+          description: The json schema for input data files for this task
+          # TODO: This doesn't render cleanly for some reason, but the body here
+          # must be a valid JSON Schema
+          # $ref: 'https://json-schema.org/draft-04/schema#'
+        data_example:
+          type: string
+          description: Example of an input data file for this task
+        config_json_schema:
```
relyt0925, line 96: By "config" is this where you visualize a flexible "json blob" to where users could request "advanced parameters" when necessary to feed into SDG that maybe aren't default (lower level things like num samples, algorithm utilized in SDG, etc). 

I like the idea of it then starting out really flexible and I guess different implementations at potentially different moments would potentially only allow a given subset of what can be sent in the config level.

api-definitions/platform/synthetic-data-generations.yaml
```diff
@@ -0,0 +1,104 @@
...
+    SDGJobBody:
+      type: object
+      required: ['output_directory', 'tasks', 'seed_data']
+      properties:
+        output_directory:
+          description: Location to place the output in
+          $ref: '../common/file-pointer.yaml#/schemas/DirectoryPointer'
+
+        tasks:
+          description: Mapping from task name to task config. The config will be validated against the task's config schema.
```
relyt0925, line 68: I see there's the potential for multiple "tasks" within one job: what do you visualize that being long term? Would it be something along the lines potentially of being able to have a "generate" task, a "mix" tasks to do random mixing of that data, and/or potentially a "filter task" to then filter some of the data and that should still all live within the context of a SDGJob?

api-definitions/common/file-pointer.yaml
leseb: I don't think we should be tied to S3, especially since other cloud providers like GCP and Azure are not compatible with it. Also S3 is a protocol not strictly an interface (I don't want to get into semantics; I'll lose that battle lol) but I hope you get my point. It's an object interface, like we have file and block. I maintain `ObjectStoragePath` which makes us agnostic or `ObjectPath`. Hope that makes sense.

gabe-l-hart:
Per revisions to https://github.com/instructlab/dev-docs/pull/71, the body of this PR will likely be moved to a new repo. I'll continue to address the comments here until the conversation on #71 is resolved

api-definitions/platform/synthetic-data-generations.yaml
gabe-l-hart: Yep, good call!

api-definitions/platform/synthetic-data-generations.yaml
gabe-l-hart: Good question. I took this from the CLI args to the library we're thinking of for the generic platform SDG implementation. I think for `InstructLab`, we'll likely only ever run a single `Task` (dataset + config + generation algorithm) per job.

api-definitions/platform/synthetic-data-generations.yaml
gabe-l-hart: Yep, that's exactly the idea here. We imagine the set of `tasks` to be extensible. Initially, this would be "build time" where the owner of the docker image would rebuild with new task implementations, but eventually we'd probably imagine users creating their own `tasks` by binding proprietary datasets/prompts/etc to existing generation algorithms. Each generation algorithm has a set of lower-level configs that can theoretically be overridden for each job, so the idea with this API is that when creating the job, the config overrides are an opaque blob, but you can query the system to understand the right schema for that blob beforehand. This avoids the need for us to keep a giant `oneOf` in the API definitions while still giving the user the ability to know the acceptable schemas that will be used for validation.

leseb:
Putting on hold given that the definitions will move to a new repo (we all agreed on that approach) once setup. Thanks!

api-definitions/common/job-status.yaml
leseb: I like the enum too as well as the remap from the Kube terminology. Thanks!

api-definitions/common/s3-connection.yaml
leseb: After reading the proposal a second time, I’m okay with it. I think I missed something during my initial review. 🤔

relyt0925:
Thanks @gabe-l-hart for your time!!!! This looks really cool and flexible

PR #81, opened 2024-06-07
Design for serving models with different backends

alimaredia:
This commit introduces a new design document for the [`ilab model
serve`] command, detailing its
functionality to serve models using different backends, specifically
`llama-cpp` and `vllm`. It outlines the command structure, including
backend-specific flags and arguments, and proposes a testing strategy
for new engine integrations.

Co-authored-by: Sébastien Han <seb@redhat.com>
Signed-off-by: Ali Maredia <amaredia@redhat.com>

### DIFF
--- a/docs/cli/ilab-model-serve-backend.md
+++ b/docs/cli/ilab-model-serve-backend.md
@@ -1,0 +1,142 @@
+# Design for `ilab model serve` command with backend support
+
+## Background
+
+With the [request from the community](https://github.com/instructlab/instructlab/issues/1106) for `ilab` to serve different backends such as [vllm](https://docs.vllm.ai/en/stable/) and the [cli redesign](ilab-model-backend.md), this design doc's purpose is to flesh out the behavior of the `ilab model serve` command.
+
+Specifically, this doc addresses the design of subcommands of `ilab model serve` that apply to
+different serving backends.
+
+## Design
+
+### Backend
+
+Since the subject of the `ilab model serve` command is a model, regardless of the format of the model, every command takes in the `--model` flag or uses its default value in the config.
+
+`ilab model serve` has a new flag `--backend` that will be used to serve models with. As of this design, the two backends `ilab` would serve with are `llama-cpp` and `vllm`.
+
+This would lead to the commands:
+
+- `ilab model serve --backend llama-cpp`
+- `ilab model serve --backend vllm`
+
+There are specific flags for `ilab model serve` that would apply to all backends. These can be viewed by running `ilab model serve --help`.
+
+The following is an overview for the flags of `ilab model serve`:
+
+```console
+ilab model serve
+|
+|_______ (backend agnostic flags)
+|
+|_______ --backend ['llama-cpp', 'vllm']
+|_______ --backend-args
+```
+
+The `backend` flag will also be available as an option in the config file (`config.yaml`). This will allow users to
+set a default backend for `ilab model serve` in the config. Also, commands like `ilab model chat`
+and `ilab data generate` that serve models in the background will use the default backend specified
+in the config. Here is an example of what the config file would look like:
+
+```yaml
+serve:
+  gpu_layers: -1
+  host_port: 127.0.0.1:8000
+  max_ctx_size: 4096
+  model_path: models/merlinite-7b-lab-Q4_K_M.gguf
+  backend: llama-cpp
+```
+
+### Backend flags
+
+The `--backend-args` flag is a string that will be passed to the backend as arguments. This flag is used to pass
+backend-specific arguments to the backend. Multiple values will be supported, however the exact formatting will be
+defined in the implementation proposal. The backend will be responsible for parsing individual arguments.
+
+It will also be available as an option in the config file (`config.yaml`). This will allow users to set default backend arguments for `ilab model serve` in the config. Here is an example of what the config file would look like:
+
+```yaml
+serve:
+  backend: llama-cpp
+  backend_args:
+    num_gpu_layers: 4
+    max_ctx_size: 1024
+```
+
+For clarity and ease of implementation, when using the `--backend-args` flag, the user must pass the
+`--backend` flag as well. This is to ensure that the backend-specific arguments are passed to the
+correct backend. Any backend-specific arguments that are not passed to the correct backend will be
+reported as an error.
+
+## Command Examples
+
+### Bare-bones but model specific command
+
+```shell
+ilab model serve --model <PATH>
+```
+
+- Serves the model at `<PATH>`.
+- If the `<PATH>` is the path for a model that can be run by `llama-cpp` then `llama-cpp` is
+  automatically used as the model serving backend. The current auto-detection logic will rely on a
+  valid GGUF file format. If the model is a valid GGUF file, then `llama-cpp` will be used as the model serving backend.
+- If the `<PATH>` is the path for a model that can be run by `vllm` then `vllm` is automatically used as the model serving backend.
+- If the model at `<PATH>` can be run by either backend, then a default backend defined in the
+  config will be used as the model serving backend. In the case where there is ambiguity and a setting is not defined, a hardcoded preference will be used (all currently supported providers do not have this issue). A future profile specification will likely replace the hardcoded fallback.
+
+### Bare-bones command
+
+```shell
+ilab model serve
+```
+
+- This command has the same behavior as the one above but the `--model` is whatever the default model path is in the config. This is the existing behavior of `ilab serve` today.
+
+### Llama-cpp backend specific commands
+
+```shell
+ilab model serve --model <PATH> --backend llama-cpp --backend-args '--num-gpu-layers 4'
+```
+
+- This command serves a model with `llama-cpp`.
+- If the model provided is not able to be served by llama-cpp, this command would error out and suggest an alternate backend to use.
+- The existing flags to `ilab serve` (besides `--model-path` & `--log-file`) are now specific to the llama-cpp backend.
+
+### vllm backend specific commands
+
+```shell
+ilab model serve --model <PATH> --backend vllm --backend-args '--chat-template <PATH>'
+```
+
+- This command serves a model with `vllm`.
+- If the path provided is not able to be served by `vllm`, this command would error out and suggest an alternate backend to use.
+- There are [dozens](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#command-line-arguments-for-the-server) of flags for vllm. Whichever arguments the community deems the most important to include, will be added as flags to `ilab model serve vllm`.
+- Any remaining arguments can be specified in the value of the flag `--vllm-args`.
+
+## Testing
+
+An additional end-to-end test will be added that for a new backend for `ilab model serve`. This new test should be triggered whenever code changes to the new backend serving code are made or before a release.
+
+This new test will do the following:
+
+1. Initialize ilab in a virtual env via `ilab config init`.
+2. Download a model via `ilab model download`.
+3. Serve the downloaded model with the new backend via `ilab model serve`.
+4. Generate synthetic data using the served model via `ilab data generate`.
+5. Chat with the served model via `ilab model chat`.
+6. Any future commands that interact with a served model should be added to the test.
+
+Some commands, like `ilab model chat` and `ilab data generate`, serve models in the background as part of the command. If automatic serving of a new backend is implemented for a command, testing of that command that will also be included in the new end-to-end test.
+
+## Handling existing backend-specific commands
+
+The existing `ilab model serve` command has flags that are specific to the `llama-cpp` backend. The current list of flags are:
+
+- `--num-gpu-layers`
+- `--max-ctx-size`
+- `--num-threads`
+
+These flags will be moved to `--backend-args` and will be used as the default arguments for
+`llama-cpp` backend. This will allow for a more consistent experience across backends. The flag will
+be supported up to two releases after the release of the new backend. After that, the flag will be
+removed. During the two releases, a warning will be printed to the user when the flag is used.
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -49,6 +49,7 @@ GiB
 Gmail
 gpu
 Guang
+hardcoded
 hipBLAS
 ilab
 impactful
@@ -124,6 +125,8 @@ Shivchander
 Signoff
 Sigstore
 Srivastava
+subcommand
+subcommands
 subdirectory
 Sudalairaj
 Taj
@@ -145,6 +148,7 @@ USM
 UX
 venv
 Vishnoi
+vllm
 watsonx
 Wikisource
 wikisql


docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Design
+
+Since the subject of the `ilab model serve` command is a model, regardless of the format of the model, every command takes in the `--model` flag or uses its default value in the config.
+
+`ilab model serve` has a subcommand for each backend it can serve models with. As of this draft, the two backends `ilab` would serve with are `llama-cpp` and `vllm`.
+
+This would lead to the commands:
+
+- `ilab model serve llama-cpp`
```
leseb, line 17: As mentioned earlier, this feels unnatural: `ilab model serve llama-cpp`, I would imagine something like `ilab model serve --backend llama-cpp` instead. If additional backends or more complex configurations are added, the command structure might become less clear. Using `--backend` is immediately clear that `--backend` is an option, making the command more self-explanatory.
When designing the CLI, I like to remind myself of the following:

```bash
program action --flag <option>
```

Here llama-cpp is not an action, it's an option. That's how I see things, also understand that people might think otherwise :)

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+```shell
+ilab model serve
+```
+
+- This command has the same behavior as the one above but the `--model-path` is whatever the default model path is in the config. This is the existing behavior of `ilab serve` today.
+
+### Llama-cpp backend specific commands
+
+```shell
+ilab model serve llama-cpp --model <PATH> --llama-cpp-gpu-layers 4 --llama-cpp-max-ctx-size 1024
```
leseb, line 70: A point I've raised before, I'm not sure if it makes sense to prefix everything with `--llama-cpp` if we know we are using the llama-cpp backend.

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+- This command serves a model with `llama-cpp`.
+- If the path provided is not to a `.gguf` file, this command would error out.
```
leseb, line 74: I don't think we should rely on the extension but have a function/script to read the image and tell us the format instead.

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+### vllm backend specific commands
+
+```shell
+ilab model serve vllm --model <PATH> --vllm-chat-template <PATH> --vllm-args "<MORE VLLM SPECIFIC ARGS>"
+```
+
+- This command serves a model with `vllm`.
+- If the path provided is not a directory, this command would error out.
+- There are [dozens](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#command-line-arguments-for-the-server) of flags for vllm. Whichever arguments the community deems the most important to include, will be added as flags to `ilab model serve vllm` with `--vllm-` as a prefix.
+- Any remaining arguments can be specified in the value of the flag `--vllm-args`.
```
leseb, line 86: The only issue with this approach is how can we easily translate a given arg to an internal value? With a flag that is very easy to do. The downside with the flag is that we keep playing catching with every flag vllm produces...

leseb (commented):
Thanks for working on this proposal! I envision an ideal scenario where we can determine the appropriate backend solely based on the model input (directory or file). I'm questioning whether this is something users really need to worry about. If we can alleviate this hassle, I believe we should.

docs/cli/ilab-model-serve.md
alimaredia: I don't think I'm being very clear here, the intention of `--vllm-args` is to not play catch up with every flag vllm has, but instead is intended to be a catch-all flag for all of it's arguments, which is something I think you'd support.

For example the community deams the [vllm arguments](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#command-line-arguments-for-the-server) `--chat-template` and `--download-dir` to be ones that we use often and want to have flags for the subcommand. 

Those would get prefixed with `--vllm-` and become `--vllm-chat-template` and `--vllm-download-dir`. 

If a user wanted to specify more "obscure" arguments to vllm like the `--kv-cache-dtype` and `--block-size` when serving vllm, an example of a command would be:

`ilab model serve vllm --vllm-download-dir <PATH> --vllm-args "--kv-cache-dtype auto --block-size 16"`.

The value of `--vllm-args` is just getting passed straight into the vllm upon startup. The main function of the openai compatible server take and processes all arguments on start up.

docs/cli/ilab-model-serve.md
alimaredia: I can make this statement more general, the point of the bullet is that the command auto detects that it can't serve the format specified and throws an error with feedback for the user.

docs/cli/ilab-model-serve.md
alimaredia: The reason I made it so specific is that I'm scarred by the flags for training that exist in the codebase as of 6/7/24. It's not clear which training flags belong with the mac implementation of training vs the linux implementation and on top of that the flags are very similar.

An example of how the flags could be mixed up. Current ilab serve with llama-cpp has a flag for `max-ctx-size`. `vllm` has a flag for model context length called `--max-model-len` (https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#command-line-arguments-for-the-server). 

We could do the following possibilities:
1. Have a flag for `--max-model-len` and `--max-ctx-size` similar to how we have it in current `ilab train`. I think this option is hardest on the user.

2. Have one flag for context size called `--max-ctx-size` whose value is filtered down to each backend. An issue then becomes what do you do with flags that one backend has that one doesn't. Also you've now added this flag into the set of flags that both backends can use. 

In my current design I propose we make backend agnostic flags those flags have nothing to do with the configuration of the backend itself and more for the configuration of `ilab` and what the backends are operating on (like the model path).

3. Prefix all flags related to a backend with the backend name. I think this is the most clear for the users with the downside that it's more verbose.

@cdoern what do you think about how the flags for training might look. Training will have a similar set of possible subcommands with numerous options.

n1hility: > In my current design I propose we make backend agnostic flags those flags have nothing to do with the configuration of the backend itself and more for the configuration of ilab and what the backends are operating on (like the model path).
 +1 this is the right approach to create and use generic flags yet still have a mechanism for provider specific flags.

docs/cli/ilab-model-serve.md
n1hility: One other point to consider is that it's likely we may have multiple backends that support overlapping formats. To address that we could have a config flag to specify the intended default backend. Although that can be a later design enhancement. I only mention it to show the existing design here is compatible with expansion

docs/cli/ilab-model-serve.md
n1hility: Another example of why this is necessary is that a user also needs to sometimes use hidden behavioral flags that do not have a long lifespan to work around a problem in the backend. Yet another case would be cuda flags that vllm needs to pass through

n1hility (approved):

docs/cli/ilab-model-serve.md
n1hility: IMO either option works well no strong opinion from me :)

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+
+```shell
+ilab model serve --model <PATH> 
+```
+
+- Serves the model at `<PATH>`.
+- If the `<PATH>` is the path for a `.gguf` file then `llama-cpp` is automatically used as the model serving backend.
+- If the path is a directory, `vllm` is automatically used as the backend.
+- The path provided to `vllm` can either be to a Hugging Face repo or a directory with `.safetensors`.
+
```
mairin, line 58: Yes! Love this mechanism. Gives us a way to be opinionated - provide a good default - but also flexible enough that the user who has their own opinions can express them!

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+### Bare-bones command
+
+```shell
+ilab model serve
+```
+
+- This command has the same behavior as the one above but the `--model-path` is whatever the default model path is in the config. This is the existing behavior of `ilab serve` today.
```
mairin, line 65: is there an inconsistency here? Up on line 51 you're calling it `--model` - here on 65 you're calling it `--model-path` ?

docs/cli/ilab-model-serve.md
mairin: @n1hility Agreed. Great idea.

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+- The existing flags to `ilab serve` (besides `--model-path` & `--log-file`) are now specific to the llama-cpp backend and are prefixed with `--llama-cpp-` to indicate that.
+
+### vllm backend specific commands
+
+```shell
+ilab model serve vllm --model <PATH> --vllm-chat-template <PATH> --vllm-args "<MORE VLLM SPECIFIC ARGS>"
+```
+
+- This command serves a model with `vllm`.
+- If the path provided is not a directory, this command would error out.
```
mairin, line 84: Instead of error out, I'd suggest it suggest the correct backend. E.g., 


`You specified vllm but you've provided a gguf file which is not supported by vllm. Would you like to run the gguf file using llamacpp?`

docs/cli/ilab-model-serve.md
mairin: I like the catchall approach. It's one we used to handle arguments to plugins we don't control in the ChRIS project. It has worked well.

mairin (changes requested):

docs/cli/ilab-model-serve.md
leseb: Ok now it's clearer, I'm ok with the `--vllm-args` value being passed as is straight upon vllm startup.

docs/cli/ilab-model-serve.md
leseb: @mairin I don't think the second bullet can apply given that we are going the route of generic flags and not re-using backend-specific flags. This is described here: https://github.com/instructlab/dev-docs/pull/81/files#r1631408310

Other arguments from me on why I don't like the positional argument in this scenario:

* clarity: I know the value that I'm passing to `--backend` is a backend, when I pass `llama-cpp`, I might have no clue what it is
* stability: once we introduce `ilab model serve llama-cpp` we are blocked from adding any other things than a backend. Using `--backend` removes that problem. What if one day we want to add another verb/noun to `ilab model serve <whatever>`.
* consistency: it's a common CLI design practice to use flags to specify options

docs/cli/ilab-model-serve.md
leseb: I'm open to building generic flags, but it feels like we're creating a third layer of abstraction, which might introduce some ambiguity. Users will have to juggle raw flags from llama-cpp, as well as flags from ilab, which could complicate their mental model when writing commands. However, I do like the idea of re-using sensible flags and handling the backend mapping internally.

In the end, we could adopt two strategies:

* Backend-Agnostic Flags: Select the most common flags that make sense across multiple backends.
* Backend-Specific Flags: Use --vllm-args (or similar) for backend-specific configurations.

There might still be issues, such as someone passing ilab model serve --backend llama-cpp --llama-cpp-max-ctx-size 1 --llama-cpp-args "llama-cpp-max-ctx-size 2". However, we can address this through proper validation and error messaging.

docs/cli/ilab-model-serve.md
russellb: One point that helps me decide these choices --

For llamacpp vs vllm, are the rest of the --flags identical? If not I'd prefer separate commands vs a --backend flag. I would prefer only doing something like `--backend` if the rest of the flags apply equally to both (and any future backend)

docs/cli/ilab-model-serve.md
alimaredia: I think you're suggesting that the command would steer the user to use the correct backend instead of the command being run again.

I think this is an awesome idea but out of scope for a first PR of introducing a `vllm` backend. I'll be sure to make a follow up issue for this feature.

docs/cli/ilab-model-serve.md
alimaredia: After doing some more research I'm removing the backend specific prefixes from the backend specific flags. The way we're using `click` for subcommands, flags between backends cannot be shared.

So running `ilab model serve vllm --help` would yield a different message than `ilab model serve --help` and `ilab model serve llama-cpp --help`. 

For this reason I'm not worried about users mixing up arguments as long as there are subcommands.

This is another reason why I believe going subcommands instead of having a  `--backend` flag to choose the backend.

docs/cli/ilab-model-serve.md
alimaredia: This is meant to be `--model` since we serve models not model paths!

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,101 @@
...
+
+- `ilab model serve llama-cpp`
+- `ilab model serve vllm`
+
+The flags for `ilab model serve` are either backend agnostic or specific to a subcommand and are prefixed to indicate that. The intention behind this choice is to ensure there is no mix up of flags between backends.
+
+```console
+ilab model serve
+|
+|_______ --model
```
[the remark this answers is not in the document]
alimaredia: I meant to have this be the flags for either backend, not another subcommand.

docs/cli/ilab-model-serve.md
alimaredia: @leseb Of the arguments you mentioned the one that stuck out to me the most is stability, in the sense that the next subcommand after `ilab model serve` has to be a backend. I understand wanting the flexibility for that subcommand in the future, but I'd argue we see the need for it now so we should just do it. 

Also I don't think it's the end of the world if the possibilities for the next subcommand aren't consistent. In fact if you start at the `ilab model` level, the only commonality is that the next level of sub-commands are related to models. The same principle applies to `ilab model serve`; the subcommands are related to model serving.

leseb: > @leseb Of the arguments you mentioned the one that stuck out to me the most is stability, in the sense that the next subcommand after `ilab model serve` has to be a backend. I understand wanting the flexibility for that subcommand in the future, but I'd argue we see the need for it now so we should just do it.

Honestly, this isn't very future-proof. Since we're starting from scratch, we have the opportunity to design with future growth in mind and enhance user experience clarity.

> 
> Also I don't think it's the end of the world if the possibilities for the next subcommand aren't consistent. In fact if you start at the `ilab model` level, the only commonality is that the next level of sub-commands are related to models. The same principle applies to `ilab model serve`; the subcommands are related to model serving.

Yes, but today we think about `ilab model serve backend.` What if we later realized that the backend was just one of the possible sub-commands? It's not the end of the world, but it will degrade the user experience.

leseb (changes requested):

markstur (commented):
comments inline

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,103 @@
...
+```shell
+ilab model serve
+```
+
+- This command has the same behavior as the one above but the `--model` is whatever the default model path is in the config. This is the existing behavior of `ilab serve` today.
+
+### Llama-cpp backend specific commands
+
+```shell
+ilab model serve --model <PATH> llama-cpp gpu-layers 4 max-ctx-size 1024
```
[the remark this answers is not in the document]
alimaredia: This is a mistake and it should be updated in the most recent commit.

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,103 @@
...
+```console
+ilab model serve
+|
+|_______ --model
+|
+|_______ (other backend agnostic flags)
+|
+|_______ llama-cpp
+|        |
+|        |_______ --llama-cpp-num-gpus
```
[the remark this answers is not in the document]
alimaredia: This was is a mistake that got cleaned up in the most recent commit. There are no flag prefixes anymore per the design doc

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,103 @@
+# Design for `ilab model serve`
+
+## Background
+
+With the [request from the community](https://github.com/instructlab/instructlab/issues/1106) for `ilab` to serve different backends such as [vllm](https://docs.vllm.ai/en/stable/) and the [cli redesign](https://github.com/instructlab/dev-docs/pull/48), this design doc's purpose is to flesh out the behavior of the `ilab model serve` command.
```
russellb, line 5: The CLI redesign link can be updated to point to the merged doc

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,103 @@
...
+
+Specifically the design of subcommand's of `ilab model serve` that apply to different serving backends.
```
russellb, line 7:
```suggestion
Specifically, this doc addresses the design of subcommands of `ilab model serve` that apply to different serving backends.
```

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,103 @@
...
+### Bare-bones but model specific command
+
+```shell
+ilab model serve --model <PATH> 
+```
+
+- Serves the model at `<PATH>`.
+- If the `<PATH>` is the path for a model that can be run by `llama-cpp` then `llama-cpp` is automatically used as the model serving backend.
+- If the `<PATH>` is the path for a model that can be run by `vllm` then `vllm` is automatically used as the model serving backend.
+- If the model at `<PATH>` can be run by either backend, then a default backend defined in the config will be used as the model serving backend.
```
russellb, line 59: What if no backend is specified in the config?

how would express this config option? Config options and flags are the same thing. A config item here implies “—backend <foo>”

docs/cli/ilab-model-serve.md
```diff
@@ -0,0 +1,103 @@
...
+
+### vllm backend specific commands
+
+```shell
+ilab model serve --model <PATH> vllm --chat-template <PATH> --vllm-args "<MORE VLLM SPECIFIC ARGS>"
+```
+
+- This command serves a model with `vllm`.
+- If the path provided is not able to be served by `vllm`, this command would error out and suggest an alternate backend to use.
+- There are [dozens](https://docs.vllm.ai/en/latest/serving/openai_compatible_server.html#command-line-arguments-for-the-server) of flags for vllm. Whichever arguments the community deems the most important to include, will be added as flags to `ilab model serve vllm`.
```
russellb, line 87: It’s probably easier to just have a pass through mechanism. I don’t see much value in replicating their config options in our own.

russellb: I see the next line proposes that already… I’d prefer stating that pass through is the default answer unless we have a really good reason otherwise

russellb (commented):
I left some minor comments. 

It seems the only thing blocking this is whether backend is a positional argument or a flag. 

If it was clear we’d have a bunch of backend specific options, I’d lean toward a positional argument to explicitly separate the set of options for each backend. 

It seems the doc proposes a generic pass through “—vllm-args” so it seems like differences will be minimal and clear enough. 

If I had to pick I’d go with the flag but I’m supportive of either one. We just need to resolve this quickly so implementation can proceed.

docs/cli/ilab-model-serve.md
n1hility: While I was originally neutral, and perhaps leaning towards the subcommand. The recent proposals on the introduction of profiles shifted my perspective quite a bit on this one. 

I argue that with this capability users won't need to (or want to) care which backend is used to serve their model, all they really care about is that there model get served in a way that works and is optimal for the system they are using, and in a way that leads to good ilab results. Even if they do change it, often they will have only one workable choice for their system and use-case. As an example, running the full hifi SDG won't be possible with llama.cpp (at least today), and likewise, vllm is a bit too resource hungry for smaller scale systems.

Based on the above I think we should view this selection as an advanced option / parameter as opposed to a routine subcommand. There's also a compatibility aspect where users get guided into specifying a particular backend (since it's prominent when a subcommand) they might think they need to do this, and we could start to see more scripts and automation tied to it, which would make it harder for us to swap in a different technology we might introduce in the future (e.g say a server optimized for a particular model type, or specific GPU type). Granted this is still a risk with having the option at all, just likely less as an advanced option. 

Finally, I think there's an opportunity where we can make the technology more accessible by presenting higher level, more generic options. For those that are very specific to a particular backend we probably wouldn't want to continually sync ever low-level option, and those could be primarily pass through via the `[--vllm|--llama-cpp-args]` mechanism in this proposal.

Thoughts? 

I know we want to get this in asap, and ultimately I think we can make both approaches work. But I thought i would give this one more check in light of other recent design decisions.

n1hility (changes requested):
Temp requesting changes just to verify the backend vs subcommand discussion

docs/cli/ilab-model-serve.md
n1hility: +1 thats where I land. Ideally generic shared options that are intuitive and then passthrough so we don't have to keep a mirror in sync.

docs/cli/ilab-model-serve.md
n1hility: This another reason  I think we should take another look at just making this an option (consistency between config and flags)

```suggestion
- If the model at `<PATH>` can be run by either backend, then a default backend defined in the config will be used as the model serving backend. In the case where there is ambiguity and a setting is not defined, a hardcoded preference will be used (all currently supported providers do not have this issue). A future profile specification will likely replace the hardcoded fallback. 
```

docs/cli/ilab-model-serve-backend.md
```diff
@@ -0,0 +1,146 @@
...
+  backend: llama-cpp
+```
+
+### Backend flags
+
+The `--backend-args` flag is a string that will be passed to the backend as arguments. This flag is
+used to pass backend-specific arguments to the backend. Each options for the backend will be
+separated by a semicolon. For example, `--backend-args '--num-gpu-layers 4;--max-ctx-size 1024'`.
+This allows for clear separation of arguments for the backend as well as handling flag with commas
+such as `--backend-args '--my-arg a,b,c'`. The backend will be responsible for parsing the arguments.
```
russellb, line 58: how about `--backend-arg` using click's `multiple=True` ? That way the semicolon-based parsing is not necessary. 

https://click.palletsprojects.com/en/7.x/options/#multiple-options

You will still need a way to separate arguments. For example `--backend-flag "foo 123"` makes sense, but we need to split that into two arguments. Splitting on spaces will break arguments that contain spaces.  Splitting on an `=` could work ... but any option could conflict with stuff that should be passed through unmangled.

The simplest implementation would be to always require `--backend-flag`. 

```
ilab model serve --backend vllm --backend-flag foo --backend-flag 123
```

... but that seems very tedious.

Another click option to consider is multi-value options https://click.palletsprojects.com/en/7.x/options/#multi-value-options

You could have `--backend-arg` (takes one arg) and `--backend-arg-opt` (takes 2 args).

```
ilab model serve --backend vllm --backend-arg boolean-arg --backend-arg arg-with-value 123`
```

russellb (dismissed):
I'm good with the direction in general, though I have some comments on the syntax of the `--backend-args` option.

docs/cli/ilab-model-serve-backend.md
leseb: Thanks for raising this, IMO that's an implementation detail and I hope it's not blocking the design. From my perspective we need on the design to agree on the --backend-args approach, then how we implement it can be discussed during code reviews.
But I guess it's my fault and I went on with too many details :).

What do you think?

docs/cli/ilab-model-serve-backend.md
```diff
@@ -0,0 +1,146 @@
...
+`ilab model serve` has a new flag `--backend` that will be used to serve models with. As of this design, the two backends `ilab` would serve with are `llama-cpp` and `vllm`.
+
+This would lead to the commands:
+
+- `ilab model serve --backend llama-cpp`
+- `ilab model serve --backend vllm`
+
+There are specific flags for `ilab model serve` that would apply to all backends. These can be viewed by running `ilab model serve --help`.
+For the specific subcommands, each has their own flags that can be viewed via `ilab model serve <BACKEND> --help`.
+This ensures there is no mix up of flags between backends.
```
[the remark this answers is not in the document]
leseb: Gone.

docs/cli/ilab-model-serve-backend.md
russellb: > But I guess it's my fault and I went on with too many details :).

Right - I'm commenting based on the level of detail specified in the doc.

If you'd like to defer it as an implementation detail, you could remove a bunch of text and just say "we will provide a way to specify arguments to pass through to the backend" or something like that.

I think it's reasonable to cover it here, especially since the top issue debated here has been the UX, and this definitely UX impacting.

alimaredia:
@leseb Thanks for moving this forward., I think `--backend-args` is very simple.

One question I have is do we need to be able to provide the ability to turn all of the knobs in the CLI that we could in a config file? AKA does every parameter need the ability to be set in the config and on the CLI?

I'm guessing the answer is yes, but if that's not that case why don't we just have:
1. Backend agnostic flags: ex `--model`
2. A flag to specify the backend `--backend`
3. A flag to specify the config file `--config`. We have docs for all of the config file options, and populate all values in the `serve` section via the config we generate with `ilab config init`.

At least with this approach there would be one source of truth, the implementation would be clean and simple, and we would not be passing and parsing flags within flags. I've never seen another project with flags similar to my original `--vllm-args` flag, or as the all encompassing `--backend-args` flag. I know it's simple, but my initial reaction is that it seems strange to pass flags within flags.

mairin:
I would say yes for ease of use if you want to override one parameter or
whatnot in the config file.

alimaredia:
The other thing I would say, is just passing the config file might make the world simpler if we we do default config setting with profiles like @cdoern has proposed. The user would only have to focus their attention on the config file. Charlie if you have any thoughts, please chime in.

docs/cli/ilab-model-serve-backend.md
n1hility: +1 ideally we want the design docs to be future referable (at least in the immediate term) to resolve uncertainties in the implementation. It's also ok o specify possible implementation options to help  explain a design as long as you are very clear it's a possible approach to be decided on later, and usually better if you list multiple options. For example, "--backend-args will support specifying one ore more options, possibly using  character delimited values or some other common multi-arg convention. The exact mechanism will be defined in a future proposal."

leseb:
@alimaredia @mairin @n1hility @russellb, I'd like to hold off on discussing `--backend-args` until we have more people present. However, it seems we all agree on introducing a new `--backend` option. Can I get a confirmation on this? All my progress is stuck based on this single decision, and I have several PRs to submit. I'd like to make headway before I go away next week.

Much appreciated.

n1hility (approved):

mairin:
I won't hold anything up but I don't agree with making --backend a flag. I have read back through comments here several times and do not see a rationale for this change. I think it is important to understand why we are going from subcommand to flag, as the subcommand approach was discussed extensively in brainstorming this devdoc and reviewed verbally with multiple folks - even rhtdan felt it was the way to go. I would feel better if I understood *why* we are going to a flag from a subcommand.

My core concern is from the user experience end. The user experience for this project is incredibly important and if we can have a delightful one it will be a differentiator. However, I am not considering other things like what limitations click might have. 

I think having --backendargs as a catch all and starting with that for the implemetation is a great idea

Again, I will not hold this one up bc we need to complete this and have smtg working and that is the most important.

leseb:
Since some of the comments are not visible anymore after a push, I'd like to summarize why using a `ilab model serve --backend` flag is more appropriate than a positional argument like `ilab model serve <backend>`:

* Flexibility: Adding new backends or other options becomes more intuitive and consistent when using flags. It avoids the rigidity of positional arguments, making future expansions or modifications seamless. The flag `--backend` could be ported to other commands later (chat? generate) if we determine it makes sense.

* Consistency: Following a common pattern in CLI design where options are specified with flags ensures uniformity. This approach is clearer and more extendable, aligning with widely accepted CLI design principles.

* Clarity: Using `--backend` makes it immediately apparent that this is an option, enhancing the self-explanatory nature of the command. Users can quickly understand the command's structure and purpose. Also "Flags" signify optional parameters, aligning with the backend's nature as an auto-detected feature. Positional arguments typically indicate **required** inputs, which could confuse users about the backend's necessity. So the `--backend` flag help delineate optional from required parameters more effectively.

* Behavioral Context: Optional parameters, such as --backend, alter the program's behavior without being mandatory. This distinction is better communicated through flags, enhancing user comprehension.

As a side note: I'd argue that the only parameter that should be positional is the model path, as it's essential.

I'd to reference the guide from [CLIG.dev](https://clig.dev/#arguments-and-flags):

>Prefer flags to args. It’s a bit more typing, but it makes it much clearer what is going on. It also makes it easier to make changes to how you accept input in the future. Sometimes when using args, it’s impossible to add new input without breaking existing behavior or creating ambiguity.

In conclusion, using `--backend` for the backend ensures flexibility, consistency, clarity, better integration, and aligns with best practices for CLI design.

Thanks!

@mairin I hope this will change your mind :). Are you keeping your request change to express your disagreement with the latest proposal? I'm seeking clarification because your last comment said you won't hold this up. I'm fine if you want to keep it and have someone else merge if this gets other approvals.

mairin:
Sébastien and I had a call this morning and I am on board with this design now. The rationale had been squashed and he posted it above, and I also misunderstood part of how the flag was proposed to work. I am on board with this now. My main concern was making sure users do not try to use on set of arguments they learned to use with llamacpp and try to xfer them to a vllm or other backend and have a poor experience. Some things we discussed is capturing when that happens and erroring out gracefully, or letting the user know it won't work if we can detect that's going on (with an out to override and try it anyway), and also giving suggesting based on what the argument for backend A is with the equivalent argument for backend B.

mairin (approved):

PR #98, opened 2024-06-17
added sdg api library design

oindrillac:


docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,71 @@
...
+num_samples : 30
+max_retry : 5
+pipeline_steps:
+  gen_q:
+    prompt_template: "configs/gen_q.yaml"
+  filter_q:
+   prompt_template: "configs/filter_q.yaml" 
+max_new_tokens: 10000 
+# model parameters for generation
+model_name: mistralai/Mixtral-8x7B-Instruct-v0.1
```
russellb, line 62: I think there's a need to configure the model name on a per-stage basis. We may have an endpoint configured with more than one model and we want to hit different ones depending on the stage.

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,71 @@
...
+
+* num_samples is the number of synthetic samples that you wish to generate per seed example
+* max_retry is the maximum number of non-greedy retries you want to make if the num_samples is not reached. The number of samples in the generated output will be the samples achieved until max_retry is reached
+* Pipeline steps contains the steps that you want to invoke in the SDG pipeline and the prompt configurations per step. The variable names of the blocks can be anything and the prompt configs must be compatible with the teacher model
+*  max_new_tokens is the maximum number of tokens we want to generate. In other words, the size of the output sequence, not including the tokens in the prompt. 
+*  model_name is the teacher model we would want to use to generate the synthetic data.
+*  model_template and stop_token are the paramaters for generation template.
+
+```
+# run_config.yaml
```
russellb, line 51: Does this need to be a yaml file?

or is it actually a Python dict passed to the library?

I actually think it would be a nice simplification to not worry about config files at all, and from the library perspective, assume configuration is passed in via data structures.

How that config is constructed could be a problem of the library consumer.  Maybe they hardcode it.  maybe they allow a subset to be considered.  Some could be driven by CLI args, for example.

russellb (changes requested):
Some tips for the linter issues

- `make md-lint-fix` will fix most of the md-lint issues. I think there is one type of issue to fix manually after that.
- and then for `make spellcheck` you have to add entries to `.spellcheck-en-custom.txt`

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+The run configuration contains the parameters needed to run the SDG code library as well as points to the templates needed to run the SDG code as well as the prompt template and the default model system prompt template.
+
+* num_samples is the number of synthetic samples that you wish to generate per seed example
+* max_retry is the maximum number of non-greedy retries you want to make if the num_samples is not reached. The number of samples in the generated output will be the samples achieved until max_retry is reached
+* Pipeline steps contains the steps that you want to invoke in the SDG pipeline and the prompt configurations per step. The variable names of the blocks can be anything and the prompt configs must be compatible with the teacher model
+*  max_new_tokens is the maximum number of tokens we want to generate. In other words, the size of the output sequence, not including the tokens in the prompt. 
+*  model_name is the teacher model we would want to use to generate the synthetic data.
+*  model_template and stop_token are the paramaters for generation template.
```
russellb, line 48: Some suggestions for config items based on what the CLI uses today:

https://github.com/instructlab/sdg/blob/7e3a83e3385967828875e95ae5e6f11123e2ab35/src/instructlab/sdg/generate_data.py#L365-L390

- `output_dir` -- where to put the generated samples
- `num_cpus` -- how much to parallelize our inference requests.
- `server_ctx_size` / `chunk_word_count` -- do we need any options to inform how to chunk large knowledge docs?

There's others that I'm not sure we even use ... it's worth a look to see if you spot anything you think should be in the config

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,64 @@
...
+    - pipeline.py 
+    - sdg.py
+```
+
+## CLI
+
+The CLI client uses the instructlab SDG library and provides it a run configuration with input parameters.
+
+```python
+# cli_driver.py
```
russellb, line 29: I interpreted this as sample code for what `ilab` code might look like.

or did you envision something included in the `sdg` repository? The filename here made me stop and think I may have assumed wrong

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,64 @@
...
+  gen_q:
+    prompt_template: "configs/gen_q.yaml"
+  filter_q:
+   prompt_template: "configs/filter_q.yaml" 
+max_new_tokens: 10000 
+# model parameters for generation
+model_name: mistralai/Mixtral-8x7B-Instruct-v0.1
+# generation template
+model_template: <s> [INST] {sys_prompt} {prompt} [/INST]
+stop_token: </s>
```
russellb, line 62: fwiw, this is probably more flexible than absolutely necessary.

This isn't configurable via the CLI today.  We just support knowing if it's one of two known families, `merlinite` or `mixtral` (maybe not the best names, but that's what the code uses).

so the easiest thing to integrate with is allowing a `model_family` config input that would be one of those two things, and then the library should know what to do from there.

This is mostly just a suggested shortcut to reduce the work a little bit in the short term.

docs/sdg/sdg-api-interface.md
oindrillac: * SDG will return generated objects in-memory, and the cli driver will configure where to save the data, and we can have an `output_dir` param to help configure that.
* We can pass `num_cpus` as a param to SDG
* `chunk_word_count` or something similar might be parameter that will pass to the SDG library. @PalmPalm7 @suppathak are looking into chunking.

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,71 @@
...
+
+We propose the following structure for the SDG library. There will be config files that contain all the prompt templates for the pipelines. 
+
+
+```
+- src/instructlab/sdg/
+    - configs/ 
+        - gen_q.yaml 
+        - gen_a.yaml
+        - ...
```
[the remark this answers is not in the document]
oindrillac: @russellb found a way to bundled some non-python files in a source package so we can keep this as a config file. https://github.com/instructlab/schema/tree/main/src/instructlab/schema/v2 - an example

docs/sdg/sdg-api-interface.md
oindrillac: yes that is correct sample code for the CLI or the driver side

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,71 @@
...
+The CLI client uses the instructlab SDG library and provides it a run configuration with input parameters.
+
+
+```
+# cli_driver.py
+
+import sdg
+
+client = openai_client(endpoint)
+cli_sdg = sdg(run_config, client)  # run config has all the variables like num_samples, pipelinesteps etc
```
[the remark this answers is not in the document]
oindrillac: fixed this, PTAL

docs/sdg/sdg-api-interface.md
russellb: This was marked as resolved, but I don't think it got moved.

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,69 @@
...
+max_retry : 5
+pipeline_steps:
+  gen_q:
+    prompt_template: "configs/gen_q.yaml"
+  filter_q:
+   prompt_template: "configs/filter_q.yaml" 
+max_new_tokens: 10000 
+# model parameters for generation
+model_name: mistralai/Mixtral-8x7B-Instruct-v0.1
+# generation template
```
russellb, line 65:
```suggestion
# generation template - optional - will be determined based on model name if it is a known type
```

docs/sdg/sdg-api-interface.md
russellb: I put a suggested edit above to address this -- just noting that it's optional and that we can pick the right one based on the model name if not specified. This will work for the cases we care about in the short term. Having the option in place if needed seems good too.

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,69 @@
...
+    - llmblock.py 
+    - pipeline.py 
+    - sdg.py
+```
+
+![example API interface](../images/sdg-api-interface.png)
+
+## CLI
+
+The CLI client uses the instructlab SDG library and provides it a run configuration with input parameters.
```
russellb, line 28:
```suggestion
The CLI client uses the instructlab SDG library and provides it a run configuration with input parameters. The following represents a sample of what Library usage could look like.
```

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,69 @@
...
+from sdg import SDG
+import yaml
+
+client = openai_client(endpoint)
+with open('run_config.yaml', 'r') as file:
+    run_config = yaml.safe_load(file)
+cli_sdg = SDG(run_config, client)  # run config has all the variables like num_samples, pipelinesteps etc
+generated_samples = cli_sdg.generate()
+```
+
```
russellb, line 42:
```suggestion

As an initial integration milestone, we will modify the `generate_data` function in `instructlab.sdg.generate_data` to make use of the updated SDG API. This is the function the `ilab` CLI already uses, so modifying this implementation will allow us to get the updated SDG API in place without disrupting the CLI integration.

CLI integration will require additional changes later to allow passing in customizations to the SDG pipeline, but we will treat that as a follow-up implementation milestone.

```

russellb (changes requested):
Thanks for the updates! There are just two minor comments left to resolve.

- moving `model_name` to be per-step
- noting that `model_template` and `stop_token` can be optional and defaulted to the right thing for the subset of models `ilab` supports right now.
- I suggested an edit to describe two stages of CLI integration:
  1. changing `generate_data()` without impact to the CLI code
  2. adding new options to the CLI to allow custom pipeline extensions

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,69 @@
...
+The run configuration contains the parameters needed to run the SDG code library as well as points to the templates needed to run the SDG code as well as the prompt template and the default model system prompt template.
```
leseb, line 43:
```suggestion
The run configuration includes the necessary parameters for executing the SDG code library. It specifies the templates required for running the SDG code, the prompt template, and the default model system prompt template.
```

docs/sdg/sdg-api-interface.md
```diff
@@ -0,0 +1,69 @@
...
+
+* `num_samples` is the number of synthetic samples that you wish to generate per seed example.
+* `max_retry` is the maximum number of non-greedy retries you want to make if the `num_samples` is not reached. The number of samples in the generated output will be the samples achieved until `max_retry` is reached.
+* Pipeline steps contains the steps that you want to invoke in the SDG pipeline and the prompt configurations per step. The variable names of the blocks can be anything and the prompt configurations must be compatible with the teacher model.
+* `max_new_tokens` is the maximum number of tokens we want to generate. In other words, the size of the output sequence, not including the tokens in the prompt.
+* `model_name` is the teacher model we would want to use to generate the synthetic data.
+* `model_template` and `stop_token` are the parameters for generation template.
```
leseb, line 50: Should we describe both independently?

docs/sdg/sdg-api-interface.md
oindrillac: looks good!

docs/sdg/sdg-api-interface.md
oindrillac: described!

aakankshaduggal:
> its breaking on `customizations` @russellb how can I update the spellcheck library?
You might want to update here --
https://github.com/instructlab/dev-docs/blob/main/.spellcheck-en-custom.txt

docs/sdg/sdg-api-interface.md
oindrillac: @russellb made updates to this to accept a client, model-name and default model prompt

oindrillac:
https://github.com/instructlab/dev-docs/pull/103 <- so that Spell Check passes on this PR

aakankshaduggal (approved):

docs/sdg/sdg-api-interface.md
russellb: @oindrillac but it's still not specified on a per-stage basis?

aakankshaduggal (approved):

oindrillac:
done, added this here

russellb (changes requested):
please see the open comment thread about the config format and model name

docs/sdg/sdg-api-interface.md
oindrillac: the value will be a class variable that gets passed on when an object is initialized. we have also decided to keep the run config in a .py file as per your suggestion let me update the doc

oindrillac: PTAL @russellb

oindrillac: cc: @shivchander @aakankshaduggal

aakankshaduggal (approved):

russellb (approved):
thank you!

PR #103, opened 2024-06-24
updated spell check list

oindrillac:


aakankshaduggal (approved):

russellb (changes requested):
This should just be added to #98

PR #104, opened 2024-06-24
Add doc for dedicated local storage in ilab cli

RobotSail:
This document outlines an approach for config and program data management that simplifies
the user flow and creates a streamlined experience.

Signed-off-by: Oleg S <97077423+RobotSail@users.noreply.github.com>


docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
+# InstructLab CLI Local Storage Proposal
+
+## Why is this needed?
+
+For tools to be useful to the user, they generally need to work on and know about data pertinent to the user. For example, podman and docker track image blobs, kubectl tracks cluster credentials, etc.
+
+The InstructLab CLI manages a number of settings today such as the teacher model, taxonomy location, inference model, etc. using a local file known as `config.yaml`.
```
russellb, line 7: The CLI itself doesn't manage any of this. It's user config.  It generates an initial file for you, but never updates it from there.  I only bring this up because user config isn't the same as data that is never touched directly by a user.  (config vs data)

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+isn't sustainable as data is frequently moved around unbeknownst to the program. This leaves the user with two choices: either update
+the config to correctly point to either a relative path in the format
+of `../../uncle-dir/dataset.jsonl`, or worse, have to specify an absolute path.
+
+
+## What's being proposed?
+
+We propose the following:
+
+1. `ilab` will have a dedicated config directory seated at `~/.config/ilab` on Linux/MacOS systems, and the `%APPDATA%` equivalent if we decide to support Windows.
```
russellb, line 22: presumably this works on windows with WSL? I know people have been using it on WSL, already.

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+## Scope of this proposal
+
+
+This proposal provides a long-term goal of numerous components and how we expect them to interact with each other.
+
+The main point here is to move the config to live in a single place so that the `ilab` command can be executed from any directory with confidence. In addition, we require the use of refs and CRUD-style commands to simplify the management and interaction of data generated by `ilab`.
+
+The steps outlined to implement this are:
+
+1. Implement logic for `~/.config/ilab` and `~/.local/share/ilab` directories and expect the config to live and be read from there unless another config is manually specified. For simplicity, if the is read in from `~/.config/ilab` then we assume the relative paths should be realtive to the data directory `~/.local/share/ilab`.
```
russellb, line 174: A way this can work is have a data_dir that can be changed via a CLI flag or config file option, but always default to what you have specified here -- very similar to what you describe as a default location for the config dir unless otherwise specified via a flag

russellb (changes requested):
I'm generally happy with the direction here.

I'd like to hear a little more around the proposed unique refs -- how would you generate them? where would they be stored? How would they be mapped to a file (or files) on disk?

Are you intentionally leaving it vague and just "TBD based on whatever seems easy to implement" later?  or do you have anything more concrete in mind?

There's also some linter complaints, but I'm sure you saw those already.

docs/cli/ilab-local-config.md
RobotSail: If it's WSL then we should still assume it's a linux system. The only WSL-specific bits are more on the CUDA drivers side. And even then, that works fine currently.

RobotSail: I was thinking more of "raw windows", but you're right - WSL is a great alternative.

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
+# InstructLab CLI Local Storage Proposal
+
+## Why is this needed?
+
+For tools to be useful to the user, they generally need to work on and know about data pertinent to the user. For example, podman and docker track image blobs, kubectl tracks cluster credentials, etc.
```
nathan-weinberg, line 5:
```suggestion
For tools to be useful to the user, they generally need to work on and know about data pertinent to the user. For example, `podman` and `docker` track image blobs, `kubectl` tracks cluster credentials, etc.
```

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+The InstructLab CLI manages a number of settings today such as the teacher model, taxonomy location, inference model, etc. using a local file known as `config.yaml`.
+
+While this method allows ilab to avoid additional complexity of managing a dedicated directory, it’s limited by its ability to scale beyond constraining the user to run ilab in a single directory.
```
nathan-weinberg, line 9:
```suggestion
While this method allows ilab to avoid additional complexity of managing a dedicated directory, it’s limited by its ability to scale beyond constraining the user to run `ilab` in a single directory.
```

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+Since the new ilab CLI will need to not only manage datasets, downloaded models, and any intermediary files, it'll also have to make it simple for the user to orchestrate these things.
```
nathan-weinberg, line 10: What is "the new ilab CLI"?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+We presently store these paths directly in the config, however this
+isn't sustainable as data is frequently moved around unbeknownst to the program. This leaves the user with two choices: either update
+the config to correctly point to either a relative path in the format
+of `../../uncle-dir/dataset.jsonl`, or worse, have to specify an absolute path.
```
nathan-weinberg, line 15: Why is this worse?

docs/cli/ilab-local-config.md
nathan-weinberg, line 22:
```suggestion
1. `ilab` will have a dedicated config directory seated at `~/.config/ilab` on Linux/MacOS systems, and the `%APPDATA%` equivalent if Windows support is added in the future.
```

nathan-weinberg, line 22: And should this not be `~/.config/instructlab` and not `~/.config/ilab`?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+## What's being proposed?
+
+We propose the following:
+
+1. `ilab` will have a dedicated config directory seated at `~/.config/ilab` on Linux/MacOS systems, and the `%APPDATA%` equivalent if we decide to support Windows.
+2. `ilab` will have a dedicated program files directory seated at `~/.local/share/ilab` which will store model checkpoints, downloaded models, etc.
```
nathan-weinberg, line 23: Same comment as above - should this not be `~/.config/instructlab` and not `~/.config/ilab`?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+of `../../uncle-dir/dataset.jsonl`, or worse, have to specify an absolute path.
+
+
+## What's being proposed?
+
+We propose the following:
+
+1. `ilab` will have a dedicated config directory seated at `~/.config/ilab` on Linux/MacOS systems, and the `%APPDATA%` equivalent if we decide to support Windows.
+2. `ilab` will have a dedicated program files directory seated at `~/.local/share/ilab` which will store model checkpoints, downloaded models, etc. 
+3. Expectation of having config migrations in the future as it's evolved on an as-needed basis.
```
nathan-weinberg, line 24: What does this mean?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+For these reasons, we propose that this data will have a permanent home at `~/.local/share/ilab`. 
+
+This is a rough outline of all of the data pieces consumed and generated by different points in the process:
+
+| phase | consumed data | generated data |
+|-------|---------------|----------------|
+| SDG | QnA files from taxonomy, teacher model checkpoint | knowledge dataset for training, skills dataset for training |
+| Training | Input dataset, base model | Model checkpoints, processed dataset to be consumed during training |
+| Eval | Model checkpoint being evaluated | JSON data / logs containing the score |
```
nathan-weinberg, line 50:
```suggestion
| Evaluation | Model checkpoint being evaluated | JSON data / logs containing the score |
```

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+| Publish | Model checkpoints to be published | n/a |
```
nathan-weinberg, line 51:
```suggestion
| Publishing | Model checkpoints to be published | n/a |
```

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+Based on this, it's evident that we need a few different subdirectories:
+
+| path | description |
+|------|-------------|
+| `base_models/` | Stores the base models that we download |
+| `checkpoints/` | Houses the checkpoints we generate during training. Can potentially split these out further into `checkpoints/phase00`, `checkpoints/phase05`, and `checkpoints/phase10` |
```
nathan-weinberg, line 59: Why would we do this splitting? Which approach is the design doc actually proposing?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+| `datasets` | Would contain the generated datasets. Each item here would be a directory consisting of: `train.jsonl` and `test.jsonl` |
+| `logs` | Would contain logs from various `ilab` runs |
+| `eval_data` | Contains scores from various evaluation runs for a given model ref |
+
+
+## How ilab would manage this data
+
+There should be a ***simple and intuitive*** way to view & manage all of this data.
+
+The basic idea is that each piece of information can be discovered through a `list` command in some form. Each item in the listing will have the corresponding associated information, along with a unique reference that the user can use when interacting with subsequent commands.
```
nathan-weinberg, line 69: I love this consideration! Thank you for including it. Will we also expose to the user where exactly these things are living on their filesystem someone, either via the `list` command, documentation, etc?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+Finally, we would add a new `ilab model delete <ref>`  command 
+which would enable us to pass a reference to the model we want
+to delete.
+
+
+### Managing datasets
+
+New datasets would be generated through `ilab generate` (or whatever the current equivalent is) and the corresponding dataset would be saved under the `datasets/` directory. Each model would also have a unique ref associated with it.
+
+Following this, we could list out the datasets with `ilab dataset list` and further filter them out with either `knowledge` and `skills`, or `phase00`, `phase05`, and `phase10` depending on which SDG scheme is being used.
```
nathan-weinberg, line 105: Is `dataset` a scoped command group? We already have `data` and `model` - do we need another group?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+Here's a sample output of the `ilab eval list` command:
+
+```
+MODEL REF   BASE MODEL                    EVAL SCORE  METRIC     EVAL DATE 
+abcde      	ibm-granite/granite-7b-base   84.2%       MMLU       2024-06-24
+cbdea      	abcde                         78.9%       MMLU       2024-06-23
+fghij       ibm-granite/granite-7b-base   7.89        MT-Bench   2024-06-22
+```
+
+Since this would be JSON data, we don't expect it to be very expensive or consume a lot of space, so clearing it should  not be a priority.
```
nathan-weinberg, line 143:
```suggestion
Since this would be JSON data, we don't expect it to be very expensive or consume a lot of space, so clearing it should not be a priority. 
```

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+abcde      	ibm-granite/granite-7b-base   84.2%       MMLU       2024-06-24
+cbdea      	abcde                         78.9%       MMLU       2024-06-23
+fghij       ibm-granite/granite-7b-base   7.89        MT-Bench   2024-06-22
+```
+
+Since this would be JSON data, we don't expect it to be very expensive or consume a lot of space, so clearing it should  not be a priority. 
+
+
+
+But if it's desired, we could provide an `ilab eval clear` function that erases all of the evaluation data. We could also generate refs for the eval runs as well, and clear them using `ilab eval clear <eval-ref>`.
```
nathan-weinberg, line 147:
```suggestion
We could provide an `ilab eval clear` function that erases all of the evaluation data. We could also generate refs for the eval runs as well, and clear them using `ilab eval clear <eval-ref>`.
```

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+The output would look something like
```
nathan-weinberg, line 149: ???

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+## Evolving configs
+
+
+The config object is something we expect to change in the future. To maintain compatibility and decrease user frustration, we should have schema versions for our configs.
```
nathan-weinberg, line 155:
```suggestion
The config object is something we expect to change in the future. To maintain compatibility, we should have schema versions for our configs.
```
Should also explain _why_ it's expected to change

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+For instance, our current one would be a `v1` and if we decide
+to add something in the future we would update it to a `v2`.
+
```
nathan-weinberg, line 159:
```suggestion
Our initial config would be `v1` and if we decide
to add something in the future we would update it to a `v2` and so on.
```

docs/cli/ilab-local-config.md
nathan-weinberg, line 174: Several typos here that need to be fixed

nathan-weinberg (changes requested):

docs/cli/ilab-local-config.md
RobotSail: The one we're currently building for RHEL AI.

docs/cli/ilab-local-config.md
RobotSail: It's fragile and very unfriendly to the user.

docs/cli/ilab-local-config.md
RobotSail: Personally I think yes but I want to be consistent since the CLI is called ilab.

docs/cli/ilab-local-config.md
RobotSail: DB migrations but for configs.

docs/cli/ilab-local-config.md
RobotSail: Because we produce many checkpoints so it may become cumbersome to view all of them vs. for a specific phase.

docs/cli/ilab-local-config.md
RobotSail: I'm not familiar with data but it sounds like that's where it would go under.

docs/cli/ilab-local-config.md
nathan-weinberg: Absolute paths?

docs/cli/ilab-local-config.md
nathan-weinberg: Since the package is called `instructlab` I think it should be `instructlab`

RobotSail: You don't need to sell me on it, let's change it!

docs/cli/ilab-local-config.md
nathan-weinberg: Can you update the wording here to be clearer? "as it's evolved on an as-needed basis." doesn't make a lot of sense

docs/cli/ilab-local-config.md
nathan-weinberg: So which approach are you actually proposing here? If it's the former can you add this explanation as to why we might want the latter in the future?

docs/cli/ilab-local-config.md
nathan-weinberg: Sounds good to me!

RobotSail:
> I'd like to hear a little more around the proposed unique refs -- how would you generate them? where would they be stored? How would they be mapped to a file (or files) on disk?
> 
> Are you intentionally leaving it vague and just "TBD based on whatever seems easy to implement" later? or do you have anything more concrete in mind?

I was thinking that we could just use a simple UUID or SHA2 digest. For the initial version we could have the refs just be the relative paths in the `~/.local/share` directory.

docs/cli/ilab-local-config.md
nathan-weinberg: If that's the case please update the proposal to reflect that

docs/cli/ilab-local-config.md
RobotSail: This is a good point, I would say it makes a stronger point for the proposed structure of the `~/.local/share` directory.

docs/cli/ilab-local-config.md
RobotSail: Yes I agree with that.

docs/cli/ilab-local-config.md
RobotSail: How would you rephrase this?

docs/cli/ilab-local-config.md
RobotSail: What happens to your config when you move that file?

docs/cli/ilab-local-config.md
RobotSail: Sure, I'll add that.

docs/cli/ilab-local-config.md
nathan-weinberg: It breaks, but I don't see how that's worse than a relative path breaking.

RobotSail: @nathan-weinberg Think through the motions of using this method. 
First you have to run `ilab init` and make sure your config is in the same directory always. If you're running `ilab` again, you have to either make a new config or navigate back to where that config exists.

Afterwards, you either download a new dataset or generate one from SDG. In either case, you will need to specify what the path to the dataset will be. If you decide to move the dataset or run ilab in another directory, well now you'll need to again either navigate back to where your `config.yaml` exists or have to generate a new one, or specify a `--data-path`. 

Now let's say you decide to rename a file or directory, now you'll need to go back and update all of your config files.
It becomes very tedious to do, and not what we want our users to experience.

docs/cli/ilab-local-config.md
nathan-weinberg: Well that depends on what you are trying to say - if the "new ilab CLI" is one that consumes the new libraries, I would put it in the context of that, since that is something that is ubiquitous both upstream and downstream.

docs/cli/ilab-local-config.md
nathan-weinberg: Thank you!

docs/cli/ilab-local-config.md
nathan-weinberg: I agree! I'm definitely not against the approach. It's more the wording of "absolute paths are worse than relative paths when it comes to breaking things" - honestly just removing that language would be fine with me.

docs/cli/ilab-local-config.md
RobotSail: I like that more, I'll update it.

docs/cli/ilab-local-config.md
RobotSail: Either can be done here. I was hoping to leave this portion open for comments to hear the case for both. But with how we want ilab to work, leaving it as `knowledge` vs. `skills` will probably be fine. If we want to support `phase00` in the CLI then we can always revisit this.

nathan-weinberg: Ah I hear ya - honestly I trust your expertise here! Fine with either approach.

docs/cli/ilab-local-config.md
nathan-weinberg: Can you implement this? :smile:

docs/cli/ilab-local-config.md
nathan-weinberg: This was not removed in the most recent changeset

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+Additionally, The `ilab model train` command would be able
+to receive a ref to a dataset for easy access.
+
+
+### Managing Eval Runs
+
+When we run evaluations, we expect to receive a certain value which is tied to a model checkpoint.
+
+Under this new framework, the evaluation outputs could be saved under the `eval_data` subdirectory, where `ilab` could later use to get the score for a particular model.
```
nathan-weinberg, line 127: Just as a note for reference, I have implemented this somewhat in https://github.com/instructlab/instructlab/pull/1369

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+DATASET ID            PHASE TYPE        SIZE    NUM SAMPLES    TEACHER MODEL
+phase00-dataset-id    knowledge_base    500MB   650k           mistralai/mixtral-8x7B
+phase05-dataset-id    knowledge         230MB   330k           mistralai/mixtral-8x7B
+phase10-dataset-id    skils             500MB   550k           mistralai/mixtral-8x7B
+```
+
+An `ilab data delete <ref>` command would be added as well
+so that the user can easily delete generated datasets.
+
+Additionally, The `ilab model train` command would be able
```
nathan-weinberg, line 119:
```suggestion
Additionally, the `ilab model train` command would be able
```

docs/cli/ilab-local-config.md
nathan-weinberg: Don't think this made it into the latest changeset

nathan-weinberg (changes requested):

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+isn't sustainable as data is frequently moved around unbeknownst to the program. This leaves the user with two choices: either update
+the config to correctly point to either a relative path in the format
+of `../../uncle-dir/dataset.jsonl`, or worse, have to specify an absolute path.
+
+
+## What's being proposed?
+
+We propose the following:
+
+1. `ilab` will have a dedicated config directory seated at `~/.config/instructlab` on Linux/MacOS systems, and the `%APPDATA%` equivalent if Windows support is added in the future.
```
n1hility, line 22: Is your intent to support XDG?:
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#introduction

Not arguing that you should necessarily but if you intended to I would change the text to be seated at $XDG_DATA_HOME/instructlab (.local/share/instructlab)

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+Since this would be JSON data, we don't expect it to be very expensive or consume a lot of space, so clearing it should not be a priority. 
+
+
+
+We could provide an `ilab eval clear` function that erases all of the evaluation data. We could also generate refs for the eval runs as well, and clear them using `ilab eval clear <eval-ref>`.
+
+The output would look something like
+
+
+## Evolving configs
```
n1hility, line 152: This seems like a distinct proposal, was there a reason you included it?

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+the config to correctly point to either a relative path in the format
+of `../../uncle-dir/dataset.jsonl`, or worse, have to specify an absolute path.
+
+
+## What's being proposed?
+
+We propose the following:
+
+1. `ilab` will have a dedicated config directory seated at `~/.config/instructlab` on Linux/MacOS systems, and the `%APPDATA%` equivalent if Windows support is added in the future.
+2. `ilab` will have a dedicated program files directory seated at `~/.local/share/instructlab` which will store model checkpoints, downloaded models, etc.
```
n1hility, line 23: To carry through on the Windows spec I would say %LOCALAPPDATA% here.

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+The steps outlined to implement this are:
+
+1. Implement logic for the `~/.config/instructlab` and `~/.local/share/instructlab` directories, and expect the config to live and be read from there unless another config is manually specified. For simplicity, if the config we read in is located at `~/.config/instructlab` - the default location - then we'll assume relative paths should be resolved with respect to the data directory `~/.local/share/instructlab`.
+1. Implement logic to the existing commands so that datasets, models, and eval scores can be read from `ilab list`
+1. Implement ref logic and update all existing commands to be able to resolve objects from refs
+
+
+Based on these steps, #1 would be the trickiest because we assume that the paths referenced in the config file are relative to the calling process. 
+
+One approach here could be removing these paths from the default and forcing the user to specify them relatively.
```
n1hility, line 181: I am having trouble understanding what you mean here

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,184 @@
...
+
+Another approach would be implementing the list commands at the same time as #1 and setting it up so that the relative paths passed to `ilab` are relative ***to the `~/.local/share/instructlab` directory***
```
n1hility, line 183: Is there any compatibility with existing configs? For example it looks in the current dir and if there is a config.yml expects abs paths, or is this an intentional major change

n1hility: What would the experience be like without #3 (implementing refs) in your list?

n1hility (changes requested):
Overall I really like this proposal, but could use some clarifications on a few points made.

docs/cli/ilab-local-config.md
n1hility: Would list tell you were a file was in that case?

RobotSail: This would be a breaking change then but we'd need to make it. We can maintain compatibility by allowing users to provide a `--config` flag which should read in the config as we do today. Not sure how much harder this would be versus what we're doing today.

docs/cli/ilab-local-config.md
RobotSail: Thanks for the clarification!

docs/cli/ilab-local-config.md
RobotSail: @n1hility My wording here is bad but the idea is simple:

- When the config read is the default one `~/.config/ilab/config.yaml` then parse paths relative to `~/.local/share/ilab`.
- When that's NOT the config, we would parse relative paths with respect to the calling process. E.g. if `ilab` is invoked in `/foo` and the `config.model_path` references a file in `bar/granite_7b.gguf` then it would resolve to `/foo/bar/granite_7b.gguf`.
- All absolute paths will be parsed as-is

docs/cli/ilab-local-config.md
RobotSail: Yes this is definitely distinct, but will be something we'll need as a feature in the future.

docs/cli/ilab-local-config.md
RobotSail: Yes that's what I was thinking of.

docs/cli/ilab-local-config.md
RobotSail: Yes! Although as @n1hility pointed out, we'd want this to be in a separate PR / dev-doc since the config evolution touches on a different topic.

nathan-weinberg: SGTM!

docs/cli/ilab-local-config.md
hickeyma: That might be so @RobotSail but as @russellb mentioned the CLI doesn't manage this.

docs/cli/ilab-local-config.md
hickeyma: @RobotSail Are qwe going to support native Windows? if not, we should be explicit that it is WSL.

docs/cli/ilab-local-config.md
```diff
@@ -0,0 +1,173 @@
...
+## Scope of this proposal
+
+
+This proposal provides a long-term goal of numerous components and how we expect them to interact with each other.
+
+The main point here is to move the config to live in a single place so that the `ilab` command can be executed from any directory with confidence. In addition, we require the use of refs and CRUD-style commands to simplify the management and interaction of data generated by `ilab`.
+
+The steps outlined to implement this are:
+
+1. Implement logic for the `~/.config/instructlab` and `~/.local/share/instructlab` directories, and expect the config to live and be read from there unless another config is manually specified. For simplicity, if the config we read in is located at `~/.config/instructlab` - the default location - then we'll assume relative paths should be resolved with respect to the data directory `~/.local/share/instructlab`.
```
hickeyma, line 159: Should it be that relative paths are resolved with respect to the config directory?

hickeyma (changes requested):
I think overall this is a good proposal and direction. Thanks for pushing this @RobotSail.

Some additional points:

- As raised by @n1hility,  I think supporting the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is a good idea. Do you mind adding this to the proposal? You could then mention support for the environment variables as per the XDG Base Directory Specification: `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME`. 
- It was mentioned about capability to change the data and config paths. This could be implemented by providing environment variables like `$INSTRUCTLAB_PATH_CONFIG` and `$INSTRUCTLAB_PATH_CONFIG`

docs/cli/ilab-local-config.md
tiran: I strongly recommend to use https://pypi.org/project/platformdirs/ . It has the correct defaults for all important platforms.

I also like to point out that the use of XDG directories does not work well with systemd services. @mairin just mentioned that she wants systemd service support for InstructLab. XDG directories are problematic, because they are designed for human end-users with a home directory. `systemd --system` need an entirely different directory structure.

mairin: @tiran Thanks for pointing this out! I had initially suggested XDG thinking of the laptop-targeted experience. But if we want a systemd service we'll need to use system directories. Ideally we'd use XDG for the laptop-target user experience and system-wide directories for the big mama GPU hw experience.

mairin: ^ to this end, @tiran suggests `I would store the paths in config.yaml and have a option to either use XDG or a common base directory for all paths.`

danmcp:
@RobotSail Would you be willing to update this doc based on feedback and current state?  Would be good to get it merged given the state of the code has already moved past the state of this doc.

RobotSail:
@danmcp Yes we can update this one to reflect what currently exists in ilab

danmcp:
> @danmcp Yes we can update this one to reflect what currently exists in ilab

Thank you!

nathan-weinberg:
@RobotSail are you still planning on updating this?

PR #105, opened 2024-06-25
Update knowledge-submissions-past-wikipedia.md

lhawthorn:
Added new approved Knowledge submission data sources, updated status for several.

Updated documentation with new process to take in requested knowledge sources to be open a PR against this devdoc.

Related to issue #59 which should be closed once this PR is reviewed and merged.

Closes #59

PR #107, opened 2024-06-26
Update spelling per #105

jjasghar:
PR #105 needs some new words added to the spelling dictionary.

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -26,6 +26,7 @@ cuBLAS
 CUDA
 customizations
 CWD
+Cynefin
 dataset
 DCO
 Dependabot
@@ -56,6 +57,7 @@ ilab
 impactful
 Inferencing
 instructlab
+io
 ISA
 Jie
 JIT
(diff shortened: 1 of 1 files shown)


markstur (changes requested):
Knowlege and butvis are both typos so let's not add those

Also... YIKES!...  a re-sort regarding order of uppercase vs lower happened and so this PR is a big re-sort again.  Not good.  Let's not go back and forth on that.  PR should be simple additions and if we need to fix the sort again let's do that in a commit whichever order is preferred(?).

jjasghar:
i just used `:<,'>sort` 😭 

I'm assuming that would be this would be the same with `sort` too right?

markstur:
> i just used `:<,'>sort` 😭
> 
> I'm assuming that would be this would be the same with `sort` too right?

`:%sort i` would be case-insensitive to keep the current a's with A's, bees with Bees.  🐝 

add a u for bonus removal of dups.
I don't really have a preference (okay maybe case-insensitive) but don't like it thrashing back and forth

jjasghar:
OK, i got it fixed :)

markstur (approved):
such a pretty diff, thanks!

Sorry for the extra work.  I'm thinking maybe I should not have allowed the last re-sort so you kind of paid the price on this one.

nathan-weinberg (approved):

Issue #87, opened 2024-06-10
Add spelling ignore


RobotSail:
The current pyspelling script will error on paths like the virtualenv since this isn't defined in its
ignorelist. It would be nice if it could be configured to only run on a given set of directories,
rather than the entire project.

For example:
```
--------------------------------------------------------------------------------

Misspelled words:
<context> venv/lib/python3.11/site-packages/pyspelling-2.10.dist-info/licenses/LICENSE.md
--------------------------------------------------------------------------------
MERCHANTABILITY
NONINFRINGEMENT
sublicense
--------------------------------------------------------------------------------

Misspelled words:
<context> venv/lib/python3.11/site-packages/Markdown-3.6.dist-info/LICENSE.md
--------------------------------------------------------------------------------
MERCHANTABILITY
Redistributions
Stienstra
Takhteyev
--------------------------------------------------------------------------------

Misspelled words:
<context> venv/lib/python3.11/site-packages/wcmatch-8.5.2.dist-info/licenses/LICENSE.md
--------------------------------------------------------------------------------
MERCHANTABILITY
NONINFRINGEMENT
sublicense
--------------------------------------------------------------------------------

Misspelled words:
<context> venv/lib/python3.11/site-packages/soupsieve-2.5.dist-info/licenses/LICENSE.md
--------------------------------------------------------------------------------
MERCHANTABILITY
NONINFRINGEMENT
sublicense
--------------------------------------------------------------------------------

Misspelled words:
<context> venv/lib/python3.11/site-packages/bracex-2.4.dist-info/licenses/LICENSE.md
--------------------------------------------------------------------------------
MERCHANTABILITY
NONINFRINGEMENT
sublicense
--------------------------------------------------------------------------------
```

nathan-weinberg:
@RobotSail are you interested in doing this?

Issue #86, opened 2024-06-10
Create targets in makefile to automatically install the pyspelling dependency if it doesn't exist


RobotSail:
Currently in the Makefile, we have `spellcheck` as a target which calls the `pyspelling` tool. 

When the tool isn't installed, running the target results in an error. It'd be nice if the makefile tried to automatically download it for you.

nathan-weinberg:
Some good examples of this kind of implementation in https://github.com/instructlab/instructlab and https://github.com/instructlab/sdg

bjhargrave:
I am not sure why we have Makefiles in any of the instructlab projects since they are primarily python projects. I know we need c/c++ tooling to build llama_cpp for the cli, but I suspect that is the exception and thus we shouldn't require users to have a make command present for the other projects.

RobotSail:
@bjhargrave Dev-docs is primarily a doc repo with some tooling that happens to be implemented in Python. IMO I think it's fine to keep it as Makefile, especially because of how simple it is.

nathan-weinberg:
@bjhargrave AFAIK none of the Makefile targets have anything of particular concern to most users. At most they allow for running CI checks that will happen on PRs anyway.

bjhargrave:
I am questioning why they exist in the repos at all. At best they are duplicative of other processes which are used in the proper builds. And so they will get out of date as official processes change. They seem to be only a convenience for some maintainers (who are into Makefiles) rather then proper, maintained parts of the projects.

It seems like we need a policy on the use of Makefiles :-) Who are they for? What is their purpose if they are not part of the proper build processes? Or should they be part of the proper build processes so that we can know they are always valid and up-to-date. Should there be consistency across repos in the provided targets?

nathan-weinberg:
To be honest, I feel like it's a non-issue - but if you want to open up an issue in this repo for that to continue the discussion, feel free. I don't want to derail @RobotSail's original issue here.

RobotSail:
> It seems like we need a policy on the use of Makefiles

Why? They're simple, they work.

nathan-weinberg:
@RobotSail are you interested in doing this?

Issue #77, opened 2024-06-06
Define policy for adding or updating Python dependencies


tiran:
We should create a policy and guiding document how to deal with new dependencies or updates of existing dependencies.

Some thoughts:

- new dependencies should be avoided unless they bring merit. Any new dependencies adds a permanent overhead.
- new dependencies must be reviewed for license, compliance, security, and overall community best practices
- new dependencies and updates must be coordinated with stakeholders for all hardware targets (Nvidia CUDA, AMD ROCm, Intel Gaudi). Gaudi is a bit problematic, because the stack needs downstream builds and forks from Intel.

leseb:
I like the idea. Additionally, is it worth running an inspection of the current deps as well and see if some could be removed too?

bjhargrave:
Most of the repos other than `instructlab` have dependabot keeping dependencies up-to-date. Eventually we need to add dependabot to the `instructlab` certainly for the workflows but also for the python dependencies.

russellb:
> Most of the repos other than `instructlab` have dependabot keeping dependencies up-to-date. Eventually we need to add dependabot to the `instructlab` certainly for the workflows but also for the python dependencies.

Yes! I've been meaning to look at this, but haven't gotten around to it. It would be great if you were interested in working on this at some point!

bjhargrave:
> It would be great if you were interested in working on this at some point!

It is on my dance card 🕺

nathan-weinberg:
We also need to ensure any dependencies that are added are compliant from a licensing POV - @mrutkows is our resident expert on that AFAIK :smile:

Issue #61, opened 2024-05-22
Feature Request: Support system-style questions (transferred from Taxonomy)


roy651:
(Issue was moved from [here](https://github.com/instructlab/taxonomy/issues/1106) as it touches multiple areas)
When defining a new skill in the pattern of question-context-answer it seems that ilab will generate all of those when it creates the augmented dataset. The implication is that they are all dynamically generated.

I would like to propose that ilab will support an additional entry in the skill's yaml allowing for a "system"-style which will preface the question when generating the answer. This part should be static and need not be generated for each new entry. Most likely, it should be static for the entire skill, hence it should reside in the top level of the yaml with the "task_description".

I think that this capability is something that may be required by additional skills as it follows, the now so common, system/user pattern and I assume others would require that too.

roy651:
@obuzek per your comment on the other [thread](https://github.com/instructlab/taxonomy/issues/1106#issuecomment-2123265438):
I understand the mechanism you describe and I acknowledge that you're utilizing the "system" prompt as part of the SDG.
However, I think that, this design choice is something that should be relatively easy to resolve with a suitable instruction to the SDG ("Do not generate this part"...).
Assuming technically it's feasible and the discussion is about the need - One might argue that the set of Q&A (and context) samples should suffice for the model to learn the necessary "rules" and pattern of the specific skill.
I do think that this need is valid and quite widespread. As the use of LLMs expands and it not necessarily plain conversation-like, and as the complexity of input and output patterns rises, we see frequent deviations of the model from the required patterns.
I expect to see others come up with a similar request.

bjhargrave:
> yaml with the "task_description".

The task_description is already placed in the prompt. So I think it can do what you are requesting. It can include instructions to the teacher model.

roy651:
@bjhargrave 
This is not well documented but assuming that the prompt is the one that shows up on diagram 3 in the [paper](https://arxiv.org/pdf/2403.01081), then I'm not sure it's relevant, cause the task description is part of the first line and not directly related to the generated question. 
Furthermore, I imagine that that both the model that's generating the responses and the evaluator model should get this static text as part of their prompt in a relevant slot.

PR #109, opened 2024-07-02
Propose adding a file format for custom pipelines

markmc:
Related to instructlab/instructlab#1546

See instructlab/sdg#86 for an implementation


docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,101 @@
...
+## Proposal
+
+We will add an API that instantiates a `Flow` object from a file:
+
+```python
+flow = Flow.from_file('mycustomflow.yaml', params)
+block_configs = flow.render()
+```
+
+The YAML file format will mirror the API and look like this:
```
russellb, line 25: I wonder if we should put a version field on this in case we need to break compatibility in the future?

https://github.com/instructlab/sdg/issues/54

I hope it would also serve as a reminder to anyone poking around in the code that the format is not arbitrary and changes must consider compatibility.

markmc: Yes! I even put that in the objectives and then did nothing about it!

russellb: and I should have read the whole thing properly before commenting :)

russellb:
For the spellcheck job to be happy ...


```
$ cat << EOF >> .spellcheck-en-custom.txt
Params
TODO
configs
freeform
iters
merlinite
mixtral
num
src
templating
yaml
EOF

$ make spellcheck-sort
```

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,101 @@
...
+### Model Prompts
+
+Based on whether model_family is mixtral or merlinite, a different prompt is used with the teacher model
+
+```python
+_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST]"
+_MODEL_PROMPT_MERLINITE = "'<|system|>\nYou are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior.\n<|user|>\n{prompt}\n<|assistant|>\n'"
+```
+
+TODO: figure out how to include the appropriate prompt in the block config.
```
russellb, line 68: Maybe render() should be a method on the parent class, that calls something else implemented by the child class that does the stuff now, but also some validation and default setting.

For a block that doesn't have the model prompt set, we can set it based on either the explicit model family, or our best guess based on the model name (_get_model_family() does the logic here already).

Right now we are duplicating this logic everywhere, but really it could just be the default behavior unless specified otherwise.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Prompt Config Files
+
+Every LLMBlock references a separate prompt config file, and presumably a custom pipeline will provide custom prompt configs too.
+
+These prompt config files are quite simple YAML files - they contain a single object with system, introduction, principles, examples, and generation keys. See e.g. src/instructlab/sdg/configs/skills/freeform_questions.yaml
+
+TODO: figure out how we want to handle these.
```
russellb, line 76: The ones included in the library get included in the Python package and installed alongside the python source. There's a little bit of magic in the code to make sure we load it from the right place from the library's installed location.

For a custom flow, I think a path to a yaml file is still good. Where the files are installed, how they get there ... out of scope for the purposes of the CLI and sdg library, I think.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Model Serving
+
+Custom pipelines may have more unique model serving requirements. Instead of serving just one model, we may need to launch the model server with a model and an additional model with adapter.
+
+How should this be expressed in a flow configuration? Can it be inferred, or are explicit options needed?
```
russellb, line 82: I think there's two parts to this.

Right now the sdg code takes two parameters for each block:

- the openai client instance
- a model ID

I believe these are sufficient. `vllm`, for example, can host both a model and a model+adapter under two different model IDs.

There is another question, maybe separate from this doc, of how the CLI figures out what it's supposed to launch using the internal model serve APIs to do this. I don't know how to handle this yet.

One idea ... instead of a config file loaded by the sdg library, we could make this config file CLI-specific. The config file could have a section that describes the model serving requirements:

```yaml
serve:
- model_id: foo
   model_path: mixtral
- model_id: bar
   model_path: path/to/mixtral
   adapater_path: path/to/adapter

block_configs:
...
```

In this case, we'd probably do all of the config file loading in the `instructlab` repo. Instead of `Flow.from_file()` it could be, `Flow.from_dict()` or something like that.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,101 @@
...
+- model_family - e.g. mixtral or merlinite
+- model_id - a path name for the specific teacher model being used
+- num_iters - how many samples to generate
+- batched - whether the model serving backend supports batched mode
+
+TODO: figure out how these should be supplied to the blocks, and whether any YAML templating based on these are necessary.
+
+### Model Prompts
+
+Based on whether model_family is mixtral or merlinite, a different prompt is used with the teacher model
```
gabe-l-hart, line 61: It would be interesting to extend the config format above to allow for extensions to the `model_family` prompts as well. This could allow users to define prompt templates for other models without needing them hard-coded in the library.

gabe-l-hart: Also, it looks like these prompt templates are largely encoding the chat template for the model. For most models that I've seen, the chat template is actually encoded in the model metadata and applied server-side when using the `chat/completions` API. Has there been thought around using the chat API rather than applying the template client-side to the lower-level `completions` API?

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,101 @@
...
+```python
+flow = Flow.from_file('mycustomflow.yaml', params)
+block_configs = flow.render()
+```
+
+The YAML file format will mirror the API and look like this:
+
+```yaml
+block_configs:
+  - block_type: LLMBlock
```
gabe-l-hart, line 29: It looks like this proposal does not attempt to solve extending the available `block_type` implementations, right? Probably not something that needs to be solved at this point, but it seems that this would be a good opportunity to use a factory pattern of some sort. I can imagine two possible ways to do this:

1. Add a custom `metaclass` to the [Block](https://github.com/instructlab/sdg/blob/main/src/instructlab/sdg/block.py#L17) base class that maintains a class-attribute dict of `name` -> `class` for all available classes, then require users with custom block types to get their blocks imported _before_ instantiating the flow.
2. Allow the `block_type` string to include a qualified import path (e.g. `my_extension.blocks.my_block.MyBlock` and then use `importlib` to do something like `MyBlock = getattr(importlib.import_module(block_type.rsplit(".", 1)[0]), block_type.rsplit(".", 1)[-1])`

gabe-l-hart (commented):
Hi Mark! Russell pointed me at this draft since I'm plugged into various flavors of "how do we extend SDG" on the IBM Research side. I like the direction a lot and just have a few ideas/thoughts.

docs/sdg/sdg-flow-yaml.md
russellb: right - no attempt to allow custom block types right now, just configuring what the lib includes

docs/sdg/sdg-flow-yaml.md
russellb: haven't seen anyone propose it, honestly didn't even realize the difference myself :)

probably outside the scope of this doc, though

gabe-l-hart: Yes, definitely outside the scope of this proposal! Just something that would hopefully get simpler in this proposal if it were done more broadly

docs/sdg/sdg-flow-yaml.md
markmc: Added in commit 1781b8a thanks

markmc:
> For the spellcheck job to be happy ...

Fixed in commit 590cc8e thank you

docs/sdg/sdg-flow-yaml.md
markmc: Only `LLMBlock` uses `model_prompt`, so I think `model_family` should be passed to `LLMBlock` and it will choose the correct prompt

At least for now ... if we find out later that we need to supply custom prompts, then I guess a custom flow author will need to provide prompts for each model serving backend

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,125 @@
...
+These prompt config files are quite simple YAML files - they contain a single object with system, introduction, principles, examples, and generation keys. See e.g. src/instructlab/sdg/configs/skills/freeform_questions.yaml
+
+TODO: figure out how we want to handle these.
+
+### Model Serving
+
+Custom pipelines may have more unique model serving requirements. Instead of serving just one model, we may need to launch the model server with a model and an additional model with adapter.
+
+How should this be expressed in a flow configuration? Can it be inferred, or are explicit options needed?
+
```
russellb, line 107: Here are some thoughts around integrating this into the CLI. There are some immediate steps and possible work we'd do in future iterations to improve the UX.

```suggestion
### CLI Integration

As of the current version of `ilab`, it supports `simple` and `full` as parameters to `--pipeline` to select one of the two types of built-in pipelines included in the library. Once we have support for loading custom pipelines, we need a way for these to be specified with the CLI. Here is a proposed CLI UX for this:

> `ilab data generate`

Use the default pipeline, `simple`.

> `ilab data generate --pipeline full`

Use the built-in `full` pipeline.

> `ilab data generate --pipeline full --pipeline-extend path/to/config.yaml`

Use the `full` pipeline, but extend it with the pipeline defined in the provided config file.

> `ilab data generate --pipeline none --pipeline-extend path/to/other_config.yaml`

Do not use a built-in pipeline. Use only the pipeline in the custom configuration.

### Future CLI Improvements

A possible improvement would be to have a well-defined place on the filesystem where custom pipeline configs can be automatically loaded and included as options to the `--pipeline` parameter.

For example, if the config format included new parameters, `name: full-extended` and `extends: full`, and the CLI discovered and loaded it automatically, we could support `--pipeline full-extended` without needing the additional `--pipeline-extend` option.

`/usr/share/instructlab/sdg/` is a proposed location for this as a place for a distribution of InstructLab to include pre-defined custom pipelines, at least for Linux. See the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html) for more details on why this path is appropriate for this use-case.

It would also make sense to support a configuration directory for user's own custom pipeline configurations. Assuming there is a base config directory, these could go in a `sdg` subdirectory. There is a separate proposal that discusses a proposed configuration location: <https://github.com/instructlab/dev-docs/pull/104>. Note this is separate from the distribution-provided, read-only pipelines discussed above with a different location.
```

docs/sdg/sdg-flow-yaml.md
markmc: Resolved with:

> We will continue to use these config files unchanged, and custom files can be specified with an absolute path. Relative paths are assumed to be relative to the Python package e.g. `configs/skills/...`.

docs/sdg/sdg-flow-yaml.md
markmc: > One idea ... instead of a config file loaded by the sdg library, we could make this config file CLI-specific.

In #86, I converted the default flows to YAML ... and it works well I think. And since it works well, it makes sense to me to not have 2 different ways of defining flows, which helps enforce a level of parity between built-in and custom flow capabilities.

markmc: Resolving with:

> Right now the sdg code takes two parameters for each block - the OpenAI client instance, and the model ID. This should be sufficient. vLLM, for example, can host both a model and a model+adapter under two di\
fferent model IDs.
>
> In future, it may make sense to	define some model serving config (e.g. which adapter to use for SDG) as	part of	a flow definition. We will address this	in a future if enhancement, if needed.

docs/sdg/sdg-flow-yaml.md
markmc: Moving to `chat/completions` would definitely be a separate enhancement, thanks for the suggestion though - I've filed it as #87

markmc: > It would be interesting to extend the config format above to allow for extensions to the `model_family` prompts as well. This could allow users to define prompt templates for other models without needing them hard-coded in the library.

I guess this would be a "support custom model families" enhancement? Or custom model prompts for known families?

Probably best to file another issue on this to explain the use case.

docs/sdg/sdg-flow-yaml.md
markmc: Agree, custom block types is a whole other thing. Definitely would be interested to hear the use case.

docs/sdg/sdg-flow-yaml.md
russellb: Standardizing on the yaml format everywhere sounds good to me

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+> `ilab data generate --pipeline full`
+
+Use the built-in `full` pipeline.
+
+> `ilab data generate --pipeline full --pipeline-extend path/to/config.yaml`
+
+Use the `full` pipeline, but extend it with the pipeline defined in the provided config file.
+
+> `ilab data generate --pipeline none --pipeline-extend path/to/other_config.yaml`
```
hickeyma, line 126:
```suggestion
> `ilab data generate --pipeline-extend path/to/other_config.yaml`
```
This seems clunky. Would it be more usable without `pipeline` specified? Therefore, specifying the extended pipeline with no pipeline specified then it is assumed extended only.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
+# SDG API :: Add a file format for defining custom Flows
```
leseb, line 1: Was the double colon intentional?

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+flow = Flow.from_file('mycustomflow.yaml', params)
+block_configs = flow.render()
+```
+
+The YAML file format will mirror the API and look like this:
+
+```yaml
+version: 1.0
+block_configs:
+  - block_type: LLMBlock
```
leseb, line 30: Given that `block_type` is already in a `block_configs` scope, how about we simply call this `type` and avoid redundancy?

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+    block_config: # LLMBlock constructor kwargs
```
leseb, line 31: ditto, `config`?

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+      block_name: gen_knowledge
```
leseb, line 32: ditto: `name`?

docs/sdg/sdg-flow-yaml.md
leseb, line 30: Why do we list by Type and not by Name instead? it feels more natural (at least to me). So we would have:

```
version: 1.0
block_configs:
  - name: gen_knowledge
    type: LLMBlock
...
```

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+        raise IncompatibleVersionError("The custom flow file format is from a future major version.")
+    elif major <= PARSER_MAJOR and minor > PARSER_MINOR:
+        logger.warning("The custom flow file may have new features that will be ignored.")
+```
+
+### Flow Params
+
+The following parameters are supplied to a Flow constructor and used by the render() method:
+
+- client - an OpenAI completions API client for talking to the teacher model via the serving backend (i.e.  llama-cpp or vLLM)
```
leseb, line 75:
```suggestion
- client - an OpenAI completions API client for talking to the teacher model via the serving backend (i.e. llama-cpp or vLLM)
```

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+- model_family - e.g. mixtral or merlinite
+- model_id - a path name for the specific teacher model being used
+- num_iters - how many samples to generate
```
leseb, line 78: I'm not sure if `num_iters` conveys well its description `how many samples to generate`.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+
+Do not use a built-in pipeline. Use only the pipeline in the custom configuration.
+
+### Future CLI Improvements
+
+A possible improvement would be to have a well-defined place on the filesystem where custom pipeline configs can be automatically loaded and included as options to the `--pipeline` parameter.
+
+For example, if the config format included new parameters, `name: full-extended` and `extends: full`, and the CLI discovered and loaded it automatically, we could support `--pipeline full-extended` without needing the additional `--pipeline-extend` option.
+
+`/usr/share/instructlab/sdg/` is a proposed location for this as a place for a distribution of InstructLab to include pre-defined custom pipelines, at least for Linux. See the [Filesystem Hierarchy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html) for more details on why this path is appropriate for this use-case.
```
leseb, line 136: If we have a location with pipeline examples then a nice to have would be to have a `ilab date generate --list-pipelines` perhaps? Not necessarily in the scope of this design but food for thought.

leseb (changes requested):
Thanks for the proposal, a few comments inline.

docs/sdg/sdg-flow-yaml.md
markmc: I agree it's clunky, I don't love it

But I think it's important `--pipeline` has a default value that isn't affected by other arguments - so your suggestion reads "extend the default pipeline" IMO

@russellb

docs/sdg/sdg-flow-yaml.md
markmc: I like double colon! Changed it for you :heart:

docs/sdg/sdg-flow-yaml.md
markmc: Yep, agree - I was avoiding changing the existing API, but you're right ... now's the chance to make it not suck

docs/sdg/sdg-flow-yaml.md
markmc: Definitely in the realm of API change here, but #77 will change it to `num_instructions_to_generate` :+1:

docs/sdg/sdg-flow-yaml.md
markmc: Added

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,157 @@
...
+
+Every LLMBlock references a separate prompt config file, and presumably a custom pipeline will provide custom prompt configs too.
+
+These prompt config files are quite simple YAML files - they contain a single object with system, introduction, principles, examples, and generation keys. See e.g. src/instructlab/sdg/configs/skills/freeform_questions.yaml
+
+We will continue to use these config files unchanged, and custom files can be specified with an absolute path. Relative paths are assumed to be relative to the Python package e.g. `configs/skills/...`.
+
+### Future - Model Serving
+
+Custom pipelines may have more unique model serving requirements. Instead of serving just one model, we may need to launch the model server with a model and an additional model with adapter.
```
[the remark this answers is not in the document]
markmc: Updated this section based on the discussion in instructlab/sdg#86

docs/sdg/sdg-flow-yaml.md
russellb: I think having a default value for `--pipeline` that changes depending on other values seems less desirable than having to specify `--pipeline none`. This is probably the least likely combination of values to use, anyway.

Another option could be allowing both file paths and the aliases (`simple`, `full`) as arguments to `--pipeline`. So a custom pipeline that isn't extending one could just be:

```
ilab data generate --pipeline path/to/custom_pipeline.yaml
```

russellb: or yet another alternative, drop `--pipeline-extend` and just allow `--pipeline` to be specified multiple times.

```
ilab data generate --pipeline simple
```

```
ilab data generate --pipeline path/to/custom_pipeline.yaml
```

```
ilab data generate --pipeline full --pipeline path/to/extensions.yaml
```

docs/sdg/sdg-flow-yaml.md
leseb: I'm honored, thank you Sir!

leseb (approved):
Thanks!

docs/sdg/sdg-flow-yaml.md
hickeyma: Ok, aybe its best to be explicit with `pipeline`. I am ok to leave it as is for now.

hickeyma (approved):
LGTM, thanks @markmc, nice proposal

markmc:
Thanks for the reviews.

Please see also #61 - I'd like to use this file format change as an opportunity to further simplify by removing the "Flow" noun from the API

The implication for this dev doc is that I would change it from talking about defining "custom flows" to "custom pipelines" ... the doc already uses these terms interchangeably, which just highlights the confusion

russellb (approved):
looks great - thank you!

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+We will add an API that instantiates a `Pipeline` object from a list of files:
+
+```python
+pipeline = Pipeline.from_flows(ctx, ['mycustomflow.yaml'])
+```
+
+The YAML file format will mirror the API and look like this:
+
+```yaml
```
xukai92, line 26: this is essentially reverting some design change/decision we made as per the discussion in https://github.com/instructlab/dev-docs/pull/98: the flow config was originally proposed and implemented as a YAML file with the almost same structure but during the review process it was suggested to be presented as Python codes

i strongly believe this design PR should attribute this properly and probably should be named as something like "reviving YAML config files"

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+    if major > PARSER_MAJOR:
+        raise IncompatibleVersionError("The custom flow file format is from a future major version.")
+    elif major <= PARSER_MAJOR and minor > PARSER_MINOR:
+        logger.warning("The custom flow file may have new features that will be ignored.")
+```
+
+### Flow Params
+
+The following parameters are supplied to a Flow constructor and used by the render() method:
```
xukai92, line 70: i don't think client should be part of the config, nor num_instructions_to_generate and batched.
why? because the same flow should be triggered at runtime with a flexible combination of those

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+Based on whether model_family is mixtral or merlinite, a different prompt is used with the teacher model
+
+```python
+_MODEL_PROMPT_MIXTRAL = "<s> [INST] {prompt} [/INST]"
+_MODEL_PROMPT_MERLINITE = "'<|system|>\nYou are an AI language model developed by IBM Research. You are a cautious assistant. You carefully follow instructions. You are helpful and harmless and you follow ethical guidelines and promote positive behavior.\n<|user|>\n{prompt}\n<|assistant|>\n'"
+```
+
+For now, we assume that the `LLMBlock` class will choose the appropriate model prompt based on the family and that there is no need to specify a custom prompt.
+
+### Prompt Config Files
```
xukai92, line 91: why files over strings as in the YAML itself?
(i'm not saying the latter is better, just wondering how the decision is made here)

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+Every LLMBlock references a separate prompt config file, and presumably a custom pipeline will provide custom prompt configs too.
+
+These prompt config files are quite simple YAML files - they contain a single object with system, introduction, principles, examples, and generation keys. See e.g. src/instructlab/sdg/configs/skills/freeform_questions.yaml
+
+We will continue to use these config files unchanged, and custom files can be specified with an absolute path. Relative paths are assumed to be relative to the Python package e.g. `configs/skills/...`.
+
+### Future - Model Serving
+
+Custom pipelines may have more unique model serving requirements. Instead of serving just one model, we may need to launch the model server with a model and an additional model with adapter. vLLM, for example, can host both a model and a model+adapter under two different model IDs.
```
xukai92, line 101: I think this (and the above with endpoint per workflow) over complicates the design.
we should simply assume that a single endpoint can serve all models we need (which is fairly simply to achieve as one can either using vllm to achieve so or can wrap multiple model endpoints with a single one to router them models)

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+The pipeline author needs some way of disambiguating between these multiple models - i.e. the definition of each `LLMBlock` needs to specify a particular model.
+
+Right now the `Pipeline` constructor takes two relevant parameters - the OpenAI client instance, and the model ID for the default model. It's important to note that this model ID is defined by the user at runtime, and it may not match the model IDs that the pipeline author used.
+
+The use cases will be:
+
+1. Most LLMBlock definitions will use the default teacher model - and we can make the semantic that if the pipeline author doesn't specify a model in an `LLMBlock`, the default in `PipelineContext.model_id` is used.
+2. In cases where a model+adapter is served, it should usually be the pipeline author defining both the model serving configuration and the model's use in an `LLMBlock`. So the pipeline author can name the model in the serving config and reference it by that name in the pipeline config.
```
xukai92, line 110: as for vllm, when model + adapter is used, it's basically a different `model_id` from the endpoint

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+> `ilab data generate`
+
+Use the default pipeline, `simple`.
+
+> `ilab data generate --pipeline full`
+
+Use the built-in `full` pipeline.
+
+> `ilab data generate --pipeline full --pipeline-extend path/to/config.yaml`
```
xukai92, line 127: again to repeat my previous comment, I don't think it makes sense to have a way to chain workflows like this and also this is very much out of the scope of this design PR.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+In future, it may make sense to define some model serving config (e.g. which adapter to use for SDG) as part of a flow definition. We will address this in a future if enhancement, if needed.
+
+### CLI Integration
+
+As of the current version of `ilab`, it supports `simple` and `full` as parameters to `--pipeline` to select one of the two types of built-in pipelines included in the library. Once we have support for loading custom pipelines, we need a way for these to be specified with the CLI. Here is a proposed CLI UX for this:
+
+> `ilab data generate`
+
+Use the default pipeline, `simple`.
```
xukai92, line 121: I think this default should be achieved as `ilab data generate --pipeline PATH_TO_THE_SIMPLE_WORKFLOW_YAML` for the purpose of unification.
we can provide alias like simple or full to the built-in YAML files to make users' like easy.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+### No Custom Flows
+
+It would be preferable to not support custom flows, especially so early in the project because:
+
+- We will need an extensive API to support this customization, and we will need to be careful about making incompatible changes to that API once it has been adopted.
+- We would learn more about the pipelines that users are creating if they were added to the library.
+
+This approach was discarded because of strong demand from downstream users to define custom flows to encapsulate proprietary pipeline configuration..
+
+### Custom Flows as Code
```
xukai92, line 160: I'd like to hear @russellb's comments on why this is no longer preferred as the time when https://github.com/instructlab/dev-docs/pull/98 was reviewed. in particular, what changed your mind to think YAML files are more preferable now?
I think some justification is helpful for the community to understand why opposite decision have been made during the development/design of the SDG library.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+> `ilab data generate --pipeline none --pipeline-extend path/to/other_config.yaml`
+
+Do not use a built-in pipeline. Use only the pipeline in the custom configuration.
+
+### Future CLI Improvements
+
+A possible improvement would be to have a well-defined place on the filesystem where custom pipeline configs can be automatically loaded and included as options to the `--pipeline` parameter.
+
+For example, if the config format included new parameters, `name: full-extended` and `extends: full`, and the CLI discovered and loaded it automatically, we could support `--pipeline full-extended` without needing the additional `--pipeline-extend` option.
```
xukai92, line 139: again I think this is out of the scope of this design doc but I feel this worth its own way to achieve "compose" (where chaining is just the simplest composition).

docs/sdg/sdg-flow-yaml.md
xukai92: in other words, those are running details but don't affect the "semantics" of the flow.

xukai92 (changes requested):

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+- Library users can either use a custom flow standalone, or combine a custom flow with existing flows.
+- The file format and library can evolve substantially without making breaking changes.
+- Incompatible changes can be introduced while retaining support for existing custom flows for a deprecation period.
+
+## Proposal
+
+We will add an API that instantiates a `Pipeline` object from a list of files:
+
+```python
+pipeline = Pipeline.from_flows(ctx, ['mycustomflow.yaml'])
```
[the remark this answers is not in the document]
markmc: It comes from the observation that within the library (in `generate_data()`) we are creating a pipeline that chains together `MMLUBenchFlow` and `SynthKnowledgeFlow`

docs/sdg/sdg-flow-yaml.md
markmc: I wasn't involved, but from what I can see in the PR, it went like this

The originally proposed design was:

```
with open('run_config.yaml', 'r') as file:
    run_config = yaml.safe_load(file)
cli_sdg = SDG(run_config, client)  # run config has all the variables like num_samples, pipelinesteps etc
generated_samples = cli_sdg.generate()
```

and Russell asked:

> Does this need to be a yaml file?
>
> or is it actually a Python dict passed to the library?
>
> I actually think it would be a nice simplification to not worry about config files at all, and from the library perspective, assume configuration is passed in via data structures.
> 
> How that config is constructed could be a problem of the library consumer. Maybe they hardcode it. maybe they allow a subset to be considered. Some could be driven by CLI args, for example.

So it sounds like Russell anticipated there would be custom pipelines defined by YAML files, but that would be something implemented in the CLI codebase rather than the library.

You can see Russell still saw it that was [when reviewing this PR](https://github.com/instructlab/dev-docs/pull/109#discussion_r1664438995)

> instead of a config file loaded by the sdg library, we could make this config file CLI-specific.

Because he's thinking about where the model serving config will go, and that it probably shouldn't be in a config file format implemented by the SDG library

I responded with:

>In #86, I converted the default flows to YAML ... and it works well I think. And since it works well, it makes sense to me to not have 2 different ways of defining flows, which helps enforce a level of parity between built-in and custom flow capabilities.

and Russell's response was:

> Standardizing on the yaml format everywhere sounds good to me

(i.e. I'm saying "I hacked on this and, in practice, YAML works well for defining pipelines in all cases" and Russell is saying "indeed it does, I see that now")

I'll look for a way to incorporate all of this into the doc :+1:

docs/sdg/sdg-flow-yaml.md
markmc: Absolutely :+1: 

I've started referring to these as "runtime parameters", and this is why I've proposed the `PipelineContext` object

When I say "The following parameters are supplied to a Flow constructor", I'm referring to the current code, not what I'm proposing

But I'm not being very clear here. I will update :+1:

docs/sdg/sdg-flow-yaml.md
markmc: No strong opinion from me on this - it just carried over from the current `Flow` design

But I could speculate - the content of these files would overwhelm the pipeline configuration and make the pipelines hard to reason about, and perhaps there's a use case for re-using the same prompt configs in different pipelines?

docs/sdg/sdg-flow-yaml.md
markmc: Yes, that is the assumption (mostly Russell's assumption, he contributed this bit). the sentence isn't intended to suggest differently

It simply says that a pipeline will interact with more than one model. Yes, through the same endpoint

docs/sdg/sdg-flow-yaml.md
markmc: Yes. The challenge here is there are multiple personas involved

1. The pipeline author needs some way to express "talk to model A to generate qna and talk to model B to evaluate" or whatever
2. The user can launch vLLM in various ways, and we need some way to map "model A" and "model B" to the user's model serving config

That's really all this section is about

docs/sdg/sdg-flow-yaml.md
markmc: Some good discussion on that [here](https://github.com/instructlab/dev-docs/pull/109#discussion_r1668463920)

Somehow I didn't update the doc with the latest proposal there - chaining would be done by specifying `--pipeline` multiple times

But, personally, I'd be happy to omit that feature for now until there's a strong use case for it

docs/sdg/sdg-flow-yaml.md
markmc: Yeah, `simple` and `full` should be thought of as aliases

But it is a bit more complex, and it had occurred to me we'll need to get into this as part of the CLI integration

Currently, the aliases look more like this:

```
full:
    knowledge:
      - MMLU_BENCH_FLOW
      - SYNTH_KNOWLEDGE_FLOW
    freeform_skills:
      - SYNTH_FREEFORM_SKILLS_FLOW
    grounded_skills:
      - SYNTH_GROUNDED_SKILLS_FLOW
simple:
    knowledge:
      - SIMPLE_KNOWLEDGE_FLOW
    freeform_skills:
      - SIMPLE_FREEFORM_SKILLS_FLOW
    grounded_skills:
      - SIMPLE_GROUNDED_SKILLS_FLOW
```

So we need to be able to express a `(knowledge, freeform_skills, grounded_skills)` tuple really ...

docs/sdg/sdg-flow-yaml.md
markmc: I think I've got that [above](https://github.com/instructlab/dev-docs/pull/109#discussion_r1671708336) but of course happy to hear more from Russell

The change is whether the YAML format should be implemented on the CLI side or in the library, and I think Russell's initial preference for the former was because he was also thinking about the serving config ... something we haven't fully figured out yet, in fairness

In any case, I'll make an update to the doc to cover this

docs/sdg/sdg-flow-yaml.md
markmc: Noted and responded to [here](https://github.com/instructlab/dev-docs/pull/109/files/f7663a515bb6535a536e43ee5839dc4913d1e96d#r1671729188)

And ... this is under "Future CLI Improvements". It's merely acknowledging ways the design might evolve in future, which is a useful thing to do in a design doc rather than simply reject as out of scope any ideas for the future from contributors to the discussion

docs/sdg/sdg-flow-yaml.md
russellb: This all is intended to support the downstream to desire to ship custom pipeline content that would run after “full”. If we don’t expose extending in some way, we would ship a copy of “full” plus more from somewhere else and have to try to keep them in sync when fixes or improvements come up in one place or the other. 

Both work. The chaining thing lets us define stuff once. Indeed it adds some complexity to the CLI interface, though. 

Either way we have to be clear on how we intend to deal with that use case. Happy to hear other alternatives!

markmc: > This all is intended to support the downstream to desire to ship custom pipeline content that would run after “full”.

Wow, I did not realize that - I thought the downstream pipeline was standalone, no dependence on an upstream pipeline. My bad. I'll capture that in the doc. (Unless Kai corrects us on this point, of course)

russellb: We could make it that way. But the input to the custom parts is the output of “full” … so it’s this or we copy stuff or something else 🤷‍♂️

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Prompt Config Files
+
+Every LLMBlock references a separate prompt config file, and presumably a custom pipeline will provide custom prompt configs too.
+
+These prompt config files are quite simple YAML files - they contain a single object with system, introduction, principles, examples, and generation keys. See e.g. src/instructlab/sdg/configs/skills/freeform_questions.yaml
+
+We will continue to use these config files unchanged, and custom files can be specified with an absolute path. Relative paths are assumed to be relative to the Python package e.g. `configs/skills/...`.
+
+### Future - Model Serving
```
russellb, line 99: I wrote more on this topic over here. I did it separately, but by the end I felt like maybe it's content to be integrated back into this one.

https://github.com/instructlab/dev-docs/pull/115

The good news is that I think what we need is already supported by `ilab`.

markmc: Yeah, at the very least I need to capture here that this will work:

> ```
> version: "1.0"
> models:
>   - name: myfunkyadaptor
>     description: a funky adaptor for generating questions
> block_configs:
>   - block_type: LLMBlock
>     block_config:
>       block_name: gen_questions
>       config_path: configs/skills/freeform_questions.yaml
>       add_num_samples: True
>       gen_kwargs:
>         model_id: myfunkyadatpor
>       output_cols:
>         - question
>     drop_duplicates:
>       - question
> ```

And then the serving config just needs to have `--lora-modules=myfunkyadapter=path/to/my_custom_adapter`

docs/sdg/sdg-flow-yaml.md
markmc: Captured in commit 4425f8d

docs/sdg/sdg-flow-yaml.md
markmc: Ok, see commit 6457751 for how this impacts the pipeline config files

I started to incorporate #115 but it felt like the main point of that doc is the `config.yaml` impact, and it's better to describe that separately

docs/sdg/sdg-flow-yaml.md
markmc: I think this is cleared up now with commit 6457751

docs/sdg/sdg-flow-yaml.md
markmc: Ok, I hope commit 7f24fdc clarifies this

I would like an ack from @xukai92 on:

> We believe the most common case for custom pipelines is for them to extend the `full` pipeline

markmc: > We could make it that way. But the input to the custom parts is the output of “full” … so it’s this or we copy stuff or something else 🤷‍♂️

I don't think "copy stuff" is acceptable (I know you don't either) - every change to the full pipeline would need to be replicated to any custom pipeline

docs/sdg/sdg-flow-yaml.md
markmc: I think this is resolved by commit 4425f8d

docs/sdg/sdg-flow-yaml.md
markmc: Ok, I think this issue warrants a significant change to the design

I order to specific a pipeline, you need to specify a chain of blocks for knowledge, freeform_skills, and grounded_skills

I think we can see from the POC that keeping the chains for knowledge, freeform_skills, and grounded_skills in separate files aids clarity

So I think we need something like this

```
src/
  instructlab/
    sdg/
      pipelines/
        simple/
          knowledge.yaml
          freeform_skills.yaml
          grounded_skills.yaml
        full/
          knowledge.yaml   # also contains the current contents of mmlu_bench.yaml
          freeform_skills.yaml
          grounded_skills.yaml
```

and a custom pipeline would be a directory with these 3 YAML files

docs/sdg/sdg-flow-yaml.md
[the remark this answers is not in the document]
markmc: I'm [proposing below](https://github.com/instructlab/dev-docs/pull/109#discussion_r1673815390) to merge `MMLUBenchFlow` into `SynthKnowledgeFlow` so that a pipeline is a directory with `freeform_skills.yaml`, `grounded_skills.yaml`, and `knowledge.yaml`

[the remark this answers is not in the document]
xukai92: responded below

docs/sdg/sdg-flow-yaml.md
xukai92: this could be confusing what a pipeline is ... I would rather say there are actually 3 pipelines under simple/full and we route them differently in our particular frontend. this semantics should not be enforced

docs/sdg/sdg-flow-yaml.md
xukai92: sounds good!

docs/sdg/sdg-flow-yaml.md
markmc: Right, "full" or "simple" is a set of pipelines. I agree with that.

But likewise, when using custom pipelines, you should be providing a set of custom pipelines

russellb: We could possibly combine everything into a single yaml file instead of 3 separate ones in a directory?

There's related discussion over on this PR: https://github.com/instructlab/instructlab/pull/1668#pullrequestreview-2172270127

though the discussion should probably move back over here

docs/sdg/sdg-flow-yaml.md
russellb: thanks the additions in that commit resolve this thread for me

docs/sdg/sdg-flow-yaml.md
markmc: > We could possibly combine everything into a single yaml file instead of 3 separate ones in a directory?

Oh, absolutely. Except when you actually look at the files

Who do we serve by merging these 4 files? [synth_freeform_skills.yaml](https://github.com/instructlab/sdg/blob/5cc6cd1716ea40804f418ce7b8a9d556dc13a98e/src/instructlab/sdg/flows/synth_freeform_skills.yaml), [synth_grounded_skills.yaml](https://github.com/instructlab/sdg/blob/5cc6cd1716ea40804f418ce7b8a9d556dc13a98e/src/instructlab/sdg/flows/synth_grounded_skills.yaml), [mmlu_bench.yaml](https://github.com/instructlab/sdg/blob/5cc6cd1716ea40804f418ce7b8a9d556dc13a98e/src/instructlab/sdg/flows/mmlu_bench.yaml), [synth_knowledge.yaml](https://github.com/instructlab/sdg/blob/5cc6cd1716ea40804f418ce7b8a9d556dc13a98e/src/instructlab/sdg/flows/synth_knowledge.yaml)

I find them 10x easier to reason about as separate files. And I can only assume we will have more complex pipelines than these.

docs/sdg/sdg-flow-yaml.md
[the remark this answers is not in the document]
markmc: > can you point the code to me? i only see the code have a SDG that chains a list of pipelines, but not a pipeline that chains a list of flows.

I'm just saying we have a sequence of blocks in one file and a sequence of blocks in another file and we're chaining them together

docs/sdg/sdg-flow-yaml.md
markmc: Are we talking past each other? I'm referring to this: https://github.com/instructlab/sdg/blob/7bf156360f2c44b4c03570c1a67e0d63d6b69d53/src/instructlab/sdg/default_flows.py#L132-L134

docs/sdg/sdg-flow-yaml.md
russellb: > I would like an ack from @xukai92 on:
> 
> > We believe the most common case for custom pipelines is for them to extend the `full` pipeline

I asked @shivchander about this on Slack and got a confirmation. However, it's slightly more complicated than I thought!

In one case, we want to run custom configuration *after* the `full` pipeline (skills). In another, we want to run custom configuration *before* the `full` pipeline (knowledge). I missed this important ordering detail on the knowledge side.

We've got to factor this into the design somehow.

docs/sdg/sdg-flow-yaml.md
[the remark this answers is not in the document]
xukai92: But the current abstraction does so by chaining Pipeline in SDG but not multiple flows in Pipeline, right?
I'm simply suggesting to not change this behavior while adding YAML file support.

docs/sdg/sdg-flow-yaml.md
xukai92: oh, you mean those block templates. i thought you refers to prompt template as defined per model family.

resolving as it's  my misunderstanding

docs/sdg/sdg-flow-yaml.md
xukai92: i think it's a good idea to do that somehow.
do we name the pipelines in this case such that we can extract them when routing them to the corresponding parts? and where the routing logic is defined (maybe we already have some ideas in Slack)

docs/sdg/sdg-flow-yaml.md
xukai92: trying to reply to some comments from @markmc 

> I would like an ack from @xukai92 on:
> We believe the most common case for custom pipelines is for them to extend the full pipeline

I would say "reusing blocks" is the more expected common use cases, not extending the (full) pipeline.
Ideally the flexibility we want is to reuse the blocks as nodes in a DAG.
It happens to that the extension from 1.5 to 2.0 is like appending a list of blocks to another list of blocks (which looks like we are chaining two pipelines. preview: putting existing + some new blocks in a DAG with some loops is what we will work on for something like 3.0).
But I don't think we need to nail this generic design down here as it's not a GA block.

> We could make it that way. But the input to the custom parts is the output of “full” … so it’s this or we copy stuff or something else 🤷‍♂️
> I don't think "copy stuff" is acceptable (I know you don't either) - every change to the full pipeline would need to be replicated to any custom pipeline

Please see mine and Russel's comment above.

And my feeling to copying stuff here is pretty mixed as the reusability/customization at the pipeline level is sort of limited.

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,205 @@
...
+
+Use the built-in `full` pipeline.
+
+> `ilab data generate --pipeline full --pipeline path/to/config.yaml`
+
+Use the `full` pipeline, but extend it with the pipeline defined in the provided config file.
+
+> `ilab data generate --pipeline path/to/other_config.yaml`
+
+Do not use a built-in pipeline. Use only the pipeline in the custom configuration.
```
russellb, line 158: related to https://github.com/instructlab/dev-docs/pull/109#discussion_r1674537552

i think the relationships between existing and custom flows are too complicated to express in terms of command line args.  I think we're going to have to find a way to express it in the files themselves for custom flows. 

Also related to this discussion where our custom pipeline configuration is getting complex enough to need to be a directory of files: https://github.com/instructlab/dev-docs/pull/109/files#r1673815390

With these points, I think we can simplify the CLI integration back down to:

```suggestion
> `ilab data generate --pipeline full`

Use the built-in `full` pipeline.

> `ilab data generate --pipeline path/to/custom_pipeline_directory/`

Use a custom pipeline configuration. The custom pipeline may include references to the built-in flows to be used in conjunction with custom ones, but those details are contained within the yaml files in the custom directory.
```

docs/sdg/sdg-flow-yaml.md
```diff
@@ -0,0 +1,205 @@
...
+    drop_duplicates:
+      - question
+```
+
+and the user will be required to define a serving configuration like:
+
+```bash
+--lora-modules=mycustomadapter=path/to/my_custom_adapter
+```
+
```
russellb, line 135:
```suggestion

### Re-use of Built-in Pipelines

A custom pipeline may want to add to an existing built-in pipeline.  In that case, in the list of blocks, one may specify a special entry that indicates what to inherit from.

```yaml
version: "1.0"
blocks:
  - inherit_from: configs/full/freeform_skills.yaml
  - name: gen_questions
    type: LLMBlock
    ...
```

Note that an `inherit_from` entry can appear in any location in the `blocks` list (beginning, somewhere in the middle, end).

russellb: a more "pure" approach here could be a new block type, like "CallBlock" that acts like a function call of sorts, calling out to a separate pipeline definition and then returning back here .

... but that feels like overkill right now maybe?

docs/sdg/sdg-flow-yaml.md
markmc: > In one case, we want to run custom configuration _after_ the `full` pipeline (skills). In another, we want to run custom configuration _before_ the `full` pipeline (knowledge). I missed this important ordering detail on the knowledge side.

Awesome to get this clarified :+1:

markmc: >But I don't think we need to nail this generic design down here as it's not a GA block.

Great. In this kind of scenario, I'm looking out for ways our decisions now might prevent us adding support for more elaborate chaining like that in future. And it feels like some that can be added in a future release when we have a better understanding exactly what we need.

> And my feeling to copying stuff here is pretty mixed as the reusability/customization at the pipeline level is sort of limited.

"Copying stuff" is always an option of course. But if "chain two pipelines" together works for our first release, then supporting that will make the whole thing will be much more maintainable if we don't copy entire chains of blocks.

docs/sdg/sdg-flow-yaml.md
markmc: on Slack, we talked about add the idea of a `PipelineRouter` which I guess would have an interface like this:

```
class PipelineRouter:
    def __init__(self, ...);
        ...

    def generate(self, dataset) -> Dataset:
       ...
```

(hmm, that's the same interface as a Pipeline which suggests a base class that Pipeline and Router inherit from)

and there would be two variants:

```
class DirectRouter(Router):
    def __init__(self, pipeline_yaml):
        self.pipeline = load_pipeline(pipeline_yaml)

    def generate(dataset):
        return self.pipeline.generate(dataset)
```

and the use case for this would be if you are developing a pipeline against a known data set?

But the more common use case is covered by this variant that understands how our taxonomy repo is structured:

```
class TaxonomyRouter(Router):
    def __init__(self, freeform_skills_pipeline_yaml, grounded_skills_pipeline_yaml, knowledge_pipeline_yaml):
       self.freeform_skills_pipeline = load_pipeline(freeform_skills_pipeline_yaml)
       ....

   def generate(samples):
       if samples[0].get("document"):
         pipeline = self.knowledge_pipeline
      elif samples[0].get("seed_context"):
         pipeline = self.grounded_skills_pipeline
      else:
          pipeline = self.freeform_skills_pipeline
      return pipeline.generate(samples)
```

"full" and "simple" would be an alias for a set of yaml files

(the above is just a sketch, it would evolve quite a bit as you hack on it)

markmc: And the point of all of this is "full" isn't a pipeline, but instead its some logic that understands the taxonomy repo and looks at the data set before picking from one of 3 pipelines

docs/sdg/sdg-flow-yaml.md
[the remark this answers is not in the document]
markmc: The current API allows `SDG([pipeline1, pipeline2])` or `Pipeline(block_configs + block_configs)` and they are identical. I've been using the latter because "SDG" isn't a noun (it doesn't convey what it does) and because as [pep20 says](https://peps.python.org/pep-0020/) "There should be one-- and preferably only one --obvious way to do it."

This all becomes moot if we merge `MMLUBenchFlow` and `SynthKnowledgeFlow` into a `knowledge.yaml`

docs/sdg/sdg-flow-yaml.md
markmc: Sounds good to me :+1:

docs/sdg/sdg-flow-yaml.md
russellb: OK, I think this is making sense to me.

From the `ilab` CLI perspective, it only cares about the taxonomy variant. It can still just talk about a "pipeline" as a user concept, but the target is either:

* an alias that sets up a TaxonomyRouter with 3 built-in pipelines
* a path to a directory that has 3 pipeline yamls in it, and we initialize the TaxonomyRouter with that

altered syntax is over here: https://github.com/instructlab/dev-docs/pull/109/files#r1674565115

docs/sdg/sdg-flow-yaml.md
markmc: ChatGPT [suggested](https://chatgpt.com/share/7a6f323e-d190-4a19-b42c-c56d8403b93d) YAML Anchors, Aliases and Merge Keys ... but I don't think it's a great fit

How about "import" instead of "inherit"?

And a future expansion might look like:

```yaml
version: "1.0"
blocks:
  - import:
      - path: configs/full/freeform_skills.yaml
        blocks:
         - gen_questions
         - eval_questions
      - path: ...
  - name: gen_questions
    type: LLMBlock
    ...
`` `

(we would add the "import blocks" bit now, that would be a future enhancement potentially)

Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:135]: ok, so I can capture this as stages.  A shorter term ` - import_from: path/to/yaml` and then a longer term nicer expansion that's more expressive like you laid out the example.

I'll capture it like that in a doc update

File: docs/sdg/sdg-flow-yaml.md (hunk 26 above)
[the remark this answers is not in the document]
Comment (xukai92) [reply docs/sdg/sdg-flow-yaml.md:21]: The semantics between SDG or SyntheticDataGeneration and Pipeline is probably worth another discussion (I think you already have a PR on that). But here I'm only trying to say let's decouple that kind of changes from the YAML file support.
btw feel free to resolve this comment.

File: docs/sdg/sdg-flow-yaml.md (hunk 23 above)
Comment (xukai92) [reply docs/sdg/sdg-flow-yaml.md:146]: Yeah, the class definitions and methods themselves make sense to me as well.
Abstraction-wise, one thing I've been wondering though is whether routing is a thing that `Pipeline` should be responsible for, or it's a higher level concept should route pipelines (in our case, it could be the `SDG`).
Also as per Mark's word: routing sounds like "an object that has a set of pipelines, and knows which pipeline to use for different input datasets", which also doesn't sound like another pipeline but something conveys the details about the actual **synthetic data generation** process (a bunch of pipelines applied to a set of files from taxonomy, in our case).
PS: open to deal with this later as the design above already unblocks us I believe.

File: docs/sdg/sdg-flow-yaml.md (hunk 26 above)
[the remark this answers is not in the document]
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:21]: > This all becomes moot if we merge MMLUBenchFlow and SynthKnowledgeFlow into a knowledge.yaml

I'm interpreting this as we don't actually need support for a list of files under the current proposal of a single `knowledge.yaml`, so I can adjust this to be `from_file()` that loads a single yaml, since that's sufficient for the immediate requirements. We could always add an additional `from_files()` later if it comes up.

Let me know if I got that wrong.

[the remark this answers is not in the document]
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:21]: (or `from_flow()` .... or `from_yaml()` ... i typoed it, but the point being singular is good enough for the moment)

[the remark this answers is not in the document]
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md]: @xukai92 please take a look at the latest version and see if you're good with the changes to this section.

File: docs/sdg/sdg-flow-yaml.md (hunk 29 above)
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:152]: ah ha!  thank you.

Comment (markmc) [reply docs/sdg/sdg-flow-yaml.md:152]: Bleh, no - I think you were right - an import should be another type of block

```
version: 1.0
blocks:
  - name: import_1
     type: ImportBlock
     path: 1.yaml
```

plenty of room for expansion while still respecting that the blocks clause is a list of block objects

Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:152]: OK - yeah ... everything else is a hack and a special case, where this seems "right"

it's probably not any harder to implement ... i hope ...

Comment (xukai92) [reply docs/sdg/sdg-flow-yaml.md:152]: Looks good to me. The `ImportBlock` is a neat idea!

File: docs/sdg/sdg-flow-yaml.md (hunk 28 above)
Comment (xukai92) [reply docs/sdg/sdg-flow-yaml.md:175]: Nice. This 100% aligns what I think :D

File: docs/sdg/sdg-flow-yaml.md (hunk 29 above)
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:152]: I pushed an update to reflect the use of an `ImportBlock` to reference another pipeline configuration file.

File: docs/sdg/sdg-flow-yaml.md (hunk 28 above)
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:192]: I applied the proposed text here that simplified the CLI integration proposed

File: docs/sdg/sdg-flow-yaml.md (hunk 23 above)
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md:180]: Out of all of this, I think the really important details are:

* `ilab data generate --pipeline` can take a path to a directory. I updated the doc text to reflect that part already.

We expect the contents of that directory to have 3 specific files: `knowledge.yaml`, `grounded_skills.yaml`, and `freeform_skills.yaml`. I will add that in a moment.

There are also some really good points about how we can implement this in terms of abstractions. I am comfortable leaving that part out of the doc and see how it shakes out in the implementation. I have a feeling the final details will get sorted as we see what it looks like in implementation, but knowing that it's inspired by our discussion here. Let me know if anyone feels diffrently.

File: docs/sdg/sdg-flow-yaml.md (hunk 22 above)
Comment (russellb) [reply docs/sdg/sdg-flow-yaml.md]: @markmc had some good ideas on how we could expand `ImportBlock` to load specific blocks from another config instead of only allowing importing an entire config. It sounds like we have a pretty good path to add that in the near future if it makes sense.

**Review by xukai92 — APPROVED** [review]

**Comment by markmc** [comment]
I've made one small change since @xukai92's approval - changing from "flow params" to "pipeline context". I'm pretty sure  @xukai92 acked that idea already.

And given the existing approvals from @leseb, @hickeyma, and the collaboration between @russellb and me on this ... I think we're in good shape to merge.

I don't think it's useful to include all the iterations of this design in `git log`:

```
d367f7b (HEAD -> sdg-flow-yaml, markmc/sdg-flow-yaml) Change FlowParams to PipelineContext
d524534 Describe file and directory structure
6c1b7f2 Simplify CLI integration proposal
4871552 Propose new block type: ImportBlock
282d764 Rename from_flow() to from_file()
0ea86cb Change Pipeline.from_flows() to .from_flow()
1bf99b5 Pacify linter and spellchecker
7f24fdc Update the CLI UX for pipeline chaining
6457751 Update the model serving section with the latest thinking
4425f8d Explain why YAML was not initially included in the library API design
f7663a5 Update to account for actual Pipeline.from_flows() API
f0ebc8e Make spellcheck happy
c8a3626 Address the use case distinguishing between multiple models
e6e5ee1 Add --list-pipelines suggestion from Seb
0011e66 Replace num_iters with num_instructions_to_generate
0db6e40 Remove extra whitespace
6f9ddad Remove some extra verbosity from format
dc58502 Tweak title
45deb44 Propose adding a file format for custom Flows
```

Referring back to this PR is probably much more useful if you ever want to understand how the proposal evolved.

So I'm going to squash these commits and go ahead and merge.

PR #111, opened 2024-07-03
introduce additional_arguments in config.yaml

cdoern:
This dev-doc introduces `ilab-config-structure.md` a document meant to discuss how we deal with nested classes in the config. Additionally this document discusses implicit vs explicit arguments that have been previously added in different ways in key `ilab` commands.

`additional_arguments` allows ilab to implicitly support more complex flags from the libraries it calls without adding them directly to the config.

### DIFF
--- a/docs/cli/ilab-config-structure.md
+++ b/docs/cli/ilab-config-structure.md
@@ -1,0 +1,121 @@
+# Modifying the ilab config structure to be more streamlined
+
+Currently, the `config.yaml` is large with major expansions in training, evaluation, and serving. Expansions help the user understand what the various configuration options are for each of these commands. However, there is a fine line between a verbose config and a cluttered one.
+
+This document describes a new structure for various parts of the config.yaml, and begins to outline how there are different levels of expertise in `ilab` which should dictate which options are available by default.
+
+## `additional_arguments` as a field in training, serving and evaluation
+
+Taking a look at the current training config it looks like: 
+
+```yaml
+train:
+  torch_args:
+    nnodes: 1
+    node_rank: 0
+    nproc_per_node: 1
+    rdzv_endpoint: 127.0.0.1:12222
+    rdzv_id: 123
+  train_args:
+    chat_tmpl_path: /home/ec2-user/instructlab/venv/lib64/python3.11/site-packages/instructlab/training/chat_templates/ibm_generic_tmpl.py
+    ckpt_output_dir: checkpoints
+    data_output_dir: train-output
+    data_path: ./taxonomy_data
+    deepspeed_options:
+      cpu_offload_optimizer: true
+      cpu_offload_optimizer_pin_memory: false
+      cpu_offload_optimizer_ratio: 1
+      save_samples: null
+    effective_batch_size: 100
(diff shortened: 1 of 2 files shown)


leseb (changes requested):
I have mixed feelings about this proposal. The configuration file should expose all the available configuration options in the software. Unless we are embedding other software (like we do with vLLM), I don't think we should "hide" any options. If we do, how will users discover them?

Another approach would be to follow the common practice of including all configuration options in the config files but commenting them out, along with descriptions of what they are, what they do and their default value.

hickeyma (changes requested):
Thanks @cdoern for pushing this proposal.

If I am reading it correctly, you are trying to decrease the configuration properties to make the configuration file more manageable. If this is so, would it be helpful to make a table of config items per capability (generation, training, etc.) and then have columns like `Property Name`, `To Maintain`, `Reason`. That way the reader can understand the properties being maintained and why.

docs/cli/ilab-config-structure.md
```diff
@@ -0,0 +1,121 @@
...
+```yaml
+serve:
+  backend: 'vllm'
+  host_port: 127.0.0.1:8000
+  max_ctx_size: 5120
+  gpus: 2
+  llm_family: ''
+  model_path: models/merlinite-7b-lab-Q4_K_M.gguf
+  served_model_name: "merlinite"
+  additional_arguments: ["--block-size=16", "--dtype=fp8"...]
```
danmcp, line 116: How would this work if there are multiple commands callable from one config section?  Would we not need the additional_arguments name to indicate where in particular the additional args are being passed through to?

With serving, you could choose to use the same args field to pass values to vllm or llama_cpp or another serving technology.  But If there are two calls to make to vllm or llama_cpp, it becomes less clear.

cdoern:
@leseb 

> I don't think we should "hide" any options. If we do, how will users discover them?

We have had a few conversations in the past week about the ever increasing complexity of the libraries in instructlab. Its not realistic to expect every option added in the libraries to match 1:1, but ilab should also be able to support them all.

If we are operating under the assumption ilab is the main point of entry into the libraries, it should support all the complex options the libraries have to offer. However, having these all spelled out in the config is pretty bad UX. This would allow us to choose which ones the user would most commonly need to edit while also supporting the more complicated ones, just implicitly.

leseb:
> @leseb
> 
> > I don't think we should "hide" any options. If we do, how will users discover them?
> 
> We have had a few conversations in the past week about the ever increasing complexity of the libraries in instructlab. Its not realistic to expect every option added in the libraries to match 1:1 with cfg options and flags in instructlab, but ilab should also be able to support them all one way or another.
> 
> If we are operating under the assumption ilab is the main point of entry into the libraries, it should support all the complex options the libraries have to offer. However, having these all spelled out in the config is pretty bad UX. This would allow us to choose which ones the user would most commonly need to edit while also supporting the more complicated ones, just implicitly.

Ok, that sounds reasonable to me then, I missed that during my initial review. Thanks for the clarification.

nathan-weinberg:
@cdoern what is going on with this?

PR #114, opened 2024-07-10
Added docs.instructlab.ai dev-doc

jjasghar:
Added dev-doc to create the docs.instructlab.ai site.

/cc @kelbrown20

### DIFF
--- a/docs/docs-instructlab-website.md
+++ b/docs/docs-instructlab-website.md
@@ -1,0 +1,17 @@
+# `docs.instructlab.ai` website
+
+## Problem statement
+
+Reading the numerous `README.md` files for the InstructLab Project has become overwhelming and cumbersome.
+Understanding the branching structure of the different ways to implement and use InstructLab is not only
+confusing but frustrating to beginning users. Not having a centralized documentation site that is the
+first and default stop for our downstream users is doing us a disservice.
+
+## Proposed resolution
+
+1) Leveraging the static site ability of GitHub we spin up the <https://github.com/instructlab/docs> repository.
+and create a `CNAME` to point to <https://docs.instructlab.ai>.
+2) We migrate the different `README.md`s to this static site, (most likely [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/),
+and iterate organizing the documentation via PRs and useful flow of information.
+3) We leverage our TechWriters to have a single location to update official "downstream" documentation.
+4) We have the dedicated `README.md`s for each project, but stripped down to very specific niche things for the sub-project, or fast feedback development notes.


russellb (commented):
Lgtm

my only concern is that when splitting docs and code, it makes it harder to require docs along with doc-impacting code changes. 

I assume there is a group assuming ownership of this new repo? Who would that be? You could list the initial permissions in a new section. 

Back to my first point though - I would ask whoever is part of this group how they would like to handle this. Should we require an issue opened against the docs repo for doc impacting code changes? Should we require a doc PR be posted before the code can merge? What’s the minimum expectation as well as the “best case” request of people?

kelbrown20:
Thank you for creating this @jjasghar! This is also what i’ve had in mind for how the upstream documentation should look.

I was also thinking what @russellb was referring to on the best way to track doc changes and ensure the docs repo stays consistent with the code repos. I’m wondering if there is a way to track issues across repositories, where if someone changes code and creates a issue for docs, say in the InstructLab repo, you can also some how link that to the upstream doc repo. I’m not sure if thats a thing though

Or if theres a group that folks can @ when the documentation label is added and it creates an issue in the docs repo, or just pings the docs repo maintainers.
Those were a few of my ideas, WDYT?

jjasghar:
I'm guessing we use the "oh, this changed" in the repo and linked it via a PR/Issue in the `docs` repo we'll create. I think we'd have to be relatively disciplined around when things merge and ensure the docs are kept up to date. 

(One way we can do this is to make sure the "official" CHANGELOGs get hosted on the docs site, so when someone releases something, they are forced to make sure that the PRs/Changes are reflected in the downstream docs. Maybe? I dunno.

This is a difficult problem, and it's not just technical. We need to agree on something and enforce it.

For what it's worth, I think the idea of making sure when people need to add/remove things that would affect the "docs now" instead of going to the project README, they should come here and put the PR in linking that change, so when we see the change merged we update the docs with the updated change. One more step for the developer/owner of the change, but it'll give the downstream adopters (who we are ultimately doing this for) a better experience.

bjhargrave:
I am a big fan of putting docs in the `docs/` folder of a repo. Then a PR can include the code change and the corresponding doc change. Given that we have many repos, we can still do this but the build for the website will need to collect the docs files from the set of repos in order to build the complete documentation website.

Doing something like this will enable individual repos to have PRs which both change code and its docs together. Then we aggregate and build the doc website on some schedule.

leseb:
@bjhargrave, I support this proposal. Each repository would maintain its own `/docs` section for housing its documentation. When documentation is updated in a PR, the PR should be labeled "docs". The central repository responsible for building the docs should then have a workflow that monitors PR events with this label to build and publish the documentation.

A straightforward approach could be to periodically build the docs using a pull timer or similar mechanism. The downside is that the documentation won't be immediately synchronized, but this delay is likely acceptable.

Very low human intervention should be needed once this is in place.

kelbrown20:
@lhawthorn @cybette Hello! Can you also take a look at this dev-doc when you get the chance?

russellb:
before settling on the multi-repo approach, it would be cool if someone could prototype it and make sure we have a process that's going to work.

a plugin like this seems relevant?  but I haven't used it https://github.com/jdoiro3/mkdocs-multirepo-plugin

hickeyma (commented):
Thanks @jjasghar for pushing this proposal. It is something that needs organising.

The Helm community initially used docs from the repo which were built then transferred over to built and rendered for website. It was quite static and a bit laborious (makefile target was available alright) as it was usually done before cutting a release. 

It then changed to using a docs repository (https://github.com/helm/helm-www) which contained all content like blogs, docs etc. The website framework is also part of the repository. It enabled improved and more unified documentation for users. 

The key as pointed out from other people's comments is how to make sure documentation is modified when required by a contribution to the community. This is where it needs to be highlighted in the community of the contribution best practice of pushing doc changes when needed. This is where maintainers lead by example when contributing and when reviewing. It should also be stated in the contributing guides of the repos. This issue exists even when docs are in the repo.

markstur (commented):
Let's go!

One way or another, this needs to happen.  My main concern is how to get from some main branch w/ PRs that need docs to a release with the docs all updated. Versioning and submitting issues is pretty straightforward, but if the gap isn't addressed ASAP (ideally before the release is cut) it might never get caught up.

It's mostly a communication issue, but to execute we need some good practices around releases, issues, and an effective focus on making docs current.

jjasghar:
The repo has been created github.com/instructlab/docs.instructlab.ai, and until the dns-zone file is updated https://github.com/instructlab/dns-zones/pull/4 the url is: https://instructlab.github.io/docs.instructlab.ai/

I have also started a migration for the community documentation here too: https://github.com/instructlab/community/pull/361

jjasghar:
https://docs.instructlab.ai is live.

nathan-weinberg (approved):
Since this has been created should we merge this?

jjasghar:
I believe so :)

PR #117, opened 2024-07-17
Proposal for versioning the ilab config file

alimaredia:
Related to instructlab/instructlab/issues/1725


docs/schema/ilab-config-version.md
```diff
@@ -0,0 +1,27 @@
...
+
+The config file in `ilab` version `0.17` cannot be used in future versions of `ilab` due to new mandatory fields and changes to existing fields.
+
+As the configuration file grows, changes, and becomes more complex, versioning is necessary in order to support forwards and backwards compatibility.
+
+## Proposal
+
+The `ilab` configuration file has a new `version` field at the top level of the config.
+
+The value of `version` should start at `1.0`.
```
bjhargrave, line 15:
```suggestion
The value of `version` should start at `1`.
```

I would just use a simple integer unless you plan to handle semantic versioning. Like where the cli handles all 1.x variants but not 2.x.

hickeyma (commented):
Thanks for pushing the proposal @alimaredia.

Open question about versioning configuration. Should configuration be tied to a major release rather than versioning it? In other words, the configuration needs to be backwards compatible within a major release version but breaking changes can apply on the next major version.

In that case, we can change configuration that breaks backwards compatibility up until InstructLab 1.0.0 release.

nathan-weinberg:
@alimaredia what is going on with this?

PR #120, opened 2024-07-18
import  wiki documents to SDG

makelinux:


makelinux:
@russellb , @leseb would you like to review please?

Issue #124, opened 2024-08-01
Draft document detailing release procedures for new libraries 


nathan-weinberg:
We've recently introduced 3 new Python libraries in the InstructLab organization:
- https://github.com/instructlab/sdg
- https://github.com/instructlab/training
- https://github.com/instructlab/eval

Currently there is no release strategy for these libraries - all releases have been made by the Maintainer teams off of `main` at their own discretion - we need a **consistent, **org-wide**** release strategy for these libraries, similar to what we have currently for the CLI: https://github.com/instructlab/instructlab/blob/main/docs/release-strategy.md

One solution is to just use the same strategy for these libraries as we do for the CLI - we can discuss it here, but ultimately my plan is to draft an overarching document at an org-level while also introducing specific docs in the library repos as we have for the CLI.

PR #125, opened 2024-08-01
Add document detailing release strategy for libraries

nathan-weinberg:
Resolves https://github.com/instructlab/dev-docs/issues/124

### DIFF
--- a/docs/library-release-strategy.md
+++ b/docs/library-release-strategy.md
@@ -1,0 +1,36 @@
+# Library Release Strategy for InstructLab
+
+This document describes the overarching release strategy and policies for Python libraries in the InstructLab organization.
+
+## Background and Problem Statement
+
+The InstructLab organization features multiple code repositories that are tagged and released as Python libraries.
+Primarily this has been the [CLI repository](https://github.com/instructlab/instructlab) but has grown over time,
+including but not limited to:
+
+- [sdg](https://github.com/instructlab/sdg)
+- [training](https://github.com/instructlab/training)
+- [eval](https://github.com/instructlab/eval)
+- [schema](https://github.com/instructlab/schema)
+- [GPTDolomite](https://github.com/instructlab/GPTDolomite)
+- [instructlab-quantize](https://github.com/instructlab/instructlab-quantize)
+
+You can see the published versions of these libraries on PyPI [here](https://pypi.org/search/?q=instructlab).
+
+While these libraries are all owned and maintained by the InstructLab organization, only the CLI has an official
+[release strategy](https://github.com/instructlab/instructlab/blob/main/docs/release-strategy.md) documented. Other library
+releases have been handled directly by the Maintainer teams at their own discretion. The organization needs to have certain
+overarching principals around this topic, while still allowing for flexibility for each library on case-by-case basis.
+
+## Proposal
+
+By default, each existing and new library should have the following a `release-strategy.md` aligned with the CLI doc. This proposal recognizes that certain libraries may need flexibility
+on case-by-case basis - therefore, Maintainer teams are empowered to modify these documents as they see fit, so long as the
+following tenants remain consistent:
+
+1. Packages **must** be named `instructlab-<package-name>`
+1. Packages **must** follow the `X.Y.Z` numbering scheme (i.e. [semvar](https://semver.org/))
+1. Packages **must** have GitHub tagged releases named `vX.Y.Z`
+1. Packages **must** use release branches for Y-Streams of the form `release-vX.Y`
+1. Packages **must** maintain a `CHANGELOG.md`
+1. Maintainer teams **must** publicly communicate Y-Stream releases through official InstructLab channels such as Slack or Mailing Lists. Z-Stream release communication is up to Maintainer discretion.
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -63,6 +63,7 @@ GGUFs
 ggufs
 GiB
 Gmail
+GPTDolomite
 gpu
 Guang
 hacky
@@ -155,6 +156,7 @@ safetensors
 Salawu
 SDG
 sdg
+semvar
 sexualized
 SHA
 sharding


docs/library-release-strategy.md
```diff
@@ -0,0 +1,34 @@
...
+[release strategy](https://github.com/instructlab/instructlab/blob/main/docs/release-strategy.md) documented. Other library
+releases have been handled directly by the Maintainer teams at their own discretion. The organization needs to have certain
+overarching principals around this topic, while still allowing for flexibility for each library on case-by-case basis.
+
+## Proposal
+By default, each existing and new library should have the following a `release-strategy.md` aligned with the CLI doc. This proposal recognizes that certain libraries may need flexibility
+on case-by-case basis - therefore, Maintainer teams are empowered to modify these documents as they see fit, so long as the
+following tenants remain consistent:
+1. Packages **must** be named `instructlab-<package-name>`
+1. Packages **must** follow the `X.Y.Z` numbering scheme
```
dhellmann, line 32: Is this implicitly SemVer versioning? If not, what is it? If it is SemVer, can we say that explicitly?

dhellmann (commented):
We spent a lot of time describing release "models" for related projects in the OpenStack community. The [document describing those models](https://releases.openstack.org/reference/release_models.html) might be interesting reading for folks thinking about the same topic for InstructLab.

A few things to consider that aren't mentioned here:

1. It's harder than it seems to make the version numbers of the CLI and libraries line up exactly and makes things more confusing to users and rebuilders. I recommend not even trying.
2. Versioning schemes like SemVer can help signal when a new release has breaking changes, but it's still better to make releases backwards compatible as much as possible to ensure adoption of new releases.
3. It makes it easier to reason about release series if the release branch is created from a tagged commit. That can be a `.0` release or a release candidate release. 
4. For libraries, no one can easily consume release candidates because they aren't installed automatically, so over-using them will make things harder, not easier.

docs/library-release-strategy.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Proposal
+
+By default, each existing and new library should have the following a `release-strategy.md` aligned with the CLI doc. This proposal recognizes that certain libraries may need flexibility
+on case-by-case basis - therefore, Maintainer teams are empowered to modify these documents as they see fit, so long as the
+following tenants remain consistent:
+
+1. Packages **must** be named `instructlab-<package-name>`
+1. Packages **must** follow the `X.Y.Z` numbering scheme
+1. Packages **must** have GitHub tagged releases named `v.X.Y.Z`
```
markstur, line 33: There's no dot after the "v".

markstur: Seems like we should attempt to describe the vX.Y.ZaN or vX.Y.ZbN that instructlab/instructlab is using for alpha/beta but I only suggest that because it looks like that project is breaking this rule if it's just "Z".  Is it premature to formalize the alpha/beta/etc? exceptions.

docs/library-release-strategy.md
```diff
@@ -0,0 +1,36 @@
...
+1. Packages **must** use release branches for Y-Streams.
```
markstur, line 34: just a thought... I feel like these y-stream branches could possibly be "as needed" but it's less wishy-washy as stated so yeah

markstur (commented):
see comments

docs/library-release-strategy.md
nathan-weinberg: Done

docs/library-release-strategy.md
nathan-weinberg: Good catch @markstur! I want to wait for the prerelease stuff until https://github.com/instructlab/instructlab/pull/1758 lands

docs/library-release-strategy.md
nathan-weinberg: Yeah I worry if we do "as needed" it will end up being "not doing it until we are in crisis" :smile:

danmcp: Practically they are needed when the Y stream is consumed by another component (Ex: instructlab cli).  Not sure that's worth calling out vs stating as required.

docs/library-release-strategy.md
```diff
@@ -0,0 +1,36 @@
...
+By default, each existing and new library should have the following a `release-strategy.md` aligned with the CLI doc. This proposal recognizes that certain libraries may need flexibility
+on case-by-case basis - therefore, Maintainer teams are empowered to modify these documents as they see fit, so long as the
+following tenants remain consistent:
+
+1. Packages **must** be named `instructlab-<package-name>`
+1. Packages **must** follow the `X.Y.Z` numbering scheme (i.e. [semvar](https://semver.org/))
+1. Packages **must** have GitHub tagged releases named `vX.Y.Z`
+1. Packages **must** use release branches for Y-Streams.
+1. Packages **must** maintain a `CHANGELOG.md`
+1. Maintainer teams **must** publicly communicate releases through official InstructLab channels such as Slack or Mailing Lists
```
RobotSail, line 36: Does this need to happen for all releases or can it be a requirement for minor releases only and not patch releases?

nathan-weinberg: I think just for Y-Streams that's alright - I can add a note around this

cdoern: in our upstream release strategy md for `ilab`, this part was amended for z streams to only be if the maintainers deem necessary

RobotSail: Yeah that makes sense, I can see reasons for it, but probably having it be required will cause people to tune out the announcements.

RobotSail (approved):
This is all very reasonable and will help out with broader communication across different groups. LGTM.

docs/library-release-strategy.md
nathan-weinberg: Updated the text

danmcp (approved):

nathan-weinberg:
I am going to merge this given the two approvals and amount of time it has sat in the queue, if anyone wants to iterate on these policies feel free to open up a followup PR for discussion

PR #128, opened 2024-08-27
github: add stale bot to dev docs repo

nathan-weinberg:
Same action being used in other InstructLab repos

### DIFF
--- a/.github/workflows/stale_bot.yml
+++ b/.github/workflows/stale_bot.yml
@@ -1,0 +1,52 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: 'Close stale issues and PRs'
+
+on:
+  schedule:
+    - cron: '30 1 * * *'
+
+env:
+  LC_ALL: en_US.UTF-8
+
+defaults:
+  run:
+    shell: bash
+
+permissions:
+  contents: read
+
+jobs:
+  stale:
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Harden Runner"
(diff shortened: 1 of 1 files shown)


bjhargrave (approved):
You may want to con

.github/workflows/stale_bot.yml
```diff
@@ -0,0 +1,49 @@
...
+  stale:
+    permissions:
+      issues: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Harden Runner"
+        uses: step-security/harden-runner@5c7944e73c4c2a096b17a9cb74d65b6c2bbafbde # v2.9.1
+        with:
+          egress-policy: audit # TODO: change to 'egress-policy: block' after couple of runs
```
danmcp, line 29: Is this because we aren't sure if egress is needed yet?

nathan-weinberg: I believe so yes - @bjhargrave can you confirm?

bjhargrave, line 29:
```suggestion
          disable-sudo: true
          egress-policy: block
          allowed-endpoints: >
            api.github.com:443
```

For these stale workflows, we can configure the egress policy as above.

I should get around to making PRs to tighten this up for all the workflows in all the repos.

nathan-weinberg: Will make this change!

nathan-weinberg: Fixed in latest patchset

bjhargrave (approved):

danmcp (approved):

PR #129, opened 2024-09-03
CMB Release Process

jjasghar:
This is the overarching release cadence for the CMB.


nathan-weinberg:
@jjasghar can you fix the CI errors?

jjasghar:
I don't know if this is still technically accurate for what we are doing now. I think when we get the CMB back up and running and we start looking at granite-3.0 too, we might have a another path?

Granted there is no reason why we couldn't release the grante-7b-lab model  too? Should we discuss this in the triage meeting? /cc @juliadenham

PR #131, opened 2024-09-05
Introduce ilab model standards across CLI and libraries

alinaryan:
Introduce standardizing how models are stored, managed, and accessed via flags used in CLI tools and libraries to ensure consistency, usability, and maintainability.

### DIFF
--- a/docs/cli/ilab-model-management-standardization.md
+++ b/docs/cli/ilab-model-management-standardization.md
@@ -1,0 +1,81 @@
+# Standardizing model storage, management and referencing in the CLI and libraries
+
+## Introduction
+
+- **Purpose**: Standardizing how models are stored, managed and accessed via flags used in CLI tools and libraries to ensure consistency, usability, and maintainability.
+- **Scope**: Covers model directory structure in cache and all model-related flags used in the command-line interface (CLI) and libraries within the project.
+- **Audience**: Developers, maintainers, and contributors involved in the development and usage of the CLI tools and libraries.
+
+## Summary
+
+This proposal aims to establish a consistent practice for how models are managed within InstructLab. It does so in 2 parts - one proposing a consolidation of all existing model related flags, and the other proposing a change to how models are organized in the cache upon download, and thereby referenced within various operations
+
+## Motivation
+
+Currently there are 10+ instances across all major commands and subcommands that accept either `--model` , `--model-path`, `--model-name`, or `--model-dir`; all of which serve slightly different purposes, and might handle different use cases including local vs remote models, relative vs absolute paths etc. This leads to a significant amount of avoidable confusion among developers and users alike. Introducing some consolidation and consistency in this aspect would provide a big UX improvement.
+
+In addition to this, we currently support multiple sources for model downloads. There could be models that span multiple sources and also have different versions and tags that users may want to be able to work with. Currently, there is a lack of uniformity in how we organize models downloaded from Hugging Face vs OCI registries. We do not have a way to differentiate and maintain unique identities for models in cache, based on their sources. We also currently don't have the ability to support version management. Users are only able to work with one version of a given model at a time, as downloading a different version just overwrites their existing model.
+
+## Proposal
+
+### Part 1: Narrowing down to a singular model related flag
+
+#### Problem Statement - Section 1
+
+We want to streamline the number of model related flags available, while establishing consistency and predictability between their uses and handling 3 separate use-cases: absolute paths, relative paths and remote repository names.
+
+#### Suggested approach - Section 1
+
+- Retain `--model` and deprecate all other model related flags
+  - Alternatively, we could retain `--model-path` but passing in a repo name to a flag called "model-path" seems a little more awkward than passing in a path to a field just called "model"
+- `--model` should accept both paths, as well as strings (for repo names)
+- First `--model` should simply check whether the supplied path exists - This would include determining whether the provided path is absolute, or relative to wherever the calling program is.
+  - If it exists, we should run an additional check to determine if it points to a safetensor or gguf models (use existing `is_model_safetensors` and `is_model_gguf` checks for this)
+  - If path does not exist, or exists but is not determined to be a valid model, then move on
+- If path is not found after step 1, check against all tracked models/checkpoints/adapters etc currently found in `~/.local/cache` and or `~/.local/checkpoints`. If not, move on
+- If supplied content is neither an absolute path, nor a relative path - assume it is the name of a remote repo on HF and download it
+  - alternatively, we can error out here and require that user download the model explicitly via `ilab model download`
+
+This would standardize the behavior of the `--model` flag across all the commands that it appears in. There could be a dedicated model resolver function that implements the above described process.
+
+The only exception may be `ilab model download` which contains a `--model-dir` flag, which acts as a sink rather than a source. This flag could stand to benefit from being renamed to `--destination` instead.
+
+### Part 2: Standardizing what gets passed _into_ the model flag (value)
+
+#### Problem Statement - Section 2
+
+In addition to standardizing the flag itself, we must also standardize what gets passed INTO the flag, i.e the value passed to `--model`. The format used to reference models should work consistently regardless of whether the user is referencing
+a local model or a remote one. They should be able to use a consistent string to reference a given, specific model under all circumstances.
+
+#### Suggested approach - Section 2
+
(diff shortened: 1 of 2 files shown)


RobotSail:
I really like the changes proposed in this PR. It seems like we just need to change the functionality so we do:

1. If a model path matches against an OCI directory, first check if that exists
1. Check if the provided path matches against a file (doing `os.path.exists(model_path)` will cover both the absolute & relative path case)
1. If it didn't match, check against the output of `ilab model list` 

It also sounds like we need to extract the logic that's currently in `ilab model list` and make it globally available, so that way we can achieve consistent behavior across the project.

It also sounds like we'll need to do the same for the data we generate

docs/cli/ilab-model-path-standards.md
```diff
@@ -0,0 +1,40 @@
...
+Currently there are 10+ instances across all major commands and subcommands that accept either `--model` , `--model-path`, 
+`--model-name`, or `--model-dir` all of which serve slightly different purposes, and might handle different use cases including local vs remote models, relative vs absolute paths etc. This issue seeks to propose a solution that will standardize the use of model flags and make them behave consistently across all commands and sub-commands.
+
+## Problem Statement
+This aims to streamline the number of model related flags available, while establishing consistency and predictability between their uses and handling 3 separate use-cases: absolute paths, relative paths and remote repository names. 
+
+## Proposal
+- Retain `--model` and deprecate all other model related flags
+- `--model` should accept both paths, as well as str for repo names
+- First `--model` should assume whatever is passed into it is an absolute path, and should check against root to determine if that path exists
```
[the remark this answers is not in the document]
jaideepr97: 1ilab model list` will just display paths relative to `~/.local/cache/instructlab/models`, so what you're suggesting is essentially the same thing as what is proposed, with the exception of maybe supporting relative paths with respect to wherever the user is running the `ilab` command from (not relative to the default model cache dir) 

So we could replace the 1st point of the proposal with your 1st point and the rest is the same

docs/cli/ilab-model-management-standardization.md
```diff
@@ -0,0 +1,81 @@
...
+
+#### Suggested approach - Section 1
+
+- Retain `--model` and deprecate all other model related flags
+  - Alternatively, we could retain `--model-path` but passing in a repo name to a flag called "model-path" seems a little more awkward than passing in a path to a field just called "model"
+- `--model` should accept both paths, as well as strings (for repo names)
+- First `--model` should assume whatever is passed into it is an absolute path, and should check against root to determine if that path exists
+  - If it exists, we should run an additional check to determine if it points to a safetensor or gguf models (use existing `is_model_safetensors` and `is_model_gguf` checks for this)
+  - If path does not exist, or exists but is not determined to be a valid model, then move on
+- If path is not an absolute path, assume the path supplied is a relative path. Expand it by appending it to `~/.local/cache/instructlab/models` and check if the path exists. If exists and model is found, use it. If not, move on
```
JamesKunstle, line 35: You can expand relative paths in a nice way with "pathlib.Path.resolve()" which will replace heuristics like ~ with the valid replacements.

docs/cli/ilab-model-management-standardization.md
```diff
@@ -0,0 +1,81 @@
...
+- If supplied content is neither an absolute path, nor a relative path - assume it is the name of a remote repo on HF and download it
+  - alternatively, we can error out here and require that user download the model explicitly via `ilab model download`
+
+This would standardize the behavior of the `--model` flag across all the commands that it appears in. There could be a dedicated model resolver function that implements the above described process.
```
JamesKunstle, line 39: One feature to consider is what to do about model versions. It might be convenient to check if a model exists on hf with the following snippet:

https://github.com/huggingface/huggingface_hub/issues/36#issuecomment-1619942423

docs/cli/ilab-model-management-standardization.md
```diff
@@ -0,0 +1,81 @@
...
+
+The only exception may be `ilab model download` which contains a `--model-dir` flag, which acts as a sink rather than a source. This flag could stand to benefit from being renamed to `--destination` instead.
```
JamesKunstle, line 41: I think that's a good idea. Could be aliased to `--dest`

docs/cli/ilab-model-management-standardization.md
jaideepr97: @JamesKunstle versioning is discussed further down in the doc

docs/cli/ilab-model-management-standardization.md
```diff
@@ -0,0 +1,81 @@
...
+We want to streamline the number of model related flags available, while establishing consistency and predictability between their uses and handling 3 separate use-cases: absolute paths, relative paths and remote repository names.
+
+#### Suggested approach - Section 1
+
+- Retain `--model` and deprecate all other model related flags
+  - Alternatively, we could retain `--model-path` but passing in a repo name to a flag called "model-path" seems a little more awkward than passing in a path to a field just called "model"
+- `--model` should accept both paths, as well as strings (for repo names)
+- First `--model` should assume whatever is passed into it is an absolute path, and should check against root to determine if that path exists
+  - If it exists, we should run an additional check to determine if it points to a safetensor or gguf models (use existing `is_model_safetensors` and `is_model_gguf` checks for this)
+  - If path does not exist, or exists but is not determined to be a valid model, then move on
```
osilkin98, line 34: It doesn't matter if this is a relative or absolute path. You can condense the above 2 points to just check if the given path exists. Both cases will be handled with the exact same logic.

RobotSail, line 34: It doesn't matter if this is a relative or absolute path. You can condense the above 2 points to just check if the given path exists. Both cases will be handled with the exact same logic.

jaideepr97: but by default, wouldn't relative path be checked relative to where the program is running from, rather than relative to `~/.local/cache/instructlab/models`?

docs/cli/ilab-model-management-standardization.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Part 1: Narrowing down to a singular model related flag
+
+#### Problem Statement - Section 1
+
+We want to streamline the number of model related flags available, while establishing consistency and predictability between their uses and handling 3 separate use-cases: absolute paths, relative paths and remote repository names.
+
+#### Suggested approach - Section 1
+
+- Retain `--model` and deprecate all other model related flags
```
cdoern, line 29: is there a deprecation plan here or by deprecate do you mean remove?

docs/cli/ilab-model-management-standardization.md
cdoern, line 41: so, alot of these renames might need to wait for a major version bump to maintain backwards compatibility

cdoern (commented):
initial pass i'd like to discuss before continuing. Thanks for putting this up!

docs/cli/ilab-model-management-standardization.md
RobotSail: @jaideepr97 the correct way to handle this would be:

1. Check if the file at the given path exists, return if true
2. Get the contents of model list
3. If the given path matches any of the listed, return here
4. Else error it's not found

Possible input cases:
A. Given path is relative to calling program
B. Given path is an absolute path
C. Given path is a result from the listed
D. Given path doesn't exist 


In case A it will return on step 1
In case B it will also return on step 1
In case C it will return on step 3
In case D it will hit step 4

jaideepr97: workflow is updated to be:
1. Check if the supplied path is either absolute, or relative to where the program is running
2. check if supplied path is relative to default cache dir
3. if neither - error out or download model from somewhere (TBD)

docs/cli/ilab-model-management-standardization.md
jaideepr97: we put out deprecation strings for all of them similar to what we've done for the aliases and mention this in the changelog

docs/cli/ilab-model-management-standardization.md
jaideepr97: why? I think they can all follow the same timeline 
introduce a `--destination` right now and add the deprecation notice to `--model-dirs` and get rid of all the deprecated flags in the next 1-2 releases

jaideepr97:
what do we need to get this dev doc merged? its been sitting for a while

RobotSail (approved):
LGTM

cdoern (approved):
thanks for the hard work on this

nathan-weinberg (approved):

PR #136, opened 2024-09-19
Add a dev doc for resuming SDG runs

bbrowning:
Document the design for how we plan to resume interrupted SDG runs.

### DIFF
--- a/docs/sdg/sdg-resume.md
+++ b/docs/sdg/sdg-resume.md
@@ -1,0 +1,96 @@
+# Resuming Interrupted Synthetic Data Generation Runs
+
+|                |            |
+| -------------- | ---------- |
+| Date           | 2024-09-19 |
+| Scope          | Synthetic Data Generation |
+| Status         | Draft |
+| Authors        | [Ben Browning](@bbrowning) |
+
+## What
+
+Synthetic Data Generation (SDG) should track its progress and allow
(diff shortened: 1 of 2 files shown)


docs/sdg/sdg-resume.md
```diff
@@ -0,0 +1,146 @@
...
+The following types of interruption should be handled:
+
+* user-triggered interruptions (e.g. CTRL-c)
+* SIGINT interruptions (e.g. when running on a spot instance of Pod
+  which receives SIGINT)
+* system hardware failure (GPU failure, power loss)
+
+## Non-Goals
+
+This is not replacing checkpoints. Checkpoints are how we save each
```
danmcp, line 51: It might be helpful to give the context of how long a batch might take for comparison.

bbrowning: Yes, that's a good idea. And I also need to talk about how checkpoints persist across individual data generation runs, while this is focused on the ability to resume a specific interrupted SDG run.

bbrowning:
This is not abandoned, but how we resume SDG runs is going to change as we integrate the context-aware chunking and docling. I was planning to revisit this once we get the former integrated. Just updating status here for visibility and for the stale bot.

bbrowning:
I'm going to close this myself, as SDG has changed so much since this started it will need a complete rethinking.

PR #106, opened 2024-06-25
InstructLab and Deepsearch

jjasghar:
This is the proposal to start integrating th document conversion system Deepsearch from IBM Research and InstructLab

PR #148, opened 2024-10-25
Knowledge doc ingestion

aakankshaduggal:
Related to https://github.com/instructlab/dev-docs/issues/149 https://github.com/instructlab/sdg/issues/324

### DIFF
--- a/docs/sdg/knowledge-doc-ingestion.md
+++ b/docs/sdg/knowledge-doc-ingestion.md
@@ -1,0 +1,96 @@
+# Knowledge Document Ingestion Pipeline Design Proposal
+
+**Author**: Aakanksha Duggal
+
+## 1. Introduction
+
+As part of extending InstructLab's capabilities, this pipeline is designed to support ingestion and processing of various document formats such as Markdown, PDF, DOCX, and more. The goal is to create a unified, modular system that seamlessly integrates with **Synthetic Data Generation (SDG)** and **Retrieval-Augmented Generation (RAG)** workflows, simplifying the process for users while maintaining flexibility for future enhancements.
+
+## 2. Use Case
+
+To enable the ingestion and processing of a wide range of document types, this pipeline must handle formats including:
+
+- Markdown (MD)
+- PDF
+- TXT
+- ASCII Docs
+- DOCX
+- HTML
+- PPTX
+
+This proposal outlines how to build a pluggable system that accommodates these formats and integrates effectively into existing **SDG** and **RAG** workflows.
+
+## 3. Proposed Approach
+
+### 3.1 Custom InstructLab Schema Design
+
+- **Objective**: Define a custom "instructlab" schema that standardizes input formats for SDG and RAG pipelines. This schema will bridge the gap between various document types and the specific formats required for further processing.
+- **Modularity**: The system should allow easy extension, enabling support for new document types and processing workflows without disrupting core functionality.
+  
+The instructlab schema will serve as the intermediary format, supporting flexibility while ensuring compatibility with the ingestion process.
+
+### 3.2 PDF and Document Conversion via Docling
+
+- **Docling Integration**: We will leverage **Docling** to convert files into structured JSON, which will be the starting point for the instructlab schema. Individual components will post-process this JSON as per the requirements of the specific SDG and RAG workflows.
+  
+- **Docling v2**: We have collaborated with the Docling team to extend their tool’s capabilities, allowing conversion for PDF, HTML and DOCX formats. These new document types will be supported in the ingestion pipeline via docling's upgraded v2 release.
+
+### 3.3 Introducing the Document Chunking Command
+
+- **Command Overview**: We propose a new command, `ilab document format`, which will:
+  - Take a document path (defined in `qna.yaml`).
+  - Format and chunk the document into the desired schema.
+  
+- **Implementation Details**: Initially, this functionality will be integrated into the existing SDG repository. Over time, it can evolve into a standalone utility, allowing external integrations and wider usage.
+
+- **Example Workflow**:
+  ```bash
+  ilab document format --input path/to/document.pdf --output path/to/schema
+  ```
+  This command will ingest a document, process it into the instructlab schema, and output the result for further use in SDG or RAG workflows.
+
+### 3.4 Simplifying Git-Based Workflows for Users
+
+- **Current Challenge**: Knowledge documents are stored in Git-based repositories, which may be unfamiliar to many users.
+- **Proposed Solution**:
+  - Allow users to input a local directory and provide an automated script that:
+    1. Initializes a Git repository.
+    2. Creates the necessary branches.
+    3. Organizes files into the required structure.
+  
+  By abstracting the Git setup process, we can retain Git’s benefits (version control, backups) while simplifying the interface for non-technical users.
+
+- **Implementation Example**:
+  ```bash
+  ./setup_git_repo.sh --input /path/to/local/docs
+  ```
+  This script automates the process of structuring knowledge documents for ingestion.
+
+### 3.5 Workflow Visualization
+
+Here is a conceptual diagram illustrating the workflow from document ingestion to schema conversion and chunking:
+
+![Knowledge_Document_Ingestion_Workflow](https://github.com/user-attachments/assets/06504b1b-bc8f-4909-b6a2-732a056613c5)
+
+The pipeline begins with the ingestion of knowledge document, passes through a conversion step to instructlab schema using **Docling**, and then processes the document into a format usable by SDG and RAG workflows.
+
+## 4. Future Enhancements
+
+### 4.1 Support for Additional Data Types
+
+To extend beyond text-based documents, the pipeline will explore handling other formats, including:
+- **Audio and Video**: Incorporating media formats will require modifications to the schema and additional processing capabilities.
+- **Visual Language Models (VLMs)**: We will collaborate with the research team to align this work with visual data processing tools and extend the pipeline to handle multimedia.
+
+### 4.2 Refined Chunker Library
+
+- **Standalone Library**: The document chunking functionality will eventually be refactored into a dedicated chunking library or utility within the SDG module. This separation will make it easier to maintain and extend in future iterations.
+  
+- **Performance Optimizations**: Ongoing work will aim to reduce the time and resources needed for large-scale document chunking, particularly for multi-format documents like PDFs containing both text and images.
+
+## 5. InstructLab Schema Overview
+
+### Key Components:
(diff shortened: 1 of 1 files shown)


docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+
+### 3.2 PDF and Document Conversion via Docling
+
+- **Docling Integration**: We will leverage **Docling** to convert files into structured JSON, which will be the starting point for the instructlab schema. Individual components will post-process this JSON as per the requirements of the specific SDG and RAG workflows.
+  
+- **Docling v2**: We have collaborated with the Docling team to extend their tool’s capabilities, allowing conversion for PDF, HTML and DOCX formats. These new document types will be supported in the ingestion pipeline via docling's upgraded v2 release.
+
+### 3.3 Introducing the Document Chunking Command
+
+- **Command Overview**: We propose a new command, `ilab document format`, which will:
```
williamcaban, line 40: Is there a particular reason for using the word `format`? The first time I read the command without reading the whole document, it gave me the impression the command was for transforming a document from one format to another format (e.g. pdf to md, or pdf to json). Should we consider a word or verb that more closely resembles what is actually happening in this step? For example:

`ilab docs import --input path/to/document.pdf --output path/to/schema`
`ilab docs ingest --input path/to/document.pdf --output path/to/schema`
`ilab docs process --input path/to/document.pdf --output path/to/schema`

docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+### 4.2 Refined Chunker Library
+
+- **Standalone Library**: The document chunking functionality will eventually be refactored into a dedicated chunking library or utility within the SDG module. This separation will make it easier to maintain and extend in future iterations.
+  
+- **Performance Optimizations**: Ongoing work will aim to reduce the time and resources needed for large-scale document chunking, particularly for multi-format documents like PDFs containing both text and images.
+
+## 5. InstructLab Schema Overview
+
+### Key Components:
+- **Docling JSON Output**: The output from Docling will be the instructlab schema, which serves as the backbone for both SDG and RAG workflows. For specific details around the leaf node path or timestamp, we will include that as a part of the file nomenclature.
```
williamcaban, line 94: As part of the ingestion command, should we consider a flag where the pipeline could augment the metadata of the final output like `--metadata ./path-to-metadata.json` to add information such as attribution, timestamps, ilab version, schema version, etc.?

nathan-weinberg:
cc @juliadenham

docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+- ASCII Docs
+- DOCX
+- HTML
+- PPTX
+
+This proposal outlines how to build a pluggable system that accommodates these formats and integrates effectively into existing **SDG** and **RAG** workflows.
+
+## 3. Proposed Approach
+
+### 3.1 Custom InstructLab Schema Design
```
nathan-weinberg, line 25: It is worth pointing out we have an existing `schema` package maintained by the @instructlab/schema-maintainers: https://github.com/instructlab/schema

Right now this is only for Taxonomy schema, but we could extend this with Classes designed specifically for this usecase

docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+
+- **Objective**: Define a custom "instructlab" schema that standardizes input formats for SDG and RAG pipelines. This schema will bridge the gap between various document types and the specific formats required for further processing.
+- **Modularity**: The system should allow easy extension, enabling support for new document types and processing workflows without disrupting core functionality.
+  
+The instructlab schema will serve as the intermediary format, supporting flexibility while ensuring compatibility with the ingestion process.
+
+### 3.2 PDF and Document Conversion via Docling
+
+- **Docling Integration**: We will leverage **Docling** to convert files into structured JSON, which will be the starting point for the instructlab schema. Individual components will post-process this JSON as per the requirements of the specific SDG and RAG workflows.
```
nathan-weinberg, line 34: Could we include a link here to somewhere where folks can read more about Docling, and perhaps a bit as to why Docling is the chosen solution here?

docs/sdg/knowledge-doc-ingestion.md
nathan-weinberg: I propose `ilab data` being used instead of `ilab docs` or `ilab document` - we already have this command group implemented in the CLI, and it would be good if users can keep their data manipulation in a single command group. I'd like some folks from UX to weigh-in here as well.

docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+  
+- **Docling v2**: We have collaborated with the Docling team to extend their tool’s capabilities, allowing conversion for PDF, HTML and DOCX formats. These new document types will be supported in the ingestion pipeline via docling's upgraded v2 release.
+
+### 3.3 Introducing the Document Chunking Command
+
+- **Command Overview**: We propose a new command, `ilab document format`, which will:
+  - Take a document path (defined in `qna.yaml`).
+  - Format and chunk the document into the desired schema.
+  
+- **Implementation Details**: Initially, this functionality will be integrated into the existing SDG repository. Over time, it can evolve into a standalone utility, allowing external integrations and wider usage.
```
nathan-weinberg, line 44: What would be the motivation for moving this out of the SDG repository? "allowing external integrations and wider usage" doesn't really tell me much

docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+  ```bash
+  ilab document format --input path/to/document.pdf --output path/to/schema
+  ```
+  This command will ingest a document, process it into the instructlab schema, and output the result for further use in SDG or RAG workflows.
+
+### 3.4 Simplifying Git-Based Workflows for Users
+
+- **Current Challenge**: Knowledge documents are stored in Git-based repositories, which may be unfamiliar to many users.
+- **Proposed Solution**:
+  - Allow users to input a local directory and provide an automated script that:
```
nathan-weinberg, line 56: Rather than a "script," why not just have this be part of the code? We can detect if a given directory is git-tracked (e.g. by checking for a `.git` subdirectory) and do the manipulation described if not

docs/sdg/knowledge-doc-ingestion.md
```diff
@@ -0,0 +1,96 @@
...
+  ```bash
+  ./setup_git_repo.sh --input /path/to/local/docs
+  ```
+  This script automates the process of structuring knowledge documents for ingestion.
+
+### 3.5 Workflow Visualization
+
+Here is a conceptual diagram illustrating the workflow from document ingestion to schema conversion and chunking:
+
+![Knowledge_Document_Ingestion_Workflow](https://github.com/user-attachments/assets/06504b1b-bc8f-4909-b6a2-732a056613c5)
```
nathan-weinberg, line 73: Very nice diagram!

nathan-weinberg (commented):
This is great @aakankshaduggal thank you for writing it up!

I would like to see this shared in `#sdg` in upstream Slack and maybe also sent out to `dev@instructlab.ai` so community members can weigh-in as well

nathan-weinberg:
Is this related? https://github.com/instructlab/dev-docs/pull/120 cc @makelinux

nathan-weinberg:
This one from @jjasghar also seems related? https://github.com/instructlab/dev-docs/pull/106

nathan-weinberg:
One more https://github.com/instructlab/dev-docs/pull/64

docs/sdg/knowledge-doc-ingestion.md
JustinXHale: @williamcaban I agree with your sentiments about `format` its very ambiguous. I think you are spot on with your suggestions. Here are some suggestions based on each prompt to provide some clarity for the user. 

`ingest` Provide the document path to ingest and process into the desired scheme.
`process` Provide the document path to process into the scheme.
`import`Specifiy the document path to import and format according to the scheme requirements.
`chunk` Enter the document path to split and structure for the scheme.

JustinXHale: If we think the document-related commands are going to expand considerably, then this might be the time to create a new command group. If the document processing will remain a small subset, then integrating them under the data group would simplify the CLI structure. 

`ingest` Provide the document path to ingest and process into the desired scheme.
`process` Provide the document path to process into the scheme.
`import` Specify the document path to import and format according to the scheme requirements.
`chunk` Enter the document path to split and structure for the scheme.

JustinXHale: If think the **document-related commands** are going to expand considerably, then this might be the time to create a new command group. If the document processing will remain **a small subset**, then **integrating** them under the `data` group would simplify the CLI structure. 

`ingest` Provide the document path to ingest and process into the desired scheme.
`process` Provide the document path to process into the scheme.
`import` Specify the document path to import and format according to the scheme requirements.
`chunk` Enter the document path to split and structure for the scheme.

relyt0925:
@aakankshaduggal (trying to envision overall flow in my head): I almost view this as proposing two independent yet related enhancements. One is the ability to define references to "documents" in a variety of ways versus just through git references. The other is actually talking about new document formats and how they would be injested.

so do you envision a user will still declaratively define "pointers" in their taxonomy to the backing doc storage similar to what is done today in knowledge like the following example:

```
document:
  repo: https://github.com/relyt0925/rbc-knowledge
  commit: 99dae176de4927940aee4faaeb0f645b3ee4582b
  patterns:
    - pdf_chunk*.md
```

However this "declarative definition" is now more flexible in the sense that it no longer has to necessarily just be
`repo, commit, pattern` It could be something like `filepath` within the base of a taxonomy which could look something like this

```
document:
  local_directory: documents/docchunks/ 
  patterns:
    - pdf_chunk0.md
```

Which would then in ilab data generate when I am processing the leaf node lead to the sdg process looking in a local path relative to the "taxonomy base" path for the documents to use in sdg?

relyt0925:
Then 2: the document type enhancement

First question: would it also be accurate to say that as we add in new document types (independent of the ways we reference them): we are still going to keep the declarative nature of the taxonomy where a user will explicitly reference the document in the taxonomy section. SDG then will handle when looking at the document determining it's type and then if it needs to be processed by docling and chunked. It will then produce the chunks (in the example of a 3 MB PDF file about 250 md chunks are produced): and handle ensuring those are processed as the "set" of documents for sdg? This would continue if multiple pdf files were defined? 

I am curious if you are envisioning things remaining in that flow versus what I would call a "pre processing" flow where users have to expilictly use the tooling to get the pdf docs as a pre req step, then create a knowledge repo (that would always only contain markdown documents), and then create a leaf node that points to the markdown documents only. Does that make sense the difference at a high level on what I am talking on?

PR #151, opened 2024-10-29
stalebot: reduce amount of time for PR to turn stale

nathan-weinberg:
We have lots of Dev Docs that get abandoned by their authors. This leads to confusion in the org as to what is actual policy and what is proposed. If a WIP Dev Doc is inactive, it should be closed sooner rather than later.

I have left issues alone but fixed a typo in the text.


markstur (approved):
lgtm

PR #155, opened 2024-11-14
Refactor preprocessing and postprocessing in SDG

jwm4:
This is a draft proposal that describes pros and cons of various options for refactoring preprocessing and postprocessing in SDG.  It includes draft decisions, but the purpose of the draft is not to assert that these are the final decisions but rather to act as a vehicle for reaching final decisions.  So comments are strongly encouraged, and readers should not expect that the decisions in the current draft will be unchanged between now and the time when this document is merged.


docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+- As with Option 1, this approach suffers from the fact that keeping interconnected components in the same repository provides less pressure to consistently document the API contracts between them.  In the case of Option 1, the interconnected components that would not have as much pressure to be documented would be preprocessing/postprocessing and core SDG.  In the case of Option 3, the interconnected components that would not have as much pressure to be documented would be the CLI and preprocessing/postprocessing.  However, in both cases, this con could be alleviated by just having the discipline to document the APIs well even without such pressure.
+- As with Option 2, this approach would not enable user flow 2.1.  Maybe that's fine since it is not on our requirements list.
```
cdoern, line 114: I think if we separate `prep`, `generate`, and `process` into different commands, a user could run `prep`+`generate` from the CLI if all of the code lives there and is separate from the core SDG code

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+
+Con:
+
+- Avoids the pros of both Option 1 and Option 2.
+- As with Option 1, this approach involves a lot of coordination.  There are a lot of stakehoders involved in the CLI and locating preprocessing and postprocessing there drags those stakeholders into issues relating to preprocessing and postprocessing.  However, as with Option 1, coordinating across stakeholders is something an open source project needs to do well anyway to remain engaged with the community. 
+- As with Option 1, this approach suffers from the fact that keeping interconnected components in the same repository provides less pressure to consistently document the API contracts between them.  In the case of Option 1, the interconnected components that would not have as much pressure to be documented would be preprocessing/postprocessing and core SDG.  In the case of Option 3, the interconnected components that would not have as much pressure to be documented would be the CLI and preprocessing/postprocessing.  However, in both cases, this con could be alleviated by just having the discipline to document the APIs well even without such pressure.
```
cdoern, line 113: I believe that this option forces more documentation and contract maintenance than option 1.

Having the CLI own pre and post processing will enforce the SDG loop to remain compatible with the user contracts set by the `prep`, `generate`, and `process` commands. If all code lives in `instructlab-sdg` we risk siloing and only focusing on the "modular" usecases that team commonly deals with for the product.

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+
+Conclusion:
+
+- The cost of having a separate reposity is so high that we would only consider this option as a last resort.
+
+### Option 3: Move preprocessing and postprocessing into the CLI repository
+
+Pro:
+
+- The CLI already has a lot of "supporting" (non-core) functionality, so it would respect established precedent to include preprocessing and/or postprocessing here.
```
cdoern, line 104: I think a key `pro` left out here is the amount of functionality in the CLI:

1. download
2. serve
3. chat
4. train (community)
5. list
6. edit
7. init

it contains most user facing logic aside from what we call the "core" parts of the workflow (SDG, Train, Eval). Moving taxonomy/doc processing to the CLI fits with the design intention of the 3 libraries when they were created.

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+- The ezxpected RAG functionality in 2025 will have some complex interactions with both preprocessing and postprocessing, perhaps even involving user flows in which the core SDG functionality is not needed.  In that case, it would be confusing to have the code path for RAG include a call out to the SDG library for doing preprocessing but not actually doing the core SDG.
+- It would just be simpler to explain to all stakeholders if the functionality that I've been calling "core SDG" was really just called "SDG".  We can't do that now because the SDG library has preprocessing and postprocessing in it too.
+
+Conclusion:
+
+- While the cons here are substantial, so are the pros.  None of the cons really seem disqualifying.  The first pro (future changes to the formats can be more self-contained) seems particularly compelling because this is a rapidly evolving field and adding new expressive power seems like something we will want frequently.
```
cdoern, line 78: I really think the risk of siloing and breaking user contracts is one to be wary of. If the plan is SDG to be run independently, then there _will_ be tooling built around the sdg repo. The more tooling built around just running SDG independently the more risk of breaking upstream contracts. The more functionality living in SDG that isn't SDG, the more surface area there is to break.

The CLI repo is dedicated to a good UX and both upstream+downstream experience of executing ilab. I feel like we would welcome these new features as part of that suite while they may be a burden to maintain in the SDG repo over time

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Question 2: What should the commands be in the CLI?
+
+One way to support both 1.3.1 and 1.3.2 would be to have separate CLI commands for the preprocessing, core SDG, and postprocessing step .  Alternatively, a single CLI command that does all of these and also saves the outputs of preprocessing to disk would support 1.3.1 but _not_ 1.3.2.  Even if we only want to support 1.3.1, having separate CLI commands for each step might be desirable because it is just more intuitive that if a user wants to save the outputs of preprocessing to disk to have a command to do that instead of having it be a "side effect" of an omnibus SDG command.  Here is a rough outline of what the separate commands would be:
+
+- `ilab data prep` would handle all the preprocessing (the first three bullets in the Context section above, plus any additional preprocessing we add in the future).
```
cdoern, line 47: As we discussed, I like the split of these commands, the modularity makes sense to me!

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+
+Conclusion:
+
+- This seems like a reasonable option.  The cons are mostly manageable.  However, overall the pros of Option 1 seem more compelling.
+
+### Option 4: Preprocessing and postprocessing go to different locations
+
+We could also mix and match any of the above options separately for preprocessing and postprocessing.  For example, preprocessing could move to the CLI repo and postprocessing could stay in the SDG repo.  Or preprocessing could move to a new repository and postprocessing could move to a different new repository or the _same_ new repository.  Enumerating all possible permutations of where each could go and enumerating pros and cons of each of them would make this document unbearably long.  If anyone wants to advocate for a small number of specific permutations, we will add them to this document.
+
+## Question 4: Should preprocessing, postprocessing, and core SDG be separate Python packages?
```
cdoern, line 124: separate packages introduces 2 more version bumps per ilab release. As it is, we barely have enough time to test+validate each package with the upstream CLI, I think keeping that as code modules but not python packages makes the most sense

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+
+If we choose Option 1 (leave preprocessing and postprocessing in SDG) then we still have the option to separate them into distinct Python packages.  That would get us some of the benefits of Option 2 (moving them to a different repository) while avoiding *some* of the costs of Option 2.  In particular, it would make the boundaries clearer and put more pressure on the developers of preprocessing, postprocessing, and core SDG to have well documented contracts for how each of these elements interact.  With that said, it would also bring in some additional complexity and increase the amount of work involved.
+
+## Decisions
+
+Since this is a draft, no decisions are made yet.  However, here are the current draft decisions:
+
+- We will support the following user flows: 1.1., 1.2., 1.3, 1.3.1, 1.3.2, 2.1, and 2.1 as documented in the Question 1 section above.
+- We will adopt the updates to the CLI that will be documented in Question 2 above.
+- We will leave the preprocessing in SDG as described in Question 3: Option 1.
```
cdoern, line 134: Given my comments above about the intention of each library/repository at their conception and the future plan for what we hope each part of this project to become, I believe moving the data ingestion functionality to the CLI makes the most sense. 

the CLI is the data, model, and taxonomy manager. Having it be the document manager that feeds into all of the various other parts of the workflow logically makes sense to me.

cdoern (commented):
Left a few lengthy comments so I will let people read those individually. 

I generally agree with the idea that separate libraries does not make sense. Though, I strongly feel like logically it makes sense for the "cli" repo `instructlab/instructlab` should own data ingestion as it owns model, data, config, and filesystem management generally and is dedicated to the user experience of the project.

docs/sdg/sdg-refactor.md
jwm4: These are good points.

docs/sdg/sdg-refactor.md
jwm4: I guess I would argue that this is an additional level of detail to the pro in the line above, not a missing pro.   I will expand on the existing bullet and add sub-bullets rather than add more top-level bullets for these points.

docs/sdg/sdg-refactor.md
jwm4: Right, but having the CLI own pre and post processing doesn't force documentation and contract maintenance between the CLI itself and the functions that implement these features in the CLI.  That's easily mitigated by doing the documentation and contract maintenance between the CLI and functions even without being forced to do so.  But it is a point of concern that I think mirrors the equivalent point of concern if it is all kept in SDG.

docs/sdg/sdg-refactor.md
jwm4: Sure, but by definition user flow 2.1 does not have the user _installing_ the CLI.  The flow where you do install the CLI and do the same thing as 2.1 but from the CLI is 1.3.

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+
+Also note, that for all of the flows above that start with "they have a taxonomy", there are open questions around the complexity of the taxonomy format.  Some users want to be able to provide seed data and documents without explicitly or implicitly extending some sort of base taxonomy.  If we had something simpler than the existing taxonomy format that still included some way to specify context/question/answer tuples and references to full documents (i.e., not *chunks*), would we even call that a taxonomy or would it be something else?  If we call it something else, then some or all of the flows that take in a taxonomy might also be applicable for that something else.  For the remainder of this document, however, we will assume that any sort of simplified/easier-to-use variant of a taxonomy will also be called a "taxonomy".
```
williamcaban, line 41: 1. From lab flow, a "taxonomy" is a git repo; as such, it can be cloned/pushed/shared using the normal git constructs and flow. 
2. Another potential flow is the user bringing their documents in an InstructLab-supported schema (in which case, there might not be a traditional taxonomy tree)

It might be good to phrase the proposal such that both flows are possible

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,136 @@
...
+2. User installs the SDG library only.  They want to invoke any of the following as a library call from code they write:
+    - 2.1. They have a taxonomy with some new seed data they added to it.  They want to run SDG only without any postprocessing.
+    - 2.2. They have a set of seed context/question/answer tuples and optionallly a set of document chunks.  They want to run SDG only without any postprocessing.
+
+If I understand the guidance from our product management, the flows that our users want us to support here are 1.1., 1.2., 1.3, 1.3.1, and 2.2.  However, I am not sure that I understand the guidance.  Are we confident that we do want to support those flows and do not want to support any of the others?  More specifically, do we think the users who want 2.2 might be satisfied with 1.6 instead?  Both of those are SDG-only flows, but the latter is more developer focused and the former is more business-user focused.  Also, are we really confident that all of the "SDG only" customers really want to do their own document chunking?  Might there be some customers that have seed context/question/answer tuples and documents and want us to chunk the documents for them?
```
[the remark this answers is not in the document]
jwm4: > Extending existing SDG or merging two SDG datasets from two or more distinct runs

That sounds useful.  Is your expectation that support for this should be included in the acceptance criteria for the refactoring epic or is the idea that the refactoring epic should keep this in mind as a future requirement that will be implemented in a future epic?

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,147 @@
...
+
+- `ilab data prep` would handle all the preprocessing (the first three bullets in the Context section above, plus any additional preprocessing we add in the future).
+- `ilab data generate` would take as input some data in the same format that `ilab data prep` produces and would run the core synthetic data generation *only*.  Note that this is a breaking change from the current behavior of `ilab data generate`, but that may be acceptable because the user base is still small.
+- `ilab data process` would take as input some data in the same format that `ilab data generate` produces and would run the postprocessing (the last two bullets in the Context section above, plus any additional postprocessing we add in the future).
```
anastasds, line 53: `process` -> `postprocess`? 

Can we find a better name? As you mention, there may be other types of processing or postprocessing added in the future. What would be a name that indicates the purpose of this operation?

docs/sdg/sdg-refactor.md
anastasds: Moving more application logic into the CLI seems to b directly at odds with the intended future of having REST APIs. If we move more things into the CLI, wouldn't we shortly move them back out in order to have them be able to live behind a REST API?

jwm4: I guess my expectation is that the REST API implementation would either become _part_ of the CLI repo or would be a shim that calls into the CLI repo.  Either way, calling it the "CLI repo" seems misleading.  I would be much more comfortable with this change if we were calling that repo ( https://github.com/instructlab/instructlab ) the InstructLab _core_ repo or something like that instead of just "CLI repo".

docs/sdg/sdg-refactor.md
jwm4: @cdoern , you proposed this name as I recall.  I agree with Anastas that it is kind of generic.  I guess I like `postprocess` a little better.  At one point we had called it "mix" or "mixing" but I think it is already broader than just mixing and will probably get even broader in the future.

docs/sdg/sdg-refactor.md
anastasds: As a general comment, it seems unlikely that a taxonomy will always be identified with a fit repo. We should distinguish functionality/purpose from mechanism/implementation.

docs/sdg/sdg-refactor.md
anastasds: I would not expect REST APIs to be implemented in something called a CLI.

Hold on - when people say "the CLI" are they referring to the `instructlab/instructlab` repo as a whole? If so, I did not understand that at all and explains a bunch of my confusion!

jwm4: Yes, that's my understanding.  For example, when I click on https://github.com/instructlab/instructlab and then look at the readme, the title is "InstructLab 🐶 (ilab)" but the first section heading is "[Welcome to the InstructLab CLI](https://github.com/instructlab/instructlab#welcome-to-the-instructlab-cli)".  To me that suggests that the entire repo is referred to as the CLI.  I still think the name is misleading, but rethinking the naming of that repo would be outside the scope of this dev-doc.  Also, I can see a case for why from a _user_ standpoint, that name makes sense: if you install this piece of software then you can run InstructLab via the CLI.  It just doesn't make a lot of sense from a developer standpoint, which is normally the perspective I would expect to drive decisions about github repository names.

cdoern: yeah @anastasds  I think this statement is a part of the confusion:
`I would not expect REST APIs to be implemented in something called a CLI.`

`instructlab/instructlab` is referred to as the CLI (though it is much much much more), since it was once called the `instructlab/cli` repo in olden times

The planning for the REST API is for it to "wrap" this repo, given the substantial work going on to re-design it in a way that works better with an api:
https://github.com/instructlab/instructlab/issues/1548
https://github.com/instructlab/instructlab/issues/2325

your comment about moving things out of "the cli". This will most likely, never happen. If that were to happen, we would need to have a different repo for each command and workflow currently living in the "cli".

I think this point right here is a large part of the contention. In my planning for the engine runtime team, adding functionality like this makes _sense_ given that `instructlab/instructlab` is the orchestrator of data, models, config, taxonomy, etc. Having it be the orchestrator of documents and even more taxonomy workflows fits directly into the design of that repo

cdoern: I thin planning where this functionality should go based on the idea of moving core functionality out of `instructlab/instructlab` is a misnomer as long term planning of the instruct lab (even with a rest api) does not necessitate moving anything out of the repo. An example I like to use often is: https://github.com/containers/podman

Podman seamlessly maintains a user facing CLI and a rest api in the same repo that allows them to streamline development  and support various use cases of folks accessing Podman via `podman-remote`

again though, ** if the only idea against this functionality living in the instruct lab repo is the REST API, I think this is both an unrelated topic and a misnomer, as the team aiming to implement the REST API: @jaideepr97 @alinaryan, me @nathan-weinberg, have made no plans to move any functionality out of the repo.

this would necessitate  things like:

`instructlab/download`
`instructlab/config`
`instructlab/chat`
`instructlab/serve`

which are repos that I think introduce an extreme level of modularization

docs/sdg/sdg-refactor.md
cdoern: sure, post process sounds good to me!

anastasds:
I think that this document should be contextualized by the purpose of the change. "Refactor SDG" - okay, but why? To accomplish what goal? We can keep refactoring forever if we want to.

It seems like the goal here is to modularize the parts of the codebase that deal with the data augmentation phase of the end to end workflow. In order to modularize it effectively, we need to identify and distinguish pre-processing, data generation, and post-processing.

The question that follows is where these new, modularized pieces should live. This is another architectural decision - it looks everyone thinks it should go into the CLI, at least for now. In order to do that while still achieving the goal of modularization and without creating a new repository, we will have to do it in a certain way.

The methodical line of thinking here becomes much easier when individual decisions are treated individually, which is why I have been advocating for a series of ADRs that are each individually digestible, easier to discuss, and easier to peer review. That would look something like this:

1. The SDG codebase will be refactored in order to modularize based on pre-processing, data generation, and post-processing steps.
2. Pre-processing logic for SDG will be moved into the CLI codebase.
3. Post-processing logic for SDG will be moved into the CLI codebase
4. The SDG codebase will be designed around the principle of "dataset in, dataset out".

As I discuss in the [my ADR doc](https://drive.google.com/a/redhat.com/open?id=103fFmQe6KQfPDz8CJMCb_polBU6c-g9OP5jSV9CxRaY), treating each of these decisions individually and critically engaging with the "Consequences" section of the format would yield direct next steps. For example, in order to move pre-/post-processing out of the SDG codebase, there will like have to be inter-team collaboration, there will have to be a discussion around how to incorporate it into the CLI codebase, backwards compatibility will probably require some care, and so on - these things fall naturally out of that vehicle for technical writing and decision making.

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,147 @@
...
...unstaged qna.yaml files.  However, some users might have the essential elements of a taxonomy (seed context/question/answer tuples for both skills and knowledge plus documents for knowledge) but do not want to put that information in a tree it a git repo.  For the purposes of this document, we will refer to those essential elements as "raw seed content".  The "raw seed content" includes all of the things that go into a qna.yanl file.  In the current code base, the way InstructLab gets to the raw seed content is by identifying unstaged qna.yaml files from a local clone of a taxonomy.  However, in the future we might add functionality that allows users to simply point at some raw seed content without having to tie it to a github repository for a taxonomy.  If the raw seed content includes knowledge elements (not just skills) then those knowledge elements will have references to documents.  When the raw seed content is processed, the documents are fetched, converted, and chunked (the third step in the list above).  For this document, we will use the term "processed seed content" to refer to the outputs of that processing.  So to summarize the data structure terms being discussed here:
```
bbrowning, line 22: small typo - `qna.yanl` should be `qna.yaml`

docs/sdg/sdg-refactor.md
bbrowning: I too like splitting this out, although I'm not entirely sure if this is the right dimension or if conversion of documents with Docling should be entirely separate from traversing a taxonomy, extracting data from qna.yamls, and fetching the referenced knowledge. To put it more specifically, we may want to end up with something like `ilab document convert` that converts a set of input documents to a set of docling jsons or markdown that may be subsequently hand-edited, committed to a git repository, or something else. In the spirit of this document, we probably don't have to decide this here and now. But, document conversion may have a separate iteration loop from a user's point of view, and not just be something that happens automatically during SDG.

anastasds:
Thanks for putting this together!

I suggest considering architectural goals even if they are implied by things already listed, e.g. the REST API work will probably drive modularization.

What about API guarantees? State of CI? Documentation? Testing?

docs/sdg/sdg-refactor.md
jwm4: I posted a comment [in the CLI 1.0 dev doc](https://github.com/instructlab/dev-docs/pull/159#discussion_r1850587382) advocating for putting an end to using the phrase "the CLI" to refer to `instructlab/instructlab`.  I am hoping that if we can resolve that issue there, that this can be a reasonable conclusion to this thread too.  Let's see how that one turns out.

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Question 2: What should the commands be in the CLI?
+
+One way to support both 1.3.1 and 1.3.2 would be to have separate CLI commands for the preprocessing, core SDG, and postprocessing step .  Alternatively, a single CLI command that does all of these and also saves the outputs of preprocessing to disk would support 1.3.1 but *not* 1.3.2.  Even if we only want to support 1.3.1, having separate CLI commands for each step might be desirable because it is just more intuitive that if a user wants to save the outputs of preprocessing to disk to have a command to do that instead of having it be a "side effect" of an omnibus SDG command.  Here is a rough outline of what the separate commands would be:
+
+- `ilab data prep` would handle all the preprocessing (the first three bullets in the Context section above, plus any additional preprocessing we add in the future).
+- `ilab data generate` would take as input some data in the same format that `ilab data prep` produces and would run the core synthetic data generation *only*.  Note that this is a breaking change from the current behavior of `ilab data generate`, but that may be acceptable because the user base is still small.
```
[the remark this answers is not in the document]
jwm4: @cdoern and I discussed this, and he felt strongly that providing backward compatibility was not worth the cost.  In general, the value of backward compatibility is high when the user base is large relative to the expected future user base.  I think that's mostly not the case for InstructLab, so I agree with @cdoern .

docs/sdg/sdg-refactor.md
khaledsulayman: I lean towards `ilab data mix` as well as I don't think ambiguity or generic-ness helps us much here.

In mixing, we already have the concept of the 'recipe' yaml files which serve as formulas by which to sample and mix data. To that end, I think we should have the mixing command be cohesive and self-descriptive, if we want to empower users to tweak or write their own recipe files. This way, at least the flags we expose in the `ilab data mix` command make sense contextually.

Also, I'd argue that if any clear other candidates emerge as post processing steps we'd want to do, if they are different enough to not be lumped into mixing, that'd probably justify having a separate command if we do want to make it configurable to users.

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,159 @@
...
+- As with Option 1, this approach suffers from the fact that keeping interconnected components in the same repository provides less pressure to consistently document the API contracts between them.  In the case of Option 1, the interconnected components that would not have as much pressure to be documented would be preprocessing/postprocessing and core SDG.  In the case of Option 3, the interconnected components that would not have as much pressure to be documented would be the command-line interface code and preprocessing/postprocessing.  However, in both cases, this con could be alleviated by just having the discipline to document the APIs well even without such pressure.
+- As with Option 2, this approach would not enable user flow 2.1.  That's fine since it is not on our requirements list.
+
+Conclusion:
+
+- This seems like a reasonable option.  The cons are mostly manageable.  However, overall the pros of Option 1 seem more compelling.
```
khaledsulayman, line 129: This line might be better updated to reflect what is stated below under '## Decisions'

docs/sdg/sdg-refactor.md
```diff
@@ -0,0 +1,159 @@
...
+
+Con:
+
+- Avoids the pros of both Option 1 and Option 2.
+- As with Option 1, this approach involves a lot of coordination.  There are a lot of stakeholders involved in the `instructlab/instructlab` repository and locating preprocessing and postprocessing there drags those stakeholders into issues relating to preprocessing and postprocessing.  However, as with Option 1, coordinating across stakeholders is something an open source project needs to do well anyway to remain engaged with the community.
+- As with Option 1, this approach suffers from the fact that keeping interconnected components in the same repository provides less pressure to consistently document the API contracts between them.  In the case of Option 1, the interconnected components that would not have as much pressure to be documented would be preprocessing/postprocessing and core SDG.  In the case of Option 3, the interconnected components that would not have as much pressure to be documented would be the command-line interface code and preprocessing/postprocessing.  However, in both cases, this con could be alleviated by just having the discipline to document the APIs well even without such pressure.
```
khaledsulayman, line 124: as it pertains to this point, I will say that the dataset in / dataset out principle should also alleviate some of the burden

nathan-weinberg (approved):

docs/sdg/sdg-refactor.md
jwm4: I have been hearing conflicting reports about whether all of the postprocessing we do really is data mixing.  If we have a command like `ilab data mix` and we want to do any postprocessing that is not really data mixing then we have two unappealing options:

1. Add that as part of `ilab data mix` and accept that the name is not really very accurate.
2. Add another command that runs before or after `ilab data mix` or possibly one command that runs before and another that runs after for pre-mix postprocessing and post-mix postprocessing.

My main concern is that we will actually pick option two because we have a long track record of not updating our names for concepts when those names no longer apply.  For example, the qna.yaml files for skills have `question` and `answer` pairs because that is the format but really most skills (unlike knowledge) are not really about answering questions but rather about fulfilling a request and providing a response.  Renaming things is a hassle so we stick with the old names.  This suggests that going forward we should try to use more generic names that will be broad enough to cover not only what we do but also what we expect to do.

anastasds: One of the goals of the domain modeling workshop was to encourage folks to think about names from a more user-centric perspective, which would mean trying to choose more accurate names at least. This seems like a good time to spend some time picking good names, like how we spent time trying to name our bounded contexts at the workshop.

anastasds:
> I think that this document should be contextualized by the purpose of the change. "Refactor SDG" - okay, but why? To accomplish what goal? We can keep refactoring forever if we want to.
>
>It seems like the goal here is to modularize the parts of the codebase that deal with the data augmentation phase of the end to end workflow. In order to modularize it effectively, we need to identify and distinguish pre-processing, data generation, and post-processing.

I have not seen a response on this and feel that it is important.

docs/sdg/sdg-refactor.md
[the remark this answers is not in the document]
jwm4: FWIW, I have updated the document to leave both options open since we don't seem to be resolving this here, and we want the rest of what's in this document to merge if possible.

docs/sdg/sdg-refactor.md
jwm4: Done!

jwm4:
@jjasghar writes:
> So reading this, isn't going back to puttind sdg back into the "core" package defeating the purpose of of "core" anyway. I thought the goal per the renaming of core was to start modularizing things.

Sorry for the confusion!  The plan here is to keep the *actual* synthetic data generation in the `/sdg` repo where it is now, but to move all the preprocessing and postprocessing to core.  I will update the "goals" section to make that clearer.

jjasghar:
This is extremely well thought out, and clear now. Thank you. I approve this to be merged.

jjasghar (approved):

PR #160, opened 2024-11-19
Introduce dev-doc for process management

cdoern:
ilab should be able to run detached processes, re-attach to them, and monitor them. This document introduces a first iteration of that process management

### DIFF
--- a/docs/cli/ilab-processes.md
+++ b/docs/cli/ilab-processes.md
@@ -1,0 +1,64 @@
+# Processes in InstructLab
+
+The ability to detach from processes is crucial to the user experience of InstructLab. However, the concept of multi-processing, process management, and the monitoring of processes is very complex.
+
+It is important to try and add this concept in as simply as possible, expanding on the state reporting, logging, and other features as we go along.
+
+## Phased approach to InstructLab Processes
+
+This document is going to describe phase 1 of implementing processes in InstructLab. Phase 1 is to be described as the "ilab simple process management system". This will depend purely on python packages, PID tracking, and log files to create the experience of detachable processes. The key here is the concept of the UUID, allowing a future REST API to keep track of InstructLab processes using these unique identifiers.
+
+We can re-visit all this in phase 2, when we discuss if we want to utilize something like systemd or a more in-depth process-monitor repo to track processes.
+
+### Phase 1
+
+Phase one would focus on adding the concept of detaching from processes, re-attaching to them, and managing the various artifacts from the processes.
+
+Process management would only apply to `ilab data generate` and `ilab model train` in a first iteration. This would be followed by commands like `ilab model evaluate`, `ilab model serve`, and `ilab model download`. All of these commands have long running processes that would benefit from detachment.
+
+The workflow would allow for:
+
+`ilab data generate -dt` (run a detached generation process)
+`ilab model train -dt` (run a detached training process)
+
+`ilab process list`
+
+```console=
++------------+-------+--------------------------------------+------------------------------------------------------------------------------------------------------------------+----------+
+| Type       | PID   | UUID                                 | Log File                                                                                                         | Runtime  |
++------------+-------+--------------------------------------+------------------------------------------------------------------------------------------------------------------+----------+
+| Generation | 39832 | 82d00a5b-5ed5-4cfd-9a75-a87e4f420b27 | /Users/charliedoern/.local/share/instructlab/logs/generation/generation-82d00a5b-5ed5-4cfd-9a75-a87e4f420b27.log | 69:26:28 |
+| Generation | 40791 | 09f9d301-4fd9-4045-bfda-8a56f1d96016 | /Users/charliedoern/.local/share/instructlab/logs/generation/generation-09f9d301-4fd9-4045-bfda-8a56f1d96016.log | 68:45:40 |
+| Generation | 47390 | 4ccabfa5-604f-49c6-b5c3-730ce328d62a | /Users/charliedoern/.local/share/instructlab/logs/generation/generation-4ccabfa5-604f-49c6-b5c3-730ce328d62a.log | 67:26:33 |
+| Generation | 50872 | 093ac2e9-080c-45fe-89c5-43d508d6369c | /Users/charliedoern/.local/share/instructlab/logs/generation/generation-093ac2e9-080c-45fe-89c5-43d508d6369c.log | 05:24:56 |
++------------+-------+--------------------------------------+------------------------------------------------------------------------------------------------------------------+----------+
+```
+
+`ilab process attach <UUID>`
+
+This command would re-attach to the given process, allowing to user to view the live logs of the process. `attach` would trail the log file and listen for user-input to kill the process.
+
+These commands will be done in a very simple way at first using the following architecture:
+
+1. a detached process be re-attachable by tailing the log file and then allowing the user to ctrl+c the process as normal using `KeyboardInterrupt`
+2. The process registry will be maintained for tracking UUIDs created via the `uuid` python package, the PID of the actual process, a `log_file` where the process will be outputting its logs to so that the user can re-attach, and the start time of the process. The log file directory will be tracked using our `DEFAULTS` package and will be standard throughout releases.
+
+The general flow would be:
+
+1. a user runs `ilab data generate -dt`
+2. a UUID, PID, and log file is added to the process registry.
+3. the process would exit, and print the UUID of the sdg run
+4. a user could attach to this process using `ilab process attach <UUID>`.
+5. This command would look in the process registry for the PID and/or UUID, get the log file, tail the log file, and listen for a ctrl+c keyboard interrupt.
+
+This allows us to detach from processes while still running them in the background and maintain log files all without the use of anything other than UUID and subprocess.
+
+#### Log file management
+
+If existing log files from the various libraries exist, those will be used in this scenario. If they do not, InstructLab will manage writing process logs to disk. Regardless of whether the libraries maintain their own log file, InstructLab will need to co-locate the log files in a centralized directory.
+
+If a log file exists, it will be copied and renamed into the following directory format:
+
+`~/.local/share/instructlab/logs/<command_name>/<command_name>-<timestamp>.log`
+
+If the log file does not exist, InstructLab will create one with this format. Libraries are responsible for standardizing where their logs are stored if they already exist so the Core package can access them in a uniform fashion and copy them to the proper directory.
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -29,6 +29,7 @@ Containerfile
 cpp
 cuBLAS
 CUDA
+ctrl
 customizations
 CVE
 CVEs
@@ -131,6 +132,7 @@ Params
 Pareja
 PEFT
 Pereira
+PID
 PlantUML
 PLOS
 pluggable
@@ -188,8 +190,10 @@ Standup
 subcommand
 subcommands
 subdirectory
+subprocess
 Sudalairaj
 supportability
+systemd
 Taj
 tatsu
 TBD
@@ -209,6 +213,8 @@ ui
 unquantized
 unstaged
 USM
+UUID
+UUIDs
 UX
 venv
 Vishnoi


docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,53 @@
...
+
+The ability to detach from processes is crucial to the user experience of InstructLab. However, the concept of multi-processing, process management, and the monitoring of processes is very complex.
+
+It is important to try and add this concept in as simply as possible, expanding on the state reporting, logging, and other features as we go along.
+
+## Phased approach to InstructLab Processes
+
+This document is going to describe phase 1 of implementing processes in InstructLab. Phase 1 is to be described as the "ilab simple process management system". This will depend purely on python packages, PID tracking, and log files to create the experience of detachable processes. The key here is the concept of the UUID, allowing a future REST API to keep track of InstructLab processes using these unique identifiers. 
+
+We can re-visit all this in phase 2, when we discuss if we want to utilize something lils systemd or a more in-depth process-monitor repo to track processes.
```
jaideepr97, line 11:
```suggestion
We can re-visit all this in phase 2, when we discuss if we want to utilize something like systemd or a more in-depth process-monitor repo to track processes.
```

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,53 @@
...
+`ilab process list`
+
+```console=
++----------------------------------------+------------+-------------+---------+
+| Proccess UUID                          | Type       | Duration    | Status  |
++----------------------------------------+--------------------------+---------+
+| train-2423813fhthfbnaj                 | training   | 8 hr 32 min | Running |
+| generation-6768769599sdbjklm           | generation | 3 hr 15 min | Running |
+| generation-1345661hlsmma               | generation | 22 min      | Errored |
++----------------------------------------+---------------------+--------------+
```
jaideepr97, line 31: maybe include the PID here so users can manually kill the process if it hangs?

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,53 @@
...
+
+```
+
+`ilab process attach <UUID>`
+
+This command would re-attach to the given process, allowing to user to view the live logs of the process.
```
jaideepr97, line 37: Does this mean the process outputs both, to stdout as well as a log file at all times?

docs/cli/ilab-processes.md
cdoern: whoops, thanks

docs/cli/ilab-processes.md
cdoern: we would tail the log file, meaning that its only really writing to the file but we are watching what is being written when the user attaches, and we listen for ctrl+c to kill the process

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Phased approach to InstructLab Processes
+
+This document is going to describe phase 1 of implementing processes in InstructLab. Phase 1 is to be described as the "ilab simple process management system". This will depend purely on python packages, PID tracking, and log files to create the experience of detachable processes. The key here is the concept of the UUID, allowing a future REST API to keep track of InstructLab processes using these unique identifiers. 
+
+We can re-visit all this in phase 2, when we discuss if we want to utilize something lils systemd or a more in-depth process-monitor repo to track processes.
+
+### Phase 1
+
+Phase one would focus on adding the concept of detaching from processes. This would be implemented in `ilab 0.23.0`.
```
nathan-weinberg, line 15: Do we want to commit to a certain version in this doc?

docs/cli/ilab-processes.md
nathan-weinberg: I guess we always output to a log and "attaching" just moves the terminal to essentially `tail -f` that log? Except we still would need users to be able to `Crtl-C` and such

nathan-weinberg: Some of this seems to be covered below

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,53 @@
...
+
+These commands will be done in a very simple way at first using the following architecture:
+
+1. The backbone of this process mangement will be the python multiprocessing package, using `Process` to start the detached commands and `Manager` to maintain an in-code dictionary of the existing processes.
+2. a detached process be re-attachable by tailing the log file and then allowing the user to ctrl+c the process as normal using `KeyboardInterrupt`
+3. A `Manager.dict` process registry will be maintained for tracking UUIDs created via the `uuid` python package, the PID of the actual process, and a `log_file` where the process will be outputting its logs to so that the user can re-attach. this log file dir will be tracked using our `DEFAULTS` package and will be standard throughout releases.
+    
+The general flow would be:
+
+1. a user runs `ilab data generate -dt`
```
nathan-weinberg, line 47: Which processes explicitly would get "detachablility" functionality? can we have an explicit list as part of this doc?

jaideepr97: probably the longest running ones to begin with at least
download, sdg, train

docs/cli/ilab-processes.md
cdoern: added a small section

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+
+1. The backbone of this process management will be the python multiprocessing package, using `Process` to start the detached commands and `Manager` to maintain an in-code dictionary of the existing processes.
+2. a detached process be re-attachable by tailing the log file and then allowing the user to ctrl+c the process as normal using `KeyboardInterrupt`
+3. A `Manager.dict` process registry will be maintained for tracking UUIDs created via the `uuid` python package, the PID of the actual process, and a `log_file` where the process will be outputting its logs to so that the user can re-attach. this log file directory will be tracked using our `DEFAULTS` package and will be standard throughout releases.
+
+The general flow would be:
+
+1. a user runs `ilab data generate -dt`
+2. a UUID, PID, and log file is added to the process registry.
+3. the process would exit, and print the UUID of the sdg run
```
reidliu41, line 51: I tried it will block it there after run `ilab data generate -dt` if used multiprocessing, maybe you have a better way to handle..

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+
+`ilab process attach <UUID>`
+
+This command would re-attach to the given process, allowing to user to view the live logs of the process. `attach` would trail the log file and listen for user-input to kill the process.
+
+These commands will be done in a very simple way at first using the following architecture:
+
+1. The backbone of this process management will be the python multiprocessing package, using `Process` to start the detached commands and `Manager` to maintain an in-code dictionary of the existing processes.
+2. a detached process be re-attachable by tailing the log file and then allowing the user to ctrl+c the process as normal using `KeyboardInterrupt`
+3. A `Manager.dict` process registry will be maintained for tracking UUIDs created via the `uuid` python package, the PID of the actual process, and a `log_file` where the process will be outputting its logs to so that the user can re-attach. this log file directory will be tracked using our `DEFAULTS` package and will be standard throughout releases.
```
reidliu41, line 45: if save it in memory, that mean no need to record after the task done? Or if user want to check the record later, or do some review, or else...

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+```console=
++----------------------------------------+------------+-------------+-------------+---------+
+| Proccess UUID                          | PID        |  Type       | Duration    | Status  |
++----------------------------------------+--------------------------+-------------+---------+
+| train-2423813fhthfbnaj                 |  33451     |  training   | 8 hr 32 min | Running |
+| generation-6768769599sdbjklm           |  35543     |  generation | 3 hr 15 min | Running |
+| generation-1345661hlsmma               |  45424     |  generation | 22 min      | Errored |
++----------------------------------------+---------------------+------------------+---------+
+
+```
```
reidliu41, line 35: sorry, I didn't test before, just curious, is the host can do or there is a ability to handle multi-tasks at the same time.

docs/cli/ilab-processes.md
cdoern: Saving process to a log file opens `ilab` up into being a process management system, which I don't think is the goal of this work.

There is a fine line between managing processes seamlessly for users, and maintaining process state for users to see long after processes are completed. 

We will save the log files of course.

docs/cli/ilab-processes.md
cdoern: you are right I think, I am going to have to switch to subprocess which is unfortunate.

docs/cli/ilab-processes.md
nathan-weinberg: Let's remove the reference, it's not nessessary

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+
+This document is going to describe phase 1 of implementing processes in InstructLab. Phase 1 is to be described as the "ilab simple process management system". This will depend purely on python packages, PID tracking, and log files to create the experience of detachable processes. The key here is the concept of the UUID, allowing a future REST API to keep track of InstructLab processes using these unique identifiers.
+
+We can re-visit all this in phase 2, when we discuss if we want to utilize something like systemd or a more in-depth process-monitor repo to track processes.
+
+### Phase 1
+
+Phase one would focus on adding the concept of detaching from processes. This would be implemented in `ilab 0.23.0`.
+
+Process management would only apply to `ilab data generate` and `ilab model train`.
```
nathan-weinberg, line 17: For now or in general?

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+
+`ilab process attach <UUID>`
+
+This command would re-attach to the given process, allowing to user to view the live logs of the process. `attach` would trail the log file and listen for user-input to kill the process.
+
+These commands will be done in a very simple way at first using the following architecture:
+
+1. The backbone of this process management will be the python subprocess package, using `Popen` to start the detached commands and a localized dictionary to maintain the existing processes. This will need to be writtent to disk to ensure persistence. Using a `Multiprocessing.Manager()` does not guarantee persistence if the main process exits, which it does in the case of detatching.
```
nathan-weinberg, line 43: is this a built-in package?

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+2. a detached process be re-attachable by tailing the log file and then allowing the user to ctrl+c the process as normal using `KeyboardInterrupt`
+3. The process registry will be maintained for tracking UUIDs created via the `uuid` python package, the PID of the actual process, a `log_file` where the process will be outputting its logs to so that the user can re-attach, and the start time of the process. The log file directory will be tracked using our `DEFAULTS` package and will be standard throughout releases.
```
nathan-weinberg, line 45: is this a built-in package?

docs/cli/ilab-processes.md
nathan-weinberg: What about eval?

cdoern: I can add a note about download, eval, and serve. I think this doc was initially meant to be a phase 1, but I can note how in subsequent releases we can add those commands.

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,65 @@
...
+| Proccess UUID                          | PID        |  Type       | Duration    | Status  |
++----------------------------------------+--------------------------+-------------+---------+
+| train-2423813fhthfbnaj                 |  33451     |  training   | 8 hr 32 min | Running |
+| generation-6768769599sdbjklm           |  35543     |  generation | 3 hr 15 min | Running |
+| generation-1345661hlsmma               |  45424     |  generation | 22 min      | Errored |
++----------------------------------------+---------------------+------------------+---------+
+
+```
+
+`ilab process attach <UUID>`
```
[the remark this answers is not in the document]
cdoern: That could also work! I just wonder if we'd like to have a uuid system in general. Could be useful for orchestration down the line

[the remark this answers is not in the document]
danmcp: Having the uuid version is probably useful as well

docs/cli/ilab-processes.md
cdoern: yep, you can have multiple processes at once, especially SDG if you set gpu_layers properly in your config

docs/cli/ilab-processes.md
cdoern: switched.

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,64 @@
...
+`ilab data generate -dt` (run a detached generation process)
+`ilab model train -dt` (run a detached training process)
+
+`ilab process list`
+
+```console=
++------------+-------+--------------------------------------+------------------------------------------------------------------------------------------------------------------+----------+
+| Type       | PID   | UUID                                 | Log File                                                                                                         | Runtime  |
++------------+-------+--------------------------------------+------------------------------------------------------------------------------------------------------------------+----------+
+| Generation | 39832 | 82d00a5b-5ed5-4cfd-9a75-a87e4f420b27 | /Users/charliedoern/.local/share/instructlab/logs/generation/generation-82d00a5b-5ed5-4cfd-9a75-a87e4f420b27.log | 69:26:28 |
```
nathan-weinberg, line 30: Having `Type` like this would require some kind of mapping from the command to these descriptions - maybe it would be simpler to simply map to the command, i.e. instead of `Generation` do `data generate`?

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,64 @@
...
+
+1. a user runs `ilab data generate -dt`
+2. a UUID, PID, and log file is added to the process registry.
+3. the process would exit, and print the UUID of the sdg run
+4. a user could attach to this process using `ilab process attach <UUID>`.
+5. This command would look in the process registry for the PID and/or UUID, get the log file, tail the log file, and listen for a ctrl+c keyboard interrupt.
+
+This allows us to detach from processes while still running them in the background and maintain log files all without the use of anything other than UUID and subprocess.
+
+#### log file management
```
nathan-weinberg, line 56:
```suggestion
#### Log file management
```

docs/cli/ilab-processes.md
```diff
@@ -0,0 +1,64 @@
...
+
+If existing log files from the various libraries exist, those will be used in this scenario. If they do not, InstructLab will manage writing process logs to disk. Regardless of whether the libraries maintain their own log file, InstructLab will need to co-locate the log files in a centralized directory.
+
+If a log file exists, it will be copied and renamed into the following directory format:
+
+`~/.local/share/instructlab/logs/<command_name>/<command_name>-<timestamp>.log`
+
+If the log file does not exist, InstructLab will create one with this format. Libraries are responsible for standardizing where their logs are stored if they already exist so the CLI can access them in a uniform fashion and copy them to the proper directory.
```
nathan-weinberg, line 64:
```suggestion
If the log file does not exist, InstructLab will create one with this format. Libraries are responsible for standardizing where their logs are stored if they already exist so the Core package can access them in a uniform fashion and copy them to the proper directory.
```

nathan-weinberg (approved):
few nits but otherwise LGTM

docs/cli/ilab-processes.md
cdoern: I already have this implemented, either way we would have to pass an argument to the process creation code. I think passing just `Generation` is cleaner.

cdoern:
This has been open for about a month now with two approvals. I am going to merge this as is, but I welcome and amendments in follow up PRs

PR #161, opened 2024-12-02
RAG ingestion and chat pipelines

dmartinol:
Introducing `ilab` commands changes to support the RAG ingestion and chat pipelines:
- RAG ingestion: a new command that starts from pre-processed user docs (see https://github.com/instructlab/dev-docs/pull/148) to ingest the calculated embeddings in a configurable vector DB.
- RAG chat: augment the context of the chat command with the result of a similarity search executed on the ingested database, to increase the accuracy of the response.


anastasds:
I think that the chat `/` options probably deserve some discussion and also look like they may not be a priority, but, other than that, this looks reasonable to me.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+The generated embeddings can later be retrieved to enrich the context for RAG-based chat pipelines.
+
+#### Running the pipeline with v1.4
+Currently, there is no `ilab data rag` command available to execute the RAG ingestion pipeline. As a result, it is not possible
+to address the limitations of version 1.4 using the configuration options detailed below. 
+Alternatively, we propose providing a Jupyter notebook or a standalone script to run the pipeline with default settings.
+
+The script should guide users in overriding default options and generating the RAG artifacts in the configured vector database instance. By default, support for MilvusLite will be included.
+
+> **ℹ️ CLI-based alternative:** This is an alternative to run both the document transformation and the document ingestion in
```
dmartinol, line 52: added one option to overcome today's limitations and run the docs and ingestion pipelines with no CLI changes

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+  * Podman Desktop
+  * Red Hat OpenShift AI
+  * Internal Red Hat CI systems for products or services (e.g., Lightspeed products)
+
+## 3. Proposed Commands
+**Note**: In the context of version 1.4, currently under development, no changes to the command-line interface should be
+allowed. Therefore, we also propose alternative approaches to run the same RAG pipelines using existing `ilab` commands or
+other provided tools.
+
+### 3.1 RAG Ingestion Pipeline Command
```
franciscojavierarceo, line 24: How will this be handled for fine tuning? 

I would expect that it ends up that many of the components end up being re-invented here.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+The proposal is to add a `rag` subgroup under the `data` group, with an `ingest` command, like:
+```
+ilab data rag ingest /path/to/docs/folder
+```
+
+#### Command Purpose
+This command processes embeddings generated from documents located in the */path/to/docs/folder* folder and stores them in a vector database. These embeddings are intended to be used as augmented context in a Retrieval-Augmented Generation (RAG) chat pipeline.
+
+#### Assumptions
+The documents must be in JSON format and pre-processed using the `docling` tool.
```
franciscojavierarceo, line 34: I think you only have to assume the format, not necessarily that it was processed via docling.

franciscojavierarceo: While it may be true, it's not a strict requirement.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+| rag.vectordb.uri | The vector DB service URI | `./rag-output.db` | `--vectordb-uri` |
+| rag.vectordb.token | The vector DB connection token | | `--vectordb-token` |
+| rag.vectordb.username | The vector DB connection username | | `--vectordb-username` |
+| rag.vectordb.password | The vector DB connection password | | `--vectordb-password` |
+| rag.embedding_model.name | The embedding model name | `sentence-transformers/all-minilm-l6-v2` | `--model` |
+| rag.embedding_model.token | The token to download private models | | `--model-token` |
+| **TODO** evaluation framework options | | | |
+
+Equivalent YAML document for the newly proposed options:
+```yaml
```
franciscojavierarceo, line 94: This looks like it could pretty easily be structured in Feast.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+
+## 3. Proposed Commands
+**Note**: In the context of version 1.4, currently under development, no changes to the command-line interface should be
+allowed. Therefore, we also propose alternative approaches to run the same RAG pipelines using existing `ilab` commands or
+other provided tools.
+
+### 3.1 RAG Ingestion Pipeline Command
+The proposal is to add a `rag` subgroup under the `data` group, with an `ingest` command, like:
+```
+ilab data rag ingest /path/to/docs/folder
```
jwm4, line 27: Some thoughts on this:
1. I guess broadly speaking, I was expecting the proposal for how this should be reflected in the command-line interface to come from members of the engine team, e.g., @cdoern .  However, I guess it is fine for us to propose things here and iterate with them.
2. I don't like `rag ingest` here.  I think we want something that describes what we're doing here, which is building an index rather than bringing in the term "RAG" which describes the _feature_ but not really what this specific step is doing.
3. I'm not sure how to respond to the `/path/to/docs/folder` part.  We definitely want some sort of affordance around a flow where you do an `ilab data generate` and then `ilab` just _knows_ where the outputs of that step are rather than you needing to specify it.  However, some other affordance for being able to override that location also makes sense to me.  So maybe if the folder is optional that solves this?
4. We need to figure out how this fits in with the broader refactor being considered in https://github.com/instructlab/dev-docs/pull/155
5. Also, I would like a flow some day where you can just point this step at source documents and it runs docling for you, but that's lower priority than the flow that is more tightly connected with SDG (or at least SDG preprocessing).

docs/cli/ilab-rag-retrieval.md
jwm4: Yes, I think that's right, but the format is the output format of docling, not any JSON file.  It needs to match the docling JSON schema so we can use hierarchical chunking, at least for now.  We can add support for other formats in the future.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+This command processes embeddings generated from documents located in the */path/to/docs/folder* folder and stores them in a vector database. These embeddings are intended to be used as augmented context in a Retrieval-Augmented Generation (RAG) chat pipeline.
+
+#### Assumptions
+The documents must be in JSON format and pre-processed using the `docling` tool.
+
+**Note**: The expected JSON schema is the same that we have when we generate knowledge data with `ilab data generate` and we
+add a reference document to the `qna.yaml` document(s).
+
+#### Supported Databases
+The command supports multiple vector database types. By default, it uses a local `MilvusLite` instance stored at `./rag-output.db`.
```
jwm4, line 40: I think we should have a separate dev-doc for this decision.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+>     output: /path/to/processed-docs/folder
+> rag:
+>   ...
+> ```
+> When `generate.rag.enabled` is set to `true`, the `ilab data generate command` executes the document ingestion pipeline, transforming 
+> the source documents in the `generate.rag.input` folder into the configured vector database:
+> * Step 1: Transform the documents using the existing `SDG` modules (powered by `docling`).
+> * Step 2: Ingest the pre-processed artifacts from `generate.rag.output` into the configured database.
+
+**TODO** Introduce the evaluation framework
```
jwm4, line 72: I'd prefer to have a separate dev-doc for evaluation framework because I think it is going to take longer to get that sorted out than some of these more urgent issues.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+
+
+### 3.7 Workflow Visualization
+<!-- https://excalidraw.com/#json=hyDopdXaChbf6TNuLGYE0,WkLQBF84vypx0JwAL5iwzA -->
+Ingestion pipeline:
+![ingestion-mvp](../images/ingestion-mvp.png)
+Chat pipeline enriched by RAG context:
+![rag-chat](../images/rag-chat.png)
+
+### 3.8 Proposed Implementation Stack
```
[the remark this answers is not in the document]
jwm4: Is there documentation for Feast?  If so, can you post a link here?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,264 @@
...
+The following technologies form the foundation of the proposed solution:
+
+* [Haystack](https://haystack.deepset.ai/): Framework for implementing RAG pipelines and applications.
```
jwm4, line 252: We need a dev-doc for this decision, I think.

anastasds:
@jwm4 

>This document is a good start, but needs a lot more input from a lot of stakeholders

I think I need to clarify my position when I commented with a general "looks good to me" - I think that you raise some very valid points, but also I am operating under the assumption that this document serves as a proposal for "directionally where to head right now" and that any less-than-high-level details can, will, and probably should change as we understand the problem domain better during execution. I think that a useful modus operandi is to get a few key stakeholders to give a general approval and that that is enough to get started, and then have continuous feedback cycles all the time going forward to course correct as necessary. Analysis paralysis is a real effect that is best avoided.

Not trying to beat a dead horse but this is partly why I keep advocating for atomic decision records like ADRs over all-encompassing design docs like this. A general development roadmap is a necessary thing to have, but nobody will ever have enough information to design a full system specification, especially in the context of a marketplace and a large development organization. The only constant is change.

docs/cli/ilab-rag-retrieval.md
jwm4: Are you asking about fine-tuning of the embedding model or the response-generation model or both?

franciscojavierarceo: Both.

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
franciscojavierarceo: https://docs.feast.dev/

https://docs.feast.dev/getting-started/quickstart#what-is-feast is probably a good overview

RAG support is alpha and actively being worked on.

docs/cli/ilab-rag-retrieval.md
franciscojavierarceo: +1

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
jwm4: I see support for indexing and retrieving in the [alpha-vector-database](https://docs.feast.dev/reference/alpha-vector-database) link, but not other aspects of RAG (e.g., response generation).  Is that also being worked, or is the RAG focus specifically indexing and retrieval?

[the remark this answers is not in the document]
franciscojavierarceo: It is exclusively focused on indexing and retrieval. Generation (i.e., inference) would be a separate service. In RHOAI it'd be vLLM with KServe.

docs/cli/ilab-rag-retrieval.md
dmartinol: I can clarify that it follows the `docling` format because we should use the `instructlab-sdg` package that depend on that tool. As soon as they switch to a different format/tool, we will be updated accordingly.

docs/cli/ilab-rag-retrieval.md
dmartinol: Agree, after you latest messages in the Slack discussion channel it's definitely worth parking this point for now.

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
dmartinol: @franciscojavierarceo wrt to the RAG requirement related to ilab, how can we position Feast?
are you envisioning it as a configurable option for the pipeline steps or what else?
- transformation: instead of sdg/docling
- ingestion: instead of configured vectordb and ingestion stack
- retrieval: instead of selected retrieval stack

docs/cli/ilab-rag-retrieval.md
dmartinol: > I don't like rag ingest here
2. Me neither, but I was waiting for the closure of the discussion on the related command at #148 that, IIUC, should be the preliminary step before running the embedding ingestion. Depending on the selected verb, we can update this proposal accordingly (maybe, like `ilab data index` or `ilab data generate index`?).

> I'm not sure how to respond to the /path/to/docs/folder part
3. Again, this followed the proposal for the other PR, that has both `--input` and `--output` options.

> We definitely want some sort of affordance around a flow where you do an `ilab data generate` and then ilab just knows where the outputs of that step are rather than you needing to specify it

If this is a valid use case, then yes and the parameter will be optional. We have to think carefully of how to auto-detect the json docs in this case, as the `datasets` folder is "versioned" for each `data generate` execution, so I assume the requirement is to pick all the files from the latest `documents-*` subfolder.

docs/cli/ilab-rag-retrieval.md
franciscojavierarceo: +1

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
franciscojavierarceo: > @franciscojavierarceo wrt to the RAG requirement related to ilab, how can we position Feast? are you envisioning it as a configurable option for the pipeline steps or what else?

I'd position Feast as basically handling the local experience to ingest and transform data nicely for us and to enable us to scale InstructLab RAG to RHOAI.

And to your specific items:
> 
> * transformation: instead of sdg/docling
Docling is a particular transformation applied to some input data that returns a particular set of output data. At the moment, this is done on some set of batch data. In live production use cases these documents likely exist in a production service that may change and require re-processing upon changes (i.e., ingesting individual documents and processing and persisting them either synchronously or asynchronously). SDG is retrieval + inference in the sense that the retrieved documents are injected in the context before running inference. Even if the retrieval is on batch data, there is important metadata that should be stored during the processing (e.g., if embeddings are used, which model, version of code, etc.). 
> * ingestion: instead of configured vectordb and ingestion stack
Feast has the right primitives for ingestion and the flexibility to support Batch, Streaming, and Request driven data. 

> * retrieval: instead of selected retrieval stack
Again, feast's `get_online_features` or `retrieval_online_documents` apis are good defaults already and the benefits here are that ilab can set the best defaults for ilab using Feast without having to implement them.

I can commit to upstreaming things we feel are a benefit to Feast, the point I'm trying to highlight with all of this is that we will find ourselves having to reinvent many if not all of the same details that a feature store does because ultimately prompts, documents, and chat history (more generally tokens) are just features into a model. Since RHOAI is investing into Feast, it is a natural consideration that doesn't lock us into a single VectorDB and doesn't expose us to conflicts with companies changing the license as it is part of the LF AI and Data project.

Let me know if that makes sense.

dmartinol:
I will soon publish an updated version with the outcome of the discussion with the ilab Runtime (aka CLI) team.

dmartinol:
@cdoern Could you please TAL and involve relevant people?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+#### Assumptions
+The provided documents must be in JSON format according to the InstructLab schema: this is the schema generated
+when transforming knowledge documents with the `ilab data generate` command (see 
+[Getting Started with Knowledge Contributions][1]).
+
+To simplify the execution of the transformation step, we introduce a `--transform` option which also includes the documents
+transformation, leveraging on the `instructlab-sdg` modules. 
+
+### Why We Need It
+This command streamlines the `ilab data generate` pipeline and eliminates the requirement to define a `qna` document,
```
anastasds, line 52: This is a really far-reaching design decision that could have a lot of consequences for the product. Looks like this came out of a meeting with the engine runtime team, was that recorded?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+
+The goal is not to generate training data for InstructLab-trained models but to utilize the documents for RAG 
+workflows with pre-tuned models.
+
+#### Supported Databases
+The command may support various vector database types. A default configuration will align with the selected 
+InstructLab technology stack.
+
+#### Usage
+The generated embeddings can later be retrieved to enrich the context for RAG-based chat pipelines.
```
anastasds, line 65: The embeddings themselves, not text to be substituted into a prompt template?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+    default='word',
+    envvar="ILAB_SPLITTER_SPLIT_BY",
+)
+```
+
+### 2.3 Embedding Ingestion Pipeline Options
+
+| Option Description | Default Value | CLI Flag | Environment Variable |
+|--------------------|---------------|----------|----------------------|
+| Whether to include a transformation step. | `False` | `--transform` (boolean) | `ILAB_TRANSFORM` |
```
anastasds, line 87: What would some examples of transformations be?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+| The output path of transformed documents (serve as input for the embedding ingestion pipeline). Mandatory when `--transform` is used. |  | `--transform-output` | `ILAB_TRANSFORM_OUTPUT` |
+| How to split the documents. One of `page`, `passage`, `sentence`, `word`, `line` | `word` | `--splitter-split-by` | `ILAB_SPLITTER_SPLIT_BY` |
```
anastasds, line 89: Has there been discussion about the existence of this logic with respect to the docling-based document transformation?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+| Maximum number of units in each split. | `200` | `--splitter-split-length` | `ILAB_SPLITTER_SPLIT_LENGTH` |
+| Number of overlapping units for each split. | `0` | `--splitter-split-overlap` | `ILAB_SPLITTER_SPLIT_OVERLAP` |
+| Minimum number of units per split. | `0` | `--splitter-split-threshold` | `ILAB_SPLITTER_SPLIT_THRESHOLD` |
+| Vector DB implementation, one of: `milvuslite`, **TBD** | `milvuslite` | `--vectordb-type` | `ILAB_VECTORDB_TYPE` |
+| Vector DB service URI. | `./rag-output.db` | `--vectordb-uri` | `ILAB_VECTORDB_URI` |
+| Vector DB connection token. | | `--vectordb-token` | `ILAB_VECTORDB_TOKEN` |
+| Vector DB connection username. | | `--vectordb-username` | `ILAB_VECTORDB_USERNAME` |
+| Vector DB connection password. | | `--vectordb-password` | `ILAB_VECTORDB_PASSWORD` |
+| Name of the embedding model. | `sentence-transformers/all-minilm-l6-v2` | `--model` | `ILAB_EMBEDDING_MODEL_NAME` |
+| Token to download private models. |  | `--model-token` | `ILAB_EMBEDDING_MODEL_TOKEN` |
```
anastasds, line 99: This would introduce model downloading logic in a new place while `ilab model download` [already exists](https://github.com/instructlab/instructlab?tab=readme-ov-file#-download-the-model).

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+|--------------------|---------------|----------|----------------------|
+| Whether to include a transformation step. | `False` | `--transform` (boolean) | `ILAB_TRANSFORM` |
+| The output path of transformed documents (serve as input for the embedding ingestion pipeline). Mandatory when `--transform` is used. |  | `--transform-output` | `ILAB_TRANSFORM_OUTPUT` |
+| How to split the documents. One of `page`, `passage`, `sentence`, `word`, `line` | `word` | `--splitter-split-by` | `ILAB_SPLITTER_SPLIT_BY` |
+| Maximum number of units in each split. | `200` | `--splitter-split-length` | `ILAB_SPLITTER_SPLIT_LENGTH` |
+| Number of overlapping units for each split. | `0` | `--splitter-split-overlap` | `ILAB_SPLITTER_SPLIT_OVERLAP` |
+| Minimum number of units per split. | `0` | `--splitter-split-threshold` | `ILAB_SPLITTER_SPLIT_THRESHOLD` |
+| Vector DB implementation, one of: `milvuslite`, **TBD** | `milvuslite` | `--vectordb-type` | `ILAB_VECTORDB_TYPE` |
+| Vector DB service URI. | `./rag-output.db` | `--vectordb-uri` | `ILAB_VECTORDB_URI` |
+| Vector DB connection token. | | `--vectordb-token` | `ILAB_VECTORDB_TOKEN` |
```
anastasds, line 95: How does this differ from username/password?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+
+| Option Description | Default Value | CLI Flag | Environment Variable |
+|--------------------|---------------|----------|----------------------|
+| Whether to include a transformation step. | `False` | `--transform` (boolean) | `ILAB_TRANSFORM` |
+| The output path of transformed documents (serve as input for the embedding ingestion pipeline). Mandatory when `--transform` is used. |  | `--transform-output` | `ILAB_TRANSFORM_OUTPUT` |
+| How to split the documents. One of `page`, `passage`, `sentence`, `word`, `line` | `word` | `--splitter-split-by` | `ILAB_SPLITTER_SPLIT_BY` |
+| Maximum number of units in each split. | `200` | `--splitter-split-length` | `ILAB_SPLITTER_SPLIT_LENGTH` |
+| Number of overlapping units for each split. | `0` | `--splitter-split-overlap` | `ILAB_SPLITTER_SPLIT_OVERLAP` |
+| Minimum number of units per split. | `0` | `--splitter-split-threshold` | `ILAB_SPLITTER_SPLIT_THRESHOLD` |
+| Vector DB implementation, one of: `milvuslite`, **TBD** | `milvuslite` | `--vectordb-type` | `ILAB_VECTORDB_TYPE` |
```
anastasds, line 93: Many (all?) databases allow multiple document collections. That should probably be a parameter as well.

anastasds:
There are many design decisions being made here that appear to be in a bit of a vacuum and so increase complexity of product usage and configuration while there are opportunities to streamline it instead.

@jwm4 I think we need to dedicate a significant effort to work through these as a group.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+    prompt: |
+      Given the following information, answer the question.
+      Context:
+      {{context}}
+      Question: {{question}}
+      Answer:
+```
+
+### 2.7 References
+* [Haystack-DocumentSplitter](https://github.com/deepset-ai/haystack/blob/f0c3692cf2a86c69de8738d53af925500e8a5126/haystack/components/preprocessors/document_splitter.py#L55)
```
jwm4, line 178: I think probably the Haystack splitter will wind up getting dropped from the solution in favor of something Docling-based.

jwm4 (changes requested):
This is starting to look good to me.  I still have some minor disagreements about technical details (see comments below) but mostly this is feeling like it is on the right track.

docs/cli/ilab-rag-retrieval.md
jwm4: Ah, good point!  I hadn't thought of that.  Using the existing model download sounds better to me.

docs/cli/ilab-rag-retrieval.md
dmartinol: recording link shared on Slack channel

docs/cli/ilab-rag-retrieval.md
dmartinol: Adding one more option
```md
| Vector DB connection token. | `IlabEmbeddings` | `--vectordb-collection-name` | `ILAB_VECTORDB_COLLECTION_NAME` |
```

docs/cli/ilab-rag-retrieval.md
dmartinol: Sounds good to me. 
You mean that the user should first `ilab model download -rp sentence-transformers/all-minilm-l6-v2` and the RAG pipelines (both) would validate that a local download exists for that model before proceeding? 
Or would the RAG pipelines use the ilab download function to download the model locally?

docs/cli/ilab-rag-retrieval.md
dmartinol: You are right: I took `token` from other document store examples in Haystack, but I agree it's better to focus on Milvus only for now. 
WDYT about dropping the authentication part for now and then review the decision once we define the supported stores and verify the available authentication methods?

Milvus seems to offer authentication via username and password, but other stores have a different authn method, or no authn at all (e.g. Chroma).
If we want to be more generic, what about a single `--vectordb-authentication` option where we can put a comma-separated list of the store-specific settings?
E.g.,  for Milvus it would be:
```
ilab data process --rag --vectordb-type milvus --vectordb-uri 'http://localhost:1234' \
--vectordb-authentication 'username=$MILVUS_USER,password=$MILVUS_PASSWORD'
```

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,221 @@
...
+
+### 2.6 RAG Chat Options
+As we stated in [3.1 Working Assumptions](#31-working-assumption), we will introduce new configuration options for the spceific `chat` command,
+but we'll use flags and environment variables for the options that come from the embedding ingestion pipeline command.
+
+| Configuration FQN | Description | Default Value | CLI Flag | Environment Variable |
+|-------------------|-------------|---------------|----------|----------------------|
+| chat.rag.enabled | Enable or disable the RAG pipeline. | `false` | `--rag` (boolean)| `ILAB_CHAT_RAG_ENABLED` |
+| chat.rag.retriever.top_k | The maximum number of documents to retrieve. | `10` | `--retriever-top-k` | `ILAB_CHAT_RAG_RETRIEVER_TOP_K` |
+| chat.rag.prompt | Prompt template for RAG-based queries. | Examples below | `--rag-prompt` | `ILAB_CHAT_RAG_PROMPT` |
```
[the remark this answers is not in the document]
dmartinol: - Would this deserve its own ADR?
- Is it something we should let the user to configure or are you just thinking of where to place an hardcoded prompt?

docs/cli/ilab-rag-retrieval.md
jwm4: You may have seen on Slack that the Docling hybrid chunker is now released.  I looked at the code briefly, and it looks good to me.  More details:
- [Announcement](https://github.com/DS4SD/docling/discussions/548)
- [Documentation](https://ds4sd.github.io/docling/examples/hybrid_chunking/)
- [Code](https://github.com/DS4SD/docling-core/blob/main/docling_core/transforms/chunker/hybrid_chunker.py)

I think this will be a good fit for the RAG chunking because (unlike their older hierarchical chunker), it provides chunks that are constrained to be no bigger than a fixed size for a given tokenizer and tries to make the chunks as big as possible within that size limit and the constraints of the structure.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,229 @@
...
+The proposal is to add a `process`  command to the `data` command group, with an explicit `--rag` flag to trigger
+the executions:
+```
+ilab data process --rag /path/to/docs/folder
+```
+
+The rationale behind this choice is that the `data process` command can support future workflows, making its
+introduction an investment to anticipate other needs.
+
+Since the RAG behavior is the only functionality of this new command, executions without the `--rag` option will result
```
bbrowning, line 36: Just a note that `ilab data process` does not exist yet, so I'd be careful about designing things that layer on top of it until we see when/if that gets implemented in its current form.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,229 @@
...
+in no output for now.
+
+#### Command Purpose
+Generate the embeddings from the documents at */path/to/docs/folder* folder and store them in the
+configured vector database. These embeddings are intended to be used as augmented context in a RAG-based chat pipeline.
+
+#### Assumptions
+The provided documents must be in JSON format according to the InstructLab schema: this is the schema generated
+when transforming knowledge documents with the `ilab data generate` command (see
```
bbrowning, line 45: `ilab data generate` does not output documents in an InstructLab schema, at least not as referenced here. Even once we separate out preprocessing from generation from postprocessing in `ilab data` commands, we may keep `ilab data generate` as it is today for backwards compatibility. I don't think that's decided yet.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,229 @@
...
+#### Assumptions
+The provided documents must be in JSON format according to the InstructLab schema: this is the schema generated
+when transforming knowledge documents with the `ilab data generate` command (see 
+[Getting Started with Knowledge Contributions][1]).
+
+To simplify the execution of the transformation step, we introduce a `--transform` option which also includes the documents
+transformation, leveraging on the `instructlab-sdg` modules. 
+
+### Why We Need It
+This command streamlines the `ilab data generate` pipeline and eliminates the requirement to define a `qna` document,
```
bbrowning, line 52: I'm not aware of any intention to remove the need for qna.yaml from `ilab data generate`. My latest understanding is that we hope to end up with separate commands for preprocessing qna.yaml into data samples, running a generation pipeline, and post-processing generated results into final mixed datasets. `ilab data generate` encompasses all 3 of those stages today, and may continue to. However, we do plan to have some step that starts at input data samples and runs generation pipelines, without a qna.yaml required. It will likely not be called `ilab data generate` but that's still undecided.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,229 @@
...
+For example, the `vectordb-uri` argument can be implemented using the `click` module like this:
+```py
+@click.option(
+    "--vectordb-uri",
+    default='rag-output.db',
+    envvar="ILAB_VECTORDB_URI",
+)
+```
+
+### 2.3 Embedding Ingestion Pipeline Options
```
bbrowning, line 84: I wonder if instead of wiring special support for all this into ilab, we should just consider generating and inserting vector embeddings stages in a data generation pipeline? We're moving to a model where users can supply their own custom pipeline and create their own custom pipeline blocks. So, we should ship (or the user could define) a RAG pipeline that handles turning the data samples into embeddings and storing them in a vector database all as part of our existing pipeline flows, without any code changes in SDG itself?

bbrowning (commented):
Overall I have some concerns about this approach, especially in light of the current changes happening in SDG. I think a lot of this approach is based on where SDG was and not where SDG is going, but this work wouldn't land in SDG until after we've reconciled with the research changes, have the ability to create custom Pipeline Blocks, expect users to create and execute their own Pipelines, and split out data preprocessing from data generation from data postprocessing.

I think the entire approach to generating vector embeddings and populating those in a vector database could probably be handled with the existing (post-reconcile with Research fork) SDG code along with a custom Pipeline Block implementation or two. We don't document how to do this yet, as the code is just landing, but that's our designated extension mechanism to do any random thing you want during a data generation pipeline.

docs/cli/ilab-rag-retrieval.md
anastasds: Perhaps the both most convenient and flexible solution would be to package in a default model but also allow downloading and configuring the use of a different model by doing `ilab model download...` and configuring it to be used.

anastasds:
@bbrowning 

>a custom Pipeline Block implementation or two

What would the purpose of indexing generated data be?

bbrowning:
> @bbrowning
> 
> > a custom Pipeline Block implementation or two
> 
> What would the purpose of indexing generated data be?

I don't mean indexing generated data - I mean using our pipelines concept to run a RAG pipeline that generates embeddings, populates a vector db, whatever you need - as opposed to calling an LLM for inference and data generation. Pipelines take an input dataset, have a sequences of Blocks that get executed in step, with the first block getting each input sample as input, it transforms those samples in some way, outputs samples, and the next block gets those new samples as its input. Today we mostly use this for transforming data in datasets, building prompts and calling LLMs for inference, but you could also use this concept to tokenize text and insert into a vector db. A RAG pipeline just becomes another set of pipelines shipped with the product versus code custom and specific to the RAG use-case, other than perhaps some RAG-specific Blocks we'd like to ship in the product itself.

It may be hard to understand how this all works without understanding the code of SDG including the upcoming changes to it, but we should at least try to use the designed SDG extension points of custom Blocks for part of this I think.

docs/cli/ilab-rag-retrieval.md
dmartinol: Is there any design document that you can share about this initiative?

docs/cli/ilab-rag-retrieval.md
dmartinol: Agree, and there was no intention to `remove the need for qna.yaml from ilab data generate`

docs/cli/ilab-rag-retrieval.md
dmartinol: - Agree on the `InstructLab schema` comment, but these pre-processed artifacts are identified in this way in William's document "WC - RAG Artifacts with RHEL AI (PM perspective)" (I can share a link in DM if needed)
- Of course `ilab data generate` can remain as it is today, this is ouside the purpose of this design document

docs/cli/ilab-rag-retrieval.md
dmartinol: - `ilab data process --rag input` runs the embedding pipeline: fetches pre-processed docs from `input` folder and stores the generated embeddings in the configured vector store.
- `ilab data process --rag --transform --transform-output processed input` runs the transformation pipeline before: fetches user docs from `input` folder, processes them using sdg transformation in `processed` folder, then run the previous pipeline from this folder.

docs/cli/ilab-rag-retrieval.md
dmartinol: I will sync up with ET engineers to adopt the same approach once available.

docs/cli/ilab-rag-retrieval.md
dmartinol: So it's a user responsibility to download it first 👍 
I'll add a note and also introduce a --model-dir option to supply a configurable location for looking for downloaded models

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
jwm4: Yes, I will add this to the list of planned ADRs.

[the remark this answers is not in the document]
franciscojavierarceo: > If there is a way to unify prompt templates throughout InstructLab into one place rather than adding another place to store them, that would probably be ideal.

+1

dmartinol:
@jwm4 @anastasds integrated changes from yesterday's meeting. Should we move it from Draft to Ready?

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
dmartinol: **TL;DR**: Probably doable. Is it the top priority?

- Yes, if we design the solution with the required flexibility, Feast could be plugged as one of the embedding stores  or even retriever, as you said. Something like:
```
ilab data ingest --store-type feast --store-uri 'http://localhost:875/' \
  --store-settings 'feature_name=AAA,feature=BBB' /path/to/processed/files
ilab model chat --store-type feast --store-uri='...' ...
```
- Related functions should be defined in a dedicated package like `instructlab-feast`, considering the (non-trivial) manipulations required to adapt the two environments.
- IMO, it seems a bit excessive to install and configure Feast in the CLI environment where ilab lives. Maybe this is more relevant for a production environment in RHOAI.
- The outcome for time being is that we could try to be a bit more generic in the adopted terminology. E.g. user `store` instead of `vectordb` when we reference parameter names. This can also come in handy when we integrate with embedding APIs rather than using our own embedder. @jwm4 WDYT?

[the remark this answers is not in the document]
franciscojavierarceo: When we get to the RHOAI solution we would benefit from all of the operator, RBAC, integration work we've already done, which is where my view comes from.

If we were to go down that route, ilab data ingest would just be applications of Feast with our ILab specific business logic. I think that'd be a reasonable pattern. 

I do acknowledge though that it could be bloated for what we're doing and Feast RAG is not as strong of a community as the other nor as good of a solution. The big benefit is that Feast is not single-entity backed.

These are all things I think are worth discussing in a doc (or maybe this ADR).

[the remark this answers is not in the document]
dmartinol: For the sake of understanding, could you clarify which services will become `ilab data ingest` and `ilab model chat` when we move to RHOAI?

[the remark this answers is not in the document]
franciscojavierarceo: `ilab data ingest` would in an ideal world would be Feast.

`ilab model chat` I'm not sure. It would require be KServe, _potentially_ Feast (or whatever document serving solution we choose), and whatever chat application/service solution we choose.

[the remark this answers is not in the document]
jwm4: @franciscojavierarceo writes:
> When we get to the RHOAI solution we would benefit from all of the operator, RBAC, integration work we've already done, which is where my view comes from.
> If we were to go down that route, ilab data ingest would just be applications of Feast with our ILab specific business logic. I think that'd be a reasonable pattern.
> I do acknowledge though that it could be bloated for what we're doing and Feast RAG is not as strong of a community as the other nor as good of a solution. The big benefit is that Feast is not single-entity backed.
> These are all things I think are worth discussing in a doc (or maybe this ADR).

These seem like very important topics to me.  I would rather have a separate doc for them rather than try to get it all sorted out in this doc because they also seem very complicated.  FWIW, I think I only understand bits and pieces of the vision here and not the big picture yet, but it does seem like there is a lot of value that this integration could provide for enterprise scale-out.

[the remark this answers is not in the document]
dmartinol: > Can you find other examples too?
For now we don't have many configurations exposed, and `document-store` can also apply to the chat pipeline.

Post-mvp, we could include the `retriever` settings to specify the retriever details:
- `retriever-type`: 
  - `default`: our default retriever implementation using the configured document-store.
  - <another available retriever OOTB>.
  - `api: `a dedicated service exposing well-defined APIs for retrieving embeddings (/query). Includes the document store.
  - `feast`: includes the document store.
Before going into the hell of overthinking, I would wait for the outcome of ET experiments and also carefully evaluate the solution designed by the IBM research team.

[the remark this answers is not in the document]
dmartinol: > `ilab data ingest` would in an ideal world would be Feast.
> 
> `ilab model chat` I'm not sure. It would require be KServe, _potentially_ Feast (or whatever document serving solution we choose), and whatever chat application/service solution we choose.

Agree. That being said, how can this ADR support the request to integrate with Feast?

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,276 @@
...
+```bash
+ilab model download -rp sentence-transformers/all-minilm-l6-v2
+```
+
+If the configured embedding model has not been cached, the command execution will terminate with an error. This requirement applies
+consistently to all new and updated commands.
+
+### 2.2 Document Processing Pipeline
+The proposal is to add a `process`  sub-command to the `data` command group:
+```
```
dmartinol, line 55: Review the command to fetch, by default, knowledge definitions from the taxonomy repo.
`--input` and `--output` flags are optional arguments instead.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,276 @@
...
+
+
+| Option Description | Default Value | CLI Flag | Environment Variable |
+|--------------------|---------------|----------|----------------------|
+| Base directories where models are stored. | `$HOME/.cache/instructlab/models`  | `--model-dir` | `ILAB_MODEL_DIR` |
+| Name of the embedding model. | **TBD** | `--embedding-model` | `ILAB_EMBEDDING_MODEL_NAME` |
+
+### 2.4 Embedding Ingestion Pipeline
+The proposal is to add an `ingest` sub-command to the `data` command group:
+```
```
dmartinol, line 92: Must have the same default like the processing command: the `/path/to/docs/folder` argument is optional and, when missing, the processed artifacts are searched in the datasets folder.

dmartinol:
@jwm4 @anastasds I added a "User Experience Overview" section with diagrams to clarify the workflows we discussed in the last review meeting. Please share your feedback
FYI, @ilan_pinto

dmartinol:
FYI, @jwm4 @anastasds I added a "Design Considerations" section to outline key design requirements.

docs/cli/ilab-rag-retrieval.md
franciscojavierarceo: To be clear, I am not currently convinced that Haystack is the right choice for us. I did not see a thorough evaluation of the other open source frameworks and given that they are a private company, I have concerns. I am very comfortable with Milvus as it is a part of the LFAI foundation.

dmartinol: @franciscojavierarceo you can also comment the related ADR #164

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,307 @@
...
+| Option Description | Default Value | CLI Flag | Environment Variable |
+|--------------------|---------------|----------|----------------------|
+| Vector DB implementation, one of: `milvuslite`, **TBD** | `milvuslite` | `--vectordb-type` | `ILAB_VECTORDB_TYPE` |
+| Vector DB service URI. | `./rag-output.db` | `--vectordb-uri` | `ILAB_VECTORDB_URI` |
+| Vector DB collection name. | `IlabEmbeddings` | `--vectordb-collection-name` | `ILAB_VECTORDB_COLLECTION_NAME` |
+| Base directories where models are stored. | `$HOME/.cache/instructlab/models`  | `--model-dir` | `ILAB_MODEL_DIR` |
+| Name of the embedding model. | **TBD** | `--embedding-model` | `ILAB_EMBEDDING_MODEL_NAME` |
+
+### 2.6 RAG Chat Pipeline Command
+The proposal is to add a `--rag` flag to the `model chat` command, like:
```
anastasds, line 151: Since I asked the clarifying question in a meeting, let's clearly say that if `--rag` is for overwriting file-based configuration and that if the config file has RAG enabled, the user does not need to pass this parameter.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,359 @@
...
+  * Either from knowledge taxonomy or from actual user documents.
+* A new `ilab data` sub-command to generate and ingest embeddings from pre-processed documents into a configured vector store.
+* An option to enhance the chat pipeline by using the stored embeddings to augment the context of conversations, improving relevance and accuracy.
+
+### 1.1 User Experience Overview
+The commands are tailored to support diverse user experiences, all enabling the use of RAG functionality to enrich chat sessions.
+
+### 1.2 Model Training Path
+This flow is designed for users who aim to train their own models and leverage the source documents that support knowledge submissions to enhance the chat context:
+![model-training](../images/rag-model-training.png)
```
jwm4, line 26: I like these images as-is, but they do appear to violate the official dev-docs guidelines on images as [specified here](https://github.com/instructlab/dev-docs/blob/main/README.md#images).   My inclination is to leave them as is since they look good, but if the oversight committee decides to be very strict about this guideline then we might need to redo them.

dmartinol: Images were actually generated from Excalidraw, one of the recommended tool, and I also added a link to edit them for maintenance and sharing purposes: it's in another section below this paragraph, I can move it up if needed. 
Could you clarify what is the violation that you see in the guidelines? (I had already read them before, that's why I used this tool, BTW)

dmartinol: @jwm4 @anastasds please review the "Options to Rebuild Excalidraw Diagrams:" section at the top of the document.

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,372 @@
...
+
+### 2.2 Document Processing Pipeline
+The proposal is to add a `process`  sub-command to the `data` command group.
+
+For the Taxonomy path (no Model Training): 
+```
+ilab data process --output /path/to/processed/folder
+```
+
+For the Plag-and-Play RAG path:
```
jwm4, line 87: "Plag" is misspelled.  It should be "Plug".

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,372 @@
...
+### 1.1 User Experience Overview
+The commands are tailored to support diverse user experiences, all enabling the use of RAG functionality to enrich chat sessions.
+
+### 1.2 Model Training Path
+This flow is designed for users who aim to train their own models and leverage the source documents that support knowledge submissions to enhance the chat context:
+![model-training](./images/rag-model-training.png)
+
+**Note**: documents are processed using `instructlab-sdg` package and are defined using the docling v1 schema.
+
+### 1.3 Tanomony Path (no Training)
```
jwm4, line 34: "Tanomony" should be "Taxonomy"

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,372 @@
...
+│  • /s filepath: save current session to filepath                                                                                  │
+│  • /l filepath: load filepath and start a new session                                                                             │
+│  • /L filepath: load filepath (permanently) and start a new session                                                               │
+│                                                                                                                                   │
+│ Press Alt (or Meta) and Enter or Esc Enter to end multiline input.                                                                │
+╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
+```
+
+### 2.8 RAG Chat Options
+As we stated in [2.1 Working Assumptions](#21-working-assumption), we will introduce new configuration options for the spceific `chat` command,
```
jwm4, line 268: "spceific" should be "specific"

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,372 @@
...
+but we'll use flags and environment variables for the options that come from the embedding ingestion pipeline command.
+
+| Configuration FQN | Description | Default Value | CLI Flag | Environment Variable |
+|-------------------|-------------|---------------|----------|----------------------|
+| chat.rag.enabled | Enable or disable the RAG pipeline. | `false` | `--rag` (boolean)| `ILAB_RAG` |
+| chat.rag.retriever.top_k | The maximum number of documents to retrieve. | `10` | `--retriever-top-k` | `ILAB_RETRIEVER_TOP_K` |
+| | Document store implementation, one of: `milvuslite`, **TBD** | `milvuslite` | `--document-store-type` | `ILAB_DOCUMENT_STORE_TYPE` |
+| | Document storeservice URI. | `./embeddings.db` | `--document-store-uri` | `ILAB_DOCUMENT_STORE_URI` |
```
jwm4, line 276: Should "storeservice" be "store service"?  Or is this one word for a reason?

jwm4 (changes requested):
Minor spelling issues

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,437 @@
...
+```bash
+ilab model download -rp ibm-granite/granite-embedding-125m-english
+```
+
+If the configured embedding model has not been cached, the command execution will terminate with an error. This requirement applies
+consistently to all new and updated commands.
+
+### 2.2 Document Processing Pipeline
+
+The proposal is to add a `process` sub-command to the `rag` command group.
```
jwm4, line 99: `process` should be `convert` now

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,437 @@
...
+|--------------------|---------------|----------|----------------------|
+| Document store implementation, one of: `milvuslite`, **TBD** | `milvuslite` | `--document-store-type` | `ILAB_DOCUMENT_STORE_TYPE` |
+| Document store service URI. | `./embeddings.db` | `--document-store-uri` | `ILAB_DOCUMENT_STORE_URI` |
+| Document store collection name. | `IlabEmbeddings` | `--document-store-collection-name` | `ILAB_DOCUMENT_STORE_COLLECTION_NAME` |
+| Base directories where models are stored. | `$HOME/.cache/instructlab/models`  | `--retriever-embedder-model-dir` | `ILAB_EMBEDDER_MODEL_DIR` |
+| Name of the embedding model. | **TBD** | `--retriever-embedder-model-name` | `ILAB_EMBEDDER_MODEL_NAME` |
+
+### 2.6 RAG Chat Pipeline Command
+
+The proposal is to add a `chat.rag.enable` configuration (or the equivalent `--rag` flag) to the `model chat` command, like:
```
jwm4, line 251: This is also gated on ILAB_DEV_PREVIEW, right?  If so, let's make that explicit here.

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,437 @@
...
+Question: 
+{user_query}
+Answer:
+```
+
+Future extensions should align prompt management with the existing InstructLab design.
+
+### 2.7 RAG Chat Commands
+
+The `/r` command may be added to the `ilab model chat` command to dynamically toggle the execution of the RAG pipeline.
```
jwm4, line 286: Again, let's qualify this by saying that `/r` will only appear in this help list and only work if ILAB_DEV_PREVIEW is true.

jwm4 (changes requested):
This is mostly looking good to me.  I am requesting a few minor changes.

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,438 @@
...
+  --taxonomy-path PATH  Directory where taxonomy is stored and accessed from.
+  --taxonomy-base TEXT  Branch of taxonomy used to calculate diff against.
+  --output DIRECTORY    Directory where processed docs are stored.
+  --help                Show this message and exit.
+```
+
+| Option Description | Default Value | CLI Flag | Environment Variable |
+|--------------------|---------------|----------|----------------------|
+| Location folder of user documents. In case it's missing, the taxonomy is navigated to look for updated knowledge documents.|  | `--input` | `ILAB_PROCESS_INPUT` |
+| Location folder of processed documents. |  | `--ouput` | `ILAB_PROCESS_OUTPUT` |
```
jwm4, line 161: @anastasds , should ILAB_PROCESS_INPUT and ILAB_PROCESS_OUTPUT be ILAB_CONVERT_INPUT and ILAB_CONVERT_OUTPUT, since process was renamed to convert?

anastasds: Yes, missed that, thanks - submitted fix in https://github.com/dmartinol/dev-docs/pull/4

dmartinol: Merged, thanks!

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,438 @@
...
+* Importing the provided images in Excalidraw
+* Importing the provided images in Excalidraw
+
+## 1. Introduction
+
+This document proposes enhancements to the `ilab` CLI to support workflows utilizing Retrieval-Augmented Generation
+(RAG) artifacts within `InstructLab`. The proposed changes introduce new commands and options for the embedding ingestion
+and RAG-based chat pipelines:
+
+* A new `ilab rag` command group, feature gated behind a `ILAB_DEV_PREVIEW` environment variable.
```
nathan-weinberg, line 27: What's the point of the feature gate?

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,438 @@
...
+* A new `ilab rag` sub-command  group to process customer documentation.
```
nathan-weinberg, line 28:
```suggestion
* A new `ilab rag` sub-command group to process user documentation.
```

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,438 @@
...
+```bash
+ilab model download -rp ibm-granite/granite-embedding-125m-english
+```
+
+If the configured embedding model has not been cached, the command execution will terminate with an error. This requirement applies
+consistently to all new and updated commands.
+
+### 2.2 Document Processing Pipeline
+
+The proposal is to add a `convert` sub-command to the `rag` command group.
```
nathan-weinberg, line 100: I will note there is an existing `ilab model convert` command - I think we can still use `convert` here, but if another term works for this purpose that may be simpler for users

nathan-weinberg:
@dmartinol Please squash commits before this is merged, TIA

docs/rag/ilab-rag-retrieval.md
jwm4: @cdoern and @bbrowning were concerned that if this were released as dev preview without a gate that it would set expectations that this command would continue to exist, but in reality we haven't really converged on a long term CLI for this functionality.

nathan-weinberg: Gotcha - personally I would simply prefer some kind of user alert (i.e. something like "NOTE: This is an experimental command at this time - once fully supported this warning will go away) over a env var-based feature gate - but if there was a previous convo about this I won't muck up the works, I don't feel _that_ strongly about it :smile:

anastasds: Honestly, going in and trying to implement this, it seems a lot simpler to put a warning on it. Especially since some of the new options are for existing command groups, e.g. `chat.rag.enabled`.

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,438 @@
...
+InstructLab technology stack.
+
+#### Ingestion-Usage
+
+The generated embeddings can later be retrieved from a vector database and converted to text, enriching the
+context for RAG-based chat pipelines.
+
+### 2.5 Embedding Ingestion Pipeline Options
+
+**Note**: The `--help` option will be aware of the `rag` command group only if `ILAB_DEV_PREVIEW` environment variable is set to `true`.
```
cdoern, line 217: note the env variable name here as well

docs/rag/ilab-rag-retrieval.md
cdoern: +1

cdoern (commented):
generally looks good! just a few comments on env var names and values.

docs/rag/ilab-rag-retrieval.md
jwm4: I would also prefer a warning, but @cdoern and @bbrowning seemed pretty firm in their insistence on a feature gate.

anastasds: After an offline discussion yesterday, where we landed is that experimental options would cause the application to simply exit with an error message if used without the dev flag being set.

docs/rag/ilab-rag-retrieval.md
anastasds:
```suggestion
**Note**: The `--help` option will be aware of the `rag` command group only if `ILAB_ENABLE_EXPERIMENTAL` environment variable is set to `true`.
```

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,438 @@
...
+**Note**: documents are processed using `instructlab-sdg` package and are defined using the docling v1 schema.
+
+### 1.3 Taxonomy Path (no Training)
+
+This flow is for users who have defined taxonomy knowledge but prefer not to train their own models. Instead, they aim to generate RAG artifacts from source documents to enhance the chat context:
+![taxonomy-no-training](./images/rag-taxonomy-no-training.png)
+
+**Note**: documents are processed using `docling.DocumentConverter` and are defined using the docling v2 schema.
+
+### 1.4 Plug-and-Play RAG Path
```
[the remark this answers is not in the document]
dmartinol: I will move it on top of the list, let me know if this is enough. IMO this seems the "main" use case for demo purposes, but in the long run Model Training one may take priority for production cases (together with fine-tuned embedding model, probably)

jwm4 (approved):
The concerns I raised in my last review appear to be resolved, so I am happy for this to merge now.

jwm4 (approved):

jwm4 (approved):
I've reviewed the latest round of changes and this still looks good to me.  I think it is ready for final oversight and merging.  cc: @instructlab/oversight-committee

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,455 @@
...
+ilab rag ingest --build-image --image-name=docker.io/user/my_rag_artifacts:1.0
+```
+
+Then serve it and use it in a chat session:
+
+```bash
+ilab serve --rag-embeddings --image-name=docker.io/user/my_rag_artifacts:1.0 --port 8123
+ilab model chat --rag --retriever-type api --retriever-uri http://localhost:8123
+```
+
```
reidliu41, line 447: There are many options, but during issues handling, it was found that users either don’t know about or can’t use them correctly, especially when some options need to be used together, like `--retriever-type` and `--retriever-uri`. Only the developers who wrote it might know about this. To make it easier for users to understand, can we emphasize this in the help message or is there a better way to handle it?

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,455 @@
...
+```bash
+ilab model chat --rag --query-expansion --query-expansion-prompt="$QUERY_EXPANSION_PROMPT" --query-expansion-num-of-queries=5
+```
+
+* Using retrieval strategy:
+
+```bash
+ilab model chat --rag --retrieval-strategy query-expansion --retrieval-strategy-options="prompt=$QUERY_EXPANSION_PROMPT;num_of_queries=5"
+```
```
reidliu41, line 429: This kind of option can make usage more complicated, especially when multiple parameters need to be configured together?

reidliu41 (changes requested):

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,359 @@
...
+
+### 4.2 Advanced RAG retrieval steps
+- [Ranking retriever's result][ranking]: 
+```bash
+ilab model chat --rag --ranking --ranking-top-k=5 --ranking-model=cross-encoder/ms-marco-MiniLM-L-12-v2
+```
+- [Query expansion][expansion]:
+```bash
+ilab model chat --rag --query-expansion --query-expansion-prompt="$QUERY_EXPANSION_PROMPT" --query-expansion-num-of-queries=5
+```
```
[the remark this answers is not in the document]
anastasds: There is no prompt management in the product aside from system prompts, so @dmartinol perhaps best to mark this (along with anything related to query expansion) as a notional future direction that does not have a committed or complete design in this document?

docs/rag/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,372 @@
...
+)
+```
+
+#### Local embedding models
+The embedding model used to generate text embeddings must be downloaded locally before executing the pipeline.
+
+For example, this command can be used to download the `sentence-transformers/all-minilm-l6-v2` model to the local models cache:
+```bash
+ilab model download -rp sentence-transformers/all-minilm-l6-v2
+```
```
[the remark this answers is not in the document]
jwm4: You don't need to include `--hf-token` if the `HF_TOKEN` environment variable is set to your Huggingface token.

docs/rag/ilab-rag-retrieval.md
jwm4: I agree.  I think more consideration is needed for how we will enable retrieval strategy configuration.  I think that should be addressed in a future document though; for this document, I would recommend just adding a note indicating that the entire "future enhancements" section is just some preliminary thoughts on the sorts of things we might add in the future.

docs/cli/ilab-rag-retrieval.md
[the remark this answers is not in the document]
dmartinol: https://github.com/instructlab/instructlab/issues/2870

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,359 @@
...
+| Document store collection name. | `IlabEmbeddings` | `--document-store-collection-name` | `ILAB_DOCUMENT_STORE_COLLECTION_NAME` |
+| Base directories where models are stored. | `$HOME/.cache/instructlab/models`  | `--retriever-embedder-model-dir` | `ILAB_EMBEDDER_MODEL_DIR` |
+| Name of the embedding model. | **TBD** | `--retriever-embedder-model-name` | `ILAB_EMBEDDER_MODEL_NAME` |
+
+### 2.6 RAG Chat Pipeline Command
+The proposal is to add a `chat.rag.enable` configuration (or the equivalent `--rag` flag) to the `model chat` command, like:
+```
+ilab model chat --rag
+```
+
```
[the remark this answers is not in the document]
dmartinol: > I think it would be better to have a list or check option to check what documents or collection name it have currently if it can use multiple.

Makes sense, but probably such command extensions may be discussed after an initial rollout, WDYT?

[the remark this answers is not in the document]
dmartinol: > May I know one document one collection name? Or one collection name multiple documents?

The ingestion logic puts all docs in the configured collection name. Of course we can run it multiple times, each one to a different collection. 
At inference time (chat), there's only the option to configure the collection name at startup, but it might be useful to allow for changing it dynamically along the conversation.

docs/rag/ilab-rag-retrieval.md
dmartinol: > I would recommend just adding a note indicating that the entire "future enhancements" section is just some preliminary thoughts on the sorts of things we might add in the future.

Agree

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,359 @@
...
+```bash
+% ilab data process --help
+Usage: ilab data process [OPTIONS] OUTPUT_DIR
+
+  The document processing pipeline
+
+Options:
+  --input DIRECTORY  The folder with user documents to process.
+  --help             Show this message and exit.```
+```
```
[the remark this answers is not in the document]
dmartinol: makes sense for future enhancements.

docs/cli/ilab-rag-retrieval.md
```diff
@@ -0,0 +1,359 @@
...
+
+#### Working Assumption
+The documents at the specified path have already been processed using the `data process` command or an equivalent method
+(see [Getting Started with Knowledge Contributions][ilab-knowledge]).
+
+#### Command Purpose
+Generate the embeddings from the pre-processed documents.
+* In case of Model Training path, the documents are located in the location specified by the `generate.output_dir` configuration key
+  (e.g. `_HOME_/.local/share/instructlab/datasets`).
```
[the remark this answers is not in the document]
dmartinol: The doc also states `In particular, only the latest folder with name starting by `documents-` will be explored`.
Future enhancements can provide support to point to a specific document/dataset.

docs/rag/images/rag-ingestion-and-chat.excalidraw
```diff
@@ -0,0 +1,12599 @@
+{
```
danmcp, line 1: Apologies for the dummy question as I am not an Excalidraw user (yet) but is this necessary?  I was thinking embedded scene option meant the images themselves would be editable in excalidraw after the fact and this config wouldn't be necessary.  It makes a mess our of the commit/repo.

danmcp (changes requested):
I think it would be good to see:

- [x] An answer to my newbie excalidraw question and some resolution on that really large file
- [ ] An approval from (or delegate approval) from @bbrowning and @cdoern since they lead affected areas and have given feedback
- [ ] As @nathan-weinberg suggested the PR needs to be squashed of intermediate commits

docs/rag/images/rag-ingestion-and-chat.excalidraw
dmartinol: No user is required. Just open the individual images in Excalidraw.
I removed the Excalidraw related section if it adds confusion and re-added PNGs with scenes (they got missed in some update). Hope this is clear now.
Squashing asap.

bbrowning:
Regarding my approval, and to @danmcp's point, on the surface it doesn't look like anything here expects or requires changes in SDG today. Am I reading the dev doc correctly? Things look entirely self-contained with the new `ilab rag` set of commands, and future work to reuse the same document ingestion code paths between SDG and the RAG work is deferred and out of scope of this dev doc?

The proposal seems reasonable as-written, given that it's a separate command set and gated behind a feature flag. However, it also seems to mostly or entirely impact the instructlab/instructlab repository with the new commands and code to support the experimental workflow, so I'd defer more towards the most active maintainers there to approve this, with @cdoern being a good person to wrangle those.

dmartinol:
> Regarding my approval, and to @danmcp's point, on the surface it doesn't look like anything here expects or requires changes in SDG today. Am I reading the dev doc correctly?

You are right. No changes expected on SDG package.

> future work to reuse the same document ingestion code paths between SDG and the RAG work is deferred and out of scope of this dev doc?

Correct: as you said, this is out of scope.

docs/cli/ilab-rag-retrieval.md
bbrowning: I'm resolving this thread, as how this work reconciles with SDG to ensure we consolidate on a single code path to ingest and convert documents between the RAG work and SDG is out of scope of this dev doc.

nathan-weinberg:
If it's good with @danmcp I'm happy to merge this

PR #163, opened 2024-12-06
Add ADR for what repository will contain RAG

jwm4:


docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Non-goals
+
+Evaluation of RAG will be addressed in one or more other development documents.  That topic is out of scope for this document.
+
+## Decision
+
+- There will be a new repository for RAG.
+- It will be located at https://github.com/instructlab/retrieval-augmented-generation
+- By mid-January, it will be available and working but not integrated with InstructLab.
```
franciscojavierarceo, line 34: is there some rationale behind these two dates? just to hold ourselves accountable?

jwm4: The dates are driven mainly by availability of people to do the work.  The mid-March elements of the plan will require a lot of coordination with the maintainers of the core instructlab/instructlab repo.  Those people are mostly already committed to other activities through mid-January.

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+Provide a built-in alternative for users who do not want to build their own RAG.  Do not break the existing capability at https://github.com/instructlab/rag .
+
+## Non-goals
+
+Evaluation of RAG will be addressed in one or more other development documents.  That topic is out of scope for this document.
+
+## Decision
+
+- There will be a new repository for RAG.
+- It will be located at https://github.com/instructlab/retrieval-augmented-generation
```
anastasds, line 33: How about `instructlab/retrieve`?

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+We want a retrieval-augmented generation (RAG) capability that provides outstanding results with minimal effort, is seamlessly integrated with InstructLab, and is also general enough to be used in other applications as well.
+
+## Why
+
+Many InstructLab users want to train a model and then use it to RAG.  Often they build something simple themselves for this purpose.  Two problems with this approach:
+
+- Building their own RAG is extra work.
+- Users who are not experts on RAG might not build a RAG that provides outstanding results.
+
+There is a very simple RAG capability at https://github.com/instructlab/rag .  It is not tightly integrated with InstructLab and it does not use any advanced RAG capabilities.  However, it might have some existing users so we can't just unilaterally delete it or replace it with something radically different.
```
nathan-weinberg: I would challenge this assumption - I'm seeing 3 commits from 3 months ago by @aldopareja, I severely doubt there are any users here, unless someone has evidence otherwise? https://github.com/instructlab/rag/commits/main/

IMO, it makes more sense to use an existing, well-named repository than to create a second one.

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+
+Evaluation of RAG will be addressed in one or more other development documents.  That topic is out of scope for this document.
+
+## Decision
+
+- There will be a new repository for RAG.
+- It will be located at <https://github.com/instructlab/retrieve>
+- By mid-January, it will be available and working but not integrated with InstructLab.
+- By mid-March, it will be integrated with InstructLab with the new repository being invoked by the core repository and maybe also by the SDG repository.
+- Eventually, it will be integrated with InstructLab with the new repository being invoked only by the core repository.
```
khaledsulayman, line 36: I'm not sure I agree with necessarily assigning dates to these efforts from now. I think it does make sense to split up these efforts into phases, and then we can internally scope out work for the desired releases.

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+  - Pro: This has the advantage of not adding any new dependencies.
+  - Pro: The document processing is already in SDG and chat functionality is already in core so this would require the fewest code changes.
+  - Con: Splitting the RAG functionality across multiple repositories makes it more complicated to reuse in other applications outside of InstructLab.
+  - Con: Many things we will want to do to add advanced functionality to make RAG more effective will require changes to both indexing and run-time RAG.  If those components are split across multiple repositories, that will make delivering such changes more complicated.
+- Put the indexing and run-time RAG code in <https://github.com/instructlab/instructlab> (core)
+  - Pro: This has the advantage of not adding any new dependencies.
+  - Pro: However, since the existing document processing is in SDG, the flow for indexing for RAG would be a bit complicated (i.e., it starts with a CLI call handled by the core repo then goes to SDG for some of the document processing and then back to the core for vector database indexing).   That drawback will be eliminated if/when the document processing moves into the core repository.
```
khaledsulayman, line 71: not sure it's an issue at all that it's making a call to SDG since the `instructlab/instructlab` dependencies contain SDG anyway. We don't expect users to be using `ilab` without SDG

khaledsulayman: however, I do agree with the below cons and don't think it's best to store any of this in `instructlab/instructlab` if we want to advertise this as 'our RAG solution'

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+- Put the indexing and run-time RAG code in <https://github.com/instructlab/rag> AND keep the existing RAG functionality in that repository intact.
+  - Pro: It already exists.
+  - Pro: It avoids the confusion of having two different RAG repositories in <https://github.com/instructlab/>.
+  - Con: It creates the confusion of having two different RAG solutions in the same repository.  I guess we could mitigate that with developer documentation and marking legacy stuff as "deprecated".
+- Put the indexing and run-time RAG code in <https://github.com/instructlab/rag> AND eliminate the existing RAG functionality in that repository.
+  - Pro: It already exists.
+  - Pro: It avoids the confusion of having two different RAG repositories in <https://github.com/instructlab/>.
+  - Pro: It avoids the confusion of having two different RAG solutions since we'd be eliminating the old one.
+  - Con: That stuff must be there for a reason.  Maybe someone is using it.  It is not totally clear how we would even find out.  If someone can run this down maybe we can consider this option.
```
khaledsulayman, line 98: I have to agree with the previous discussions that I think it makes sense to put everything in `instructlab/rag`. Hopefully, if any functionality is needed from what currently exists, we should be able to instrument a full-fledged RAG solution to be able to support that.

Also, more of a nit, I like the name `RAG` more for that repo. It keeps it in line with `SDG` and also just easier to say 😂

dmartinol: @jwm4 have you considered renaming the existing repo, which has no dependencies on `instructlab` artifacts, to something like `rag-experiment` and instead use this name for the new `instructlab-rag` package?

docs/retrieval-augmented-generation/rag-repo.md
franciscojavierarceo: +1.

jwm4: FWIW, I would also prefer this if we can make the internal politics of it work somehow.

docs/retrieval-augmented-generation/rag-repo.md
jwm4: I think the dates are important, but I am fine with dropping them from this document and using other venues to get aligned around these dates.  I will do that.

docs/retrieval-augmented-generation/rag-repo.md
jwm4: > not sure it's an issue at all that it's making a call to SDG since the instructlab/instructlab dependencies contain SDG anyway. We don't expect users to be using ilab without SDG

The issue is not the dependencies, it is the flow -- if document processing starts in core and then moves to SDG and then moves back to core and then moves to the RAG repo, that flow will be difficult to understand and maintain.  This should be labeled as "con" , not "pro".  I will fix that and add some more text to clarify what my concern is.

docs/retrieval-augmented-generation/rag-repo.md
jwm4: > @jwm4 have you considered renaming the existing repo, which has no dependencies on instructlab artifacts, to something like rag-experiment and instead use this name for the new instructlab-rag package?

That's being considered.  Another item being considered is moving the existing content in that repo into a branch and then putting the new stuff in main.  If we can get all the critical stakeholders on board then either of those would be fine with me.

jwm4: > Also, more of a nit, I like the name RAG more for that repo. It keeps it in line with SDG

I would also be fine with `RAG` instead of `rag` -- but definitely would not want _both_ in the game GH org.  So I am fine with renaming if that's the consensus of the reviewers.   If we get permission to use the existing repo, let's see if we can get a consensus around renaming it or not.

jaideepr97: >That's being considered. Another item being considered is moving the existing content in that repo into a branch and then putting the new stuff in main. If we can get all the critical stakeholders on board then either of those would be fine with me.

+1 to moving it to a branch and not introducing a new repo
we may want to communicate this sufficiently to the community (slack, email to dev@instructlab.ai etc) to give people a heads up before doing that 

alternatively we could also start developing the new approach in a branch before promoting it to main

docs/retrieval-augmented-generation/rag-repo.md
akashgit: not sure i know about the internal politics that are blocking this somehow but on behalf of Aldo, I can safely say, please feel free to use it as you see fit :)

docs/retrieval-augmented-generation/rag-repo.md
anastasds: Right now it looks like this repository will not be concerned with the "G" part in "RAG" - namely, it's to provide retrieval functionality for use in InstructLab. Unless I'm missing something on that point, I advocate for sidestepping this problem and simply creating the new `instructlab/retrieve` repository to work in.

jwm4: No, the plan I was trying to communicate in this document was to include the generation too.  I will update to make this clearer.

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+  - Con: If we put it out there as its own open source project, it needs a name.  It is a lot of work to come up with a good name and there will be a lot of stakeholders with an interest in the name that comes up.
+- Keep the indexing and run-time RAG code in <https://github.com/redhat-et/PaRAGon> which is an emerging technologies prototype for this work.
+  - Mostly the same pros and cons as putting it in a new repo outside InstructLab plus the following:
+  - Pro: A prototype for the code we want is already there.
+  - Pro: It already has its own distinctive name (PaRAGon).
+  - Con: The existing repository has its own simple command-line interface which is useful for the prototype but we don't want it in the capability we release because too many command-line interfaces will confuse users.
+  - Con: The name PaRAGon seems fine to me, but probably more stakeholders need to weigh in on what a name would be.
+  - Con: The `redhat-et` label suggests that this is something "owned" by Red Hat which makes sense for the prototype but not so much for something we want a community to own in the long run.
```
nathan-weinberg, line 89: perhaps we can simply move the repo into the InstructLab org?

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+- Put the indexing and run-time RAG code in <https://github.com/instructlab/rag> AND keep the existing RAG functionality in that repository intact.
+  - Pro: It already exists.
+  - Pro: It avoids the confusion of having two different RAG repositories in <https://github.com/instructlab/>.
+  - Con: It creates the confusion of having two different RAG solutions in the same repository.  I guess we could mitigate that with developer documentation and marking legacy stuff as "deprecated".
```
nathan-weinberg, line 93: We can eliminate this con as that solution does not need to be preserved

docs/retrieval-augmented-generation/rag-repo.md
nathan-weinberg: This con can be eliminated per Akash

anastasds:
I think I am settling on advocating on a prototyping phase working within a submodule in the `instructlab` repo to figure out things that are shared concerns across the product like the ones I listed above, and explicitly intend to extract something independently deployable into its own repo later.

It can be very difficult to draw component boundaries from the outset - things like config, prompt template management, model interactions, model management, and more will need to be figured out across both retrieval and inference
if we start with a submodule in the core repo, highly likely that we can extract it into its own repo in a cleaner way later
as to the code management part.

What I want to avoid is having two places where there is config management, two places where prompt templates are defined, two places where there is model management (since we'll have embedding models too), and so on.

A directory level `CODEOWNERS` in the `instructlab` repo should give us the autonomy needed during the design exploration phase of things.

docs/retrieval-augmented-generation/rag-repo.md
jwm4: There is not universal agreement around that point, but I will at least refine the text here.

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+
+- The capabilities in `retrieve` will continue to be improved with even more advanced functionality.  (Details TBD)
+- Both indexing and run-time RAG will be implemented by calls that go directly from the core repo to `retrieve`.  The SDG repository will not directly invoke the retrieve repository since the document processing will all be done in the core repository and the outputs of that document processing will be consumed by both `retrieve` and SDG.
+
+## Alternatives
+
+- Put the indexing code in <https://github.com/instructlab/sdg> (SDG) and the run-time RAG code in <https://github.com/instructlab/instructlab> (core)
+  - Pro: This has the advantage of not adding any new dependencies.
+  - Pro: The document processing is already in SDG and chat functionality is already in core so this would require the fewest code changes.
+  - Con: Splitting the RAG functionality across multiple repositories makes it more complicated to reuse in other applications outside of InstructLab.
```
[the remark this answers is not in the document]
jwm4: +1

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+We want a retrieval-augmented generation (RAG) capability that provides outstanding results with minimal effort, is seamlessly integrated with InstructLab, and is also general enough to be used in other applications as well.
+
+## Why
+
+Many InstructLab users want to train a model and then use it to RAG.  Often they build something simple themselves for this purpose.  Two problems with this approach:
+
+- Building their own RAG is extra work.
+- Users who are not experts on RAG might not build a RAG that provides outstanding results.
+
+There is a very simple RAG capability at <https://github.com/instructlab/rag> .  It is not tightly integrated with InstructLab and it does not use any advanced RAG capabilities.  However, it might have some existing users so we can't just unilaterally delete it or replace it with something radically different.
```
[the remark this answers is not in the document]
jwm4: Will do, but I do intend to leave the point here since we don't have buy in from all the stakeholders yet -- I will just soften it.

docs/retrieval-augmented-generation/rag-repo.md
cooktheryan: @nathan-weinberg any thoughts to allowing https://github.com/redhat-et/paragon to exist and we generate a fork of it. I 100% think the work should live in Instructlab but it would be nice for us to have an upstream as we have additional goals we would like to see with this work.

Also, we could flip it and bring the repo to instructlab and fork it from our side. Vice versa but you see where I am going. I think the key thing that we would want to ensure we have in the end is the git history and the existing work

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+  - Con: Putting the RAG functionality in the core repository requires any application that wants to use this functionality to bring in the entire core which then brings in all of the libraries it depend on, so this becomes an enormous dependency.  This would discourage reuse in other applications.
+  - Con: Building a great RAG capability is a long-term grand challenge that will require a lot of dedicated investment.  That makes it a poor fit for a repository that also has a lot of existing responsibilities.
+- Start by putting the code into existing InstructLab repositories (either of the above options) and then split if off into its own repository later.
+  - Pro: Gets us integrated into InstructLab sooner.
+  - Con: Adds extra work to the second phase where we have to split it off into its own repository.
+  - Con: There is a risk that we never get around to splitting it off and we wind up stuck with the cons of being jammed in to other components indefinitely.
+- Put the indexing and run-time RAG code in a new repo outside <https://github.com/instructlab/>.
```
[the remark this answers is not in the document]
jwm4: We're not doing it for *now*, so I don't think we need to have that debate now.  If/when the time comes, we'll have a separate ADR to hash it out.

docs/retrieval-augmented-generation/rag-repo.md
```diff
@@ -0,0 +1,110 @@
...
+  - Con: Putting the RAG functionality in the core repository requires any application that wants to use this functionality to bring in the entire core which then brings in all of the libraries it depend on, so this becomes an enormous dependency.  This would discourage reuse in other applications.
+  - Con: Building a great RAG capability is a long-term grand challenge that will require a lot of dedicated investment.  That makes it a poor fit for a repository that also has a lot of existing responsibilities.
+- Start by putting the code into existing InstructLab repositories (either of the above options) and then split if off into its own repository later.
+  - Pro: Gets us integrated into InstructLab sooner.
+  - Con: Adds extra work to the second phase where we have to split it off into its own repository.
+  - Con: There is a risk that we never get around to splitting it off and we wind up stuck with the cons of being jammed in to other components indefinitely.
```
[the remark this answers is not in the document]
jwm4: Agreed, but we need to move forward *now* and don't have more time to figure out the long term vision.

anastasds (approved):

docs/retrieval-augmented-generation/rag-repo.md
nathan-weinberg: Sounds good! Let's make sure to keep the ET folks in the loop

docs/retrieval-augmented-generation/rag-repo.md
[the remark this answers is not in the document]
nathan-weinberg: Sure - ultimately I won't be responsible with a potential move so fine with that if y'all are - hopefully we find that the existing code going into Core can be kept and won't need to be moved at all :smile:

nathan-weinberg (approved):
This looks good to me - @anastasds @jwm4 appreciate all the work y'all put into this! - I'd like @cdoern to approve as well but going ahead and giving this my sign-off

khaledsulayman (approved):

PR #164, opened 2024-12-09
framework to be used by RAG for pipeline

cooktheryan:


docs/retrieval-augmented-generation/rag-framework.md
```diff
@@ -0,0 +1,72 @@
...
+  - Limited flexibility and customization options as compared to the alternatives
+  - Limited out-of-the-box support for building complex, multi-component pipelines as compared to Langchain and Haystack
+  - Documentation is less well-maintained and more difficult to use as compared to the alternatives
+- RAGFlow
+  - Limited support for many of the mainstream vector DB providers
+  - Limited scalability as compared to the alternatives
+
+## Risks
+
+N/A
```
dhellmann, line 68: Bringing in a new dependency always introduces risks related to building that package (and its dependencies) downstream, as well as license compliance, long term supportability, etc. Those aren't unique to this package, but it would be good to see them addressed here anyway to show that a thorough vetting was done of the package, and its dependencies.

For example, is any part of the haystack-ai dependency list GPL or AGPL? Is its license compatible with the Instructlab license?

Does it, and all of its dependencies, appear to be actively maintained by more than one maintainer?

Are those maintainers responsive to requests to review PRs, especially for CVEs?

Is the project following good practices for releasing, including tagging releases, publishing source packages, and using trusted publishing automation?

wattsteve: @cooktheryan this type of analysis is fairly routine for OSPO. I'd recommend reaching out to @lhawthorn

erikerlandson:
There is a lot of emphasis given to haystack being focused on RAG. Does it also have a good story for supporting more generalized agentic patterns?

docs/retrieval-augmented-generation/rag-framework.md
ilya-kolchinsky: Thanks for your comments.

None of Haystack's dependencies (jinja2, lazy-imports, more-itertools, networkx, numpy, openai, pandas, posthog, python-dateutil, pyyaml, requests, tenacity, tqdm, typing-extensions) are GPL/AGPL. All of them are Apache, MIT, BSD or PSFL.

Haystack is licensed under Apache 2.0. What else makes it compatible or non-compatible with InstructLab?

Haystack is very [actively](https://github.com/deepset-ai/haystack/pulse/monthly) [maintained](https://github.com/deepset-ai/haystack/issues?q=is%3Aissue+is%3Aclosed) and [supported](https://github.com/deepset-ai/haystack/discussions). As for active maintenance of the dependencies I didn't check them all but most are very well-known and widely used libraries. Moreover, many of them are also on the dependency list of the alternatives to Haystack.

Haystack [releases tagged versions](https://github.com/deepset-ai/haystack/releases). They do not publish to Github Packages.
Could you please elaborate what is considered "trusted publishing automation" and whether there are any additional releasing practices we need to verify?

ilya-kolchinsky:
> There is a lot of emphasis given to haystack being focused on RAG. Does it also have a good story for supporting more generalized agentic patterns?

@erikerlandson, well, since Haystack supports building arbitrary graphs of operators (similarly to Langgraph), it [can potentially implement](https://docs.haystack.deepset.ai/v1.22/docs/agent) many of the agentic patterns. However, as you say, the main focus is on RAG. Since the decision here solely considers our RAG implementation, features and capabilities not directly related to RAG should mostly be out of scope.

docs/retrieval-augmented-generation/rag-framework.md
dhellmann: That would be good information to include in the body of the document.

For trusted publisher, I mean https://docs.pypi.org/trusted-publishers/. We've recently seen an instance of a package in the AI ecosystem where that trusted publishing was set up incorrectly, which led to a third-party having the ability to not only trigger releases but to inject malicious code into _multiple versions of source packages_.  https://github.com/ultralytics/ultralytics/issues/18027 for reference.

Downstream, we're building from source packages, but if we can't trust those source packages we would not want to consume the dependency.

erikerlandson:
I believe that agentic should not be *totally* out of scope; mainly because soon enough the focus is going to shift to agentic, and it would be undesirable (IMO) to be in a position of wanting to add a separate framework for agentic, or having to switch to a new framework altogether.

ilya-kolchinsky:
@erikerlandson, I agree. Since Haystack provides the functionality [to build arbitrarily complex pipelines](https://docs.haystack.deepset.ai/docs/pipelines), it is possible to extend our RAG architecture to introduce nodes for tools, reasoning and agentic actions.

docs/retrieval-augmented-generation/rag-framework.md
```diff
@@ -0,0 +1,82 @@
...
+
+Upon acceptance of this integration our next step is to include additional ADRs for the subsequent components required for the RAG pipeline. Next up will be decisions on Milvus and containerized/non-containerized offerings of that solution.
+
+Upon denial of this integration the team will need to go back and evaluate alternative technologies and ensure they meet the needs of the project goal and ensure they meet the larger project plan goals of a configurable RAG pipeline.
+
+## How
+
+A downstream should be generated of the [https://github.com/deepset-ai/haystack](https://github.com/deepset-ai/haystack) project.
+
+## Alternatives
```
jwm4, line 60: Can we add the following to the list of alternatives?:

- Do not use a framework; write everything directly in Python or some other programming language
  - This would take longer to get started.
  - This would make it a lot more work to add more vector DBs since we'd need to add additional code for each vector DB we want to support.  We would hide that work behind an abstraction layer the same way the frameworks do, but it is work to build and maintain the abstraction layer, and the frameworks do that for us (and have put a lot of time and effort into doing it well).
  - This would make it harder to bring in advanced functionality that the frameworks already provide.  For example, Haystack provides support for RAG self-correction loops which we might want some day.
  - This might make it easier to bring in advanced functionality that the frameworks do _not_ already provide.  Frameworks provide an abstraction layer that is generally useful when you want to do things that the framework developers support but often counterproductive when you want to do things that the frameworks do not support.  For example, if there is a call to the framework that collapses multiple atomic steps into a single function call, that generally makes it harder to insert your own logic in between those atomic steps.

jwm4 (changes requested):
I have requested one more alternative to be added to the list that we considered -- I don't think that will change the outcome but I want it included for completeness.

I understand that the discussion of the dependency risk will be added to the risk section, which sounds good to me.

I am not seeing any other edits to this document that I would recommend.  So I think if these changes are made, I will approve (unless something else arises before then).

ilya-kolchinsky:
@jwm4, sure, I will edit the document to include the alternative you proposed.
As for the dependencies, I have already added this earlier today. To the best of my understanding, there seem to be no problems with the dependencies, therefore I added this discussion to the 'Why' section rather than to the risks. Do you think it would be better to move it under 'Risks'?

jwm4:
Oh, I missed it under "Why".   Sorry!  I am fine with it there.  I do think there is a tail risk that *new* dependencies get added that are more problematic.  If that happens we have the following options:
1. Pin to the old version that doesn't have that dependency.  That's often OK for a while, but eventually we're likely to run into updates that we need (e.g., critical fixes, compatibility with new vectordbs, etc.).
2. Fork the project to avoid the problematic dependencies.
3. Move off of Haystack completely.

Any of these could be viable depending on the circumstances.  Furthermore, since Haystack hasn't added any dependencies that we'd consider very problematic so far, the likelihood that they will in the future seems minimal.  I'd still like this noted under Risks though for reference if you don't mind.

ilya-kolchinsky:
Done - the document is now updated as we discussed.

dmartinol (approved):
lgtm

jwm4:
@instructlab/oversight-committee -- Can someone take a look at this and see if it is ready to approve and merge?  If it is not, can we get some guidance about what would be needed to get it approved?

jjasghar:
As an oversight committee member I approve this. Please fix the markdown/spellcheck errors and then we can merge this as a the way forward.

docs/retrieval-augmented-generation/rag-framework.md
```diff
@@ -0,0 +1,97 @@
...
+
+7. **Actively maintained open source project under permissive license.** Haystack is very [actively](https://github.com/deepset-ai/haystack/pulse/monthly) [maintained](https://github.com/deepset-ai/haystack/issues?q=is%3Aissue+is%3Aclosed) and [supported](https://github.com/deepset-ai/haystack/discussions). [Tagged versions](https://github.com/deepset-ai/haystack/releases) are released on a regular basis and [trusted publishing automation](https://github.com/deepset-ai/haystack/actions/workflows/pypi_release.yml) is used. Haystack is licensed under Apache 2.0, and all of its dependencies (jinja2, lazy-imports, more-itertools, networkx, numpy, openai, pandas, posthog, python-dateutil, pyyaml, requests, tenacity, tqdm, typing-extensions) are licensed under Apache, MIT, BSD or PSFL.
+
+## Goals
+
+- We need to identify a framework that will be used to help support the RAG work stream. This is one of many pieces involved in the RAG work stream but identifying the framework used to index and retrieve data is step 1 of the larger picture.
+
+## Non-goals
+
```
mairin, line 47: It would be worth reflecting this conversation here: https://github.com/instructlab/dev-docs/pull/164#issuecomment-2529783458

Here's what I'd suggest:

`The primary goal here is to provide a framework optimized for RAG specifically and that the decision isn't focused on optimizing the choice on the best agentic pattern solution. However, Haystack can potentially implement many agentic patterns (Eg https://docs.haystack.deepset.ai/v1.22/docs/agent). Since Haystack provides the functionality [to build arbitrarily complex pipelines](https://docs.haystack.deepset.ai/docs/pipelines), it is possible to extend our RAG architecture to introduce nodes for tools, reasoning and agentic actions.`

mairin (changes requested):
Requesting some changes, some minor spelling / grammar, some just pulling some items from the PR conversation into the doc.

mairin (approved):
LGTM, thanks!

PR #167, opened 2024-12-17
feat: Code Coverage Reporting via CI

courtneypacheco:
This document provides design suggestions for CI code coverage reporting. Initially, the high-level goal is to provide a report in each pull request to highlight:
  * Current code coverage of `main` for each InstructLab repo
  * The code coverage on the _new_ lines in the PR
  * What the code coverage will become if said PR is merged (e.g., 53% coverage -> 62% coverage)

We can ultimately decide if we want to enforce a minimum coverage % in PRs, but for now, I just want to report the coverage for informational purposes


docs/ci/ci-code-coverage-reporting.md
```diff
@@ -0,0 +1,138 @@
...
+## Code Coverage Reporting Bot: Options
+
+### [Codecov](https://github.com/marketplace/codecov)
+
+* [Example Codecov job](https://github.com/ansible/vscode-ansible/runs/33887209843)
+* Pros:
+  * Free to use in open source projects.
+  * Codecov reports all the data on GitHub in a comment. [Example](https://docs.codecov.com/docs/pull-request-comments).
+* Cons / Callouts:
+  * Not applicable at this time, but if we want to use it in >1 private repo, we must purchase a customer plan.
```
RobotSail, line 43: We don't have any private repos (or plan to), so this shouldn't affect us.

docs/ci/ci-code-coverage-reporting.md
```diff
@@ -0,0 +1,138 @@
...
+  * According to Codecov: "All subscribers of a Pull Request get an email when Codecov posts a comment (only when posting a new comment), and there's nothing we can do about it.” (Read more [here](https://docs.codecov.com/docs/pull-request-comments#filtering-emails).)
+
+### [Coveralls](https://github.com/marketplace/actions/coveralls-github-action)
+
+* [Example Coveralls job](https://github.com/gophercloud/gophercloud/actions/runs/12117814428/job/33858835625?pr=3244)
+* Pros:
+  * Free to use via [MIT license here](https://github.com/coverallsapp/github-action/blob/main/LICENSE.md).
+  * Appears to be free for open source repos, but like with Codecov, we should validate that the ToS doesn’t have fine print, etc.
+  * Can report all the data in a GitHub comment. [Example](https://github.com/coverallsapp/coveralls-node-demo/pull/19#issuecomment-1035344985).
```
RobotSail, line 52: This is a really nice addition.

docs/ci/ci-code-coverage-reporting.md
```diff
@@ -0,0 +1,138 @@
...
+* Cons / Callouts:
+  * Code coverage visuals, etc. are all viewed through the Coveralls front end (coveralls.io), so if that service were to go down, then we’re technically at their mercy.
```
RobotSail, line 54: If it's FOSS/MIT licensed, could we not stand up our own instance?

RobotSail (approved):
Thanks for this dev-doc, this will be a very positive change to have in our repos. I think ultimately all of the coverage tools will be good to have, so it seems like it's more about preference than necessarily picking the wrong one.

courtneypacheco:
> I am not sure the utility of this. Unless it is someone's job to increase the coverage amount, the coverage amount is just a metric no one cares about and is more emails from GitHub for each PR run. I don't think I have ever worked on an open source project where coverage reporting has ever been really helpful.
> 
> A lighter weight idea would be to simply upload the coverage data/reports as a build artifact which someone can go get if they really care. For example (the example is test reports, but same basic idea):
> 
> https://github.com/bndtools/bnd/blob/ce60cc705bea2ef75126cb17967b34a50ab3b4a6/.github/workflows/cibuild.yml#L103-L111

Yep, I know what you mean about metrics.

I don't think the number itself tells us much about code quality either because in theory, a repo can have "100% code coverage," yet still suffer quality issues. After all, if we write code wrong and then write our unit tests wrong as well, it is possible that the unit tests pass when they shouldn't.

However, if the code coverage is sitting at 53% and a contributor creates a large PR that reduces the overall code coverage to 30% (as a gross exaggeration), then we might want to look at it with more critical eyes. Same if someone creates a PR that adds 200 lines with only 5% code coverage.

Ultimately, my goal is not to bombard anyone with tons of emails, especially from a code coverage bot! But I do want to make it easy for contributors and reviewers to see if a certain file or critical piece of code was accidentally left untested. Yes, contributors can look at the actual code coverage report file instead of us having a bot commenting about untested code lines, but having coverage information summarized right in the PR would likely be helpful for PR reviewers.

What do you think?

bjhargrave:
> having coverage information summarized right in the PR would likely be helpful for PR reviewers.

Yes, but adding the comments sends emails. :-(

In any case, we need to be careful in the GitHub workflows to avoid security issues which can lead to supply chain attacks.

courtneypacheco:
> > having coverage information summarized right in the PR would likely be helpful for PR reviewers.
> 
> Yes, but adding the comments sends emails. :-(
> 
> In any case, we need to be careful in the GitHub workflows to avoid security issues which can lead to supply chain attacks.

Yep, agreed.

A couple of options off the top of my head to mitigate the number notifications:

  - We could disable the automatic code coverage bot reporting when PRs are marked as drafts, or
  - We could disable the bot comments by default, but allow any user to call it manually, such as when the PR author or reviewer is concerned about code quality — e.g., https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch, or
  - Set up filters to block certain bot emails — e.g., https://docs.github.com/en/account-and-profile/managing-subscriptions-and-notifications-on-github/setting-up-notifications/configuring-notifications

For security concerns regarding workflows: Ansible, [Santa](https://github.com/google/santa) and other popular open source repos do actively use Codecov or Coveralls to assess code coverage. I can definitely look into how these repos have mitigated security risks via their workflow logic and work to identify any concerns we may have if we were to introduce such workflows in any of our repos.

docs/ci/ci-code-coverage-reporting.md
```diff
@@ -0,0 +1,138 @@
...
+* Contributors should be able to _easily_ identify which new lines of code in their branch need more coverage.
+  * This means contributors should not need to dig through CI logging to determine code coverage.
+  * Ideally, a bot should leave a comment to summarize the code coverage for them OR the contributor should be able to find this information without having to search through text (e.g., logging).
+* Contributors should understand how well their branch code is covered relative to the current `main` branch code.
+
+## High-Level Functional Requirements
+
+* Code coverage should be reported in PR builds as a GitHub _bot_ comment
+* New code:
+  * Should show % coverage on new lines of code.
```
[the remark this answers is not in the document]
nathan-weinberg: This doesn't seem to be suggesting a requirement, just a communication of how much of the new code is being covered by testing

docs/ci/ci-code-coverage-reporting.md
nathan-weinberg: +1

docs/ci/ci-code-coverage-reporting.md
nathan-weinberg: True, but I'd prefer to avoid the overhead (is possible)

docs/ci/ci-code-coverage-reporting.md
```diff
@@ -0,0 +1,138 @@
...
+
+(**Note**: Most expensive $$$)
+
+* [Example SonarQube job](https://github.com/ansible/vscode-ansible/runs/34526218267)
+* Pros:
+  * Interactive, intuitive interface.
+  * Explains what isn’t covered and gives detailed suggestions on how to increase code coverage. (Seems better than Codecov, but… it’s also a more refined product?).
+  * Keeps a running history of your code coverage reporting, if desired.
+* Cons / Callouts:
+  * Appears to be "free to use," but Sonar wants you to store coverage data in their cloud and storage costs will apply unless you agree to host your own Sonar service -- in which case, we'll need to pay for a license. (Not ideal.)
```
nathan-weinberg, line 66: I say we avoid this one

docs/ci/ci-code-coverage-reporting.md
```diff
@@ -0,0 +1,138 @@
...
+  * HTML
+  * XML
+  * JSON
+  * lcov
+  * annotate
+* The README should display a code coverage badge
+
+## Code Coverage Reporting Bot: Options
+
+### [Codecov](https://github.com/marketplace/codecov)
```
nathan-weinberg, line 36: This would be my preference

cdoern: same

docs/ci/ci-code-coverage-reporting.md
cdoern: 👍

cdoern (approved):
this is very thorough, I like all of the ideas here. My preference for code coverage is CodeCov

docs/ci/ci-code-coverage-reporting.md
courtneypacheco: Agreed. It's a great tool, but let's avoid it unless we need those added features down the line.

docs/ci/ci-code-coverage-reporting.md
courtneypacheco: I'll look into this more to see how we can set it up.

ktdreyer:
I switched a few repos from coveralls.io to codecov.com around the time that I was switching from Travis CI to GitHub Actions, since Codecov integrated better with GH Actions at the time.

One really nice feature of Codecov is that it has a `fail_ci_if_error: false` option. Sometimes Codecov has outages, and this allows the build to stay green even when their GitHub app is broken. Longest outage I've seen was just for a day or so, but it's still disruptive. I'd recommend that setting for sure.

PR #168, opened 2024-12-19
ADR: Vector store for RAG

anastasds:
[#2869](https://github.com/instructlab/instructlab/issues/2869)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,10 @@ dictionary.dic
 
 # python virtualenv
 venv
+
+# Emacs
+*~
+\#*\#
+.\#*
+.projectile
+.dir-locals.el
--- a/docs/rag/adrs/adr-vectordb.md
+++ b/docs/rag/adrs/adr-vectordb.md
@@ -1,0 +1,26 @@
+# Initial InstructLab Vector Store
+
+## Context
+
+One of the first choices to make in implementing RAG is to choose an initial vector store to develop against. Though the usage of frameworks like LangChain or Haystack make it easy to swap vector databases, we need a working end to end implementation for RAG that is tested against and available to install with InstructLab. There are many options (see [here](https://docs.haystack.deepset.ai/docs/choosing-a-document-store)).
+
+Our main long-term requirements are that our chosen store have fully-developed document update (and thus some sort of notion of primary key), that it be scalable to cluster size, and that it have a permissive license (Apache, MIT, or similar). Among the available choices, [Milvus](https://milvus.io/) provides strategic advantage due to its [investment from watsonx](https://www.ibm.com/new/announcements/ibm-watsonx-data-vector-database-ai-ready-data-management).
+
+Milvus can be used in-process ([Milvus Lite](https://milvus.io/docs/milvus_lite.md)), single-node ([Milvus](https://milvus.io/docs/prerequisite-docker.md)), or cluster-scale ([Milvus Distributed](https://milvus.io/docs/prerequisite-helm.md)).
+
+## Decision
+
+InstructLab will initially integrate with and use Milvus Lite for vector storage and retrieval augmented generation.
+
+## Status
+
+Accepted
+
+## Consequences
+
+* Users will have a clear [upgrade path](https://milvus.io/docs/upgrade_milvus_cluster-operator.md) from the laptop use case to cluster scale.
+* We should be able to have access to expert resources with Milvus via IBM.
+* The laptop use case of InstructLab will have a minimally resource intensive option for prototyping.
+* Since Milvus is used in watsonx, we can have confidence that it can meet expected scaling requirements.
+* Document updates can be accommodated using well-established [primary key functionality](https://milvus.io/docs/primary-field.md) and [partition key](https://milvus.io/docs/use-partition-key.md).
+* There is a risk of developing against a mature vector store leading to usage of functionality not available in some other vector store that a potential customer requires to be used.
--- a/docs/rag/adrs/template.md
+++ b/docs/rag/adrs/template.md
@@ -1,0 +1,17 @@
+# ADR 00x: Succinct title
+
+## Context
+
+_What is the context of this decision? What are the technical, social, and political factors? For example, the decision to use a particular library might be simply because most of the team is familiar with it; that is a social context. A political factor might be influences from other teams or executive decisions_
+
+## Decision
+
+_a single decision statement, written in active voice, stated in a single sentence_
+
+## Status
+
+[Proposed | Accepted | Rejected ]
+
+## Consequences
+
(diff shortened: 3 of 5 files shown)


dmartinol:
What about moving adrs/ at the root folder, with README and the template files, and have a rag/ subfolder?
Apart from that, lgtm

anastasds:
That would probably need org-wide approval and we don't have that. I'll bring it up with the wider team next year to advocate for adoption of the format.

docs/rag/adrs/template.md
```diff
@@ -0,0 +1,17 @@
+# ADR 00x: Succinct title
```
jwm4, line 1: FWIW, I really dislike serial numbers on ADRs.  Whenever there are multiple ADRs in different phases of development moving at different speeds, it creates unneeded chaos.  A team I used to be on switched from serial numbers to date stamps in the title and it made everything run much more smoothly.

docs/rag/adrs/adr-001-vectordb.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Decision
+
+InstructLab will initially integrate with and use Milvus Lite for vector storage and retrieval augmented generation.
+
+## Status
+
+Accepted
+
+## Consequences
```
jwm4, line 19: I think we need something in here about dependencies of Milvus and their licensing terms because I know there have been a lot of complaints about that.  However, I don't remember the details so let's try to run them down offline, ASAP.

.gitignore
```diff
@@ -5,3 +5,10 @@ dictionary.dic
 
 # python virtualenv
 venv
+
+# Emacs
```
jwm4, line 9: Do we need these?  If so, would it make sense to have a separate PR for them?

jwm4 (changes requested):
I like the decomposition into Context, Decision, Status, and Consequences.  It seems much smoother and more natural than the approach used in https://github.com/instructlab/dev-docs/pull/164 which feels awkward and cumbersome to me.  I have some minor change requests here, but mostly I like the direction of both the ADR template and the proposed Milvus ADR.

docs/rag/adrs/template.md
anastasds: That's a good point. 

>A team I used to be on switched from serial numbers to date stamps in the title

I think that was our old team 👍 or maybe two different teams had the same idea. Anyway, the git repository will take care of tracking when something was merged and updating a date based on merge time is similarly annoying in the case that new commits dismiss approvals, so how about removing numbering entirely and letting git commit date be the sort key when needed?

jwm4: > I think that was our old team 👍 or maybe two different teams had the same idea.

Yes, I think actually the last two teams we were on together (WD and WA) both wound up adopting this.

> how about removing numbering entirely and letting git commit date be the sort key when needed

Yes, that sounds even better, thx!

docs/rag/adrs/adr-001-vectordb.md
anastasds: I tried to address the license issue in some way in the `Context` section. As for the dependencies issue, my perspective when writing this was that those issues are always present and are not specific to this decision, so there is no actual practical reason to call them out further.

.gitignore
anastasds: I use emacs so it is something that helps me prevent staging auto-save files.

>would it make sense to have a separate PR for them?

Generally, the reason for splitting things up into separate changesets (PRs) is so that if there is a functional regression or bug in a software codebase, you can selectively revert individual changes in order to find which code change introduced the issue. This is a documentation repository, so splitting this out would be organization for organization's sake. I don't see a reason to split it out and so will not unless a compelling reason is made.

anastasds:
>I like the decomposition into Context, Decision, Status, and Consequences. It seems much smoother and more natural than the approach used in https://github.com/instructlab/dev-docs/pull/164 which feels awkward and cumbersome to me.

I'm glad the approach resonates. Within reason, people should take priority over processes and not the other way around. This is a core tenet of the Agile Manifesto which, for better or for worse, has been a solid philosophical framework for ways of working for software teams since 2004:

>Individuals and interactions over processes and tools

https://agilemanifesto.org/

docs/rag/adrs/adr-vectordb.md
```diff
@@ -0,0 +1,26 @@
+# Initial InstructLab Vector Store
+
+## Context
+
+One of the first choices to make in implementing RAG is to choose an initial vector store to develop against. Though the usage of frameworks like LangChain or Haystack make it easy to swap vector databases, we need a working end to end implementation for RAG that is tested against and available to install with InstructLab. There are many options (see [here](https://docs.haystack.deepset.ai/docs/choosing-a-document-store)).
```
anastasds, line 5: How do I disable the below linting bot comment? This is a useless nitpick. @nathan-weinberg maybe you can help here?

docs/rag/adrs/template.md
anastasds: Done.

.gitignore
jwm4: OK.  The main reason I would have split it up into a separate PR is the governance model -- this seems like it could move quickly under the rules for trivial changes while the rest of the PR may turn out to be a much heavier lift.  However, I am fine with leaving it in and seeing how it goes.

jwm4 (changes requested):
I've added a couple more comments.  Also, I do think there is a lot of controversy around the Milvus dependencies specifically, but I am OK with waiting to see if someone else wants to challenge this decision on those grounds.

I see that this PR will need an update to the spellcheck dictionary and a couple of spelling corrections to pass the spell check.  Also, I see that it is not passing DCO.  I tried using the DCO guidance to bring my last dev-doc into compliance with DCO and it didn't work, but I do think it can be done with enough github expertise (just more than I have).  So I don't think it is ready to merge.  However, the only open item from *me* is the title line of the template.

jwm4 (approved):
The issues I raised earlier are all resolved now, so I am happy with this in its current form.

cdoern (approved):
this all looks really good to me. I really appreciate the detail you put in here about ADRs in general.

docs/rag/adrs/adr-vectordb.md
nathan-weinberg: You can run `make md-lint` locally to catch things locally so you don't need to wait for the GitHub CI to run

nathan-weinberg (approved):

anastasds:
@cdoern @nathan-weinberg thanks both! I don't have a merge button, so, if you would be so kind, that would get the template and guidance merged, and a subsequent ADR will record any changes regarding vector stores based on situational / informational changes.

PR #178, opened 2025-01-14
ADR: Feature Gating Strategy

anastasds:


jwm4 (approved):
This seems like a good idea to me.

jwm4:
I think the main point of the PR is to get agreement around using the OpenShift terminology of `DevPreviewNoUpgrade` and `TechPreviewNoUpgrade` to label feature gates, but I agree that implicitly this would also imply high level agreement that we want feature gates at all.

nathan-weinberg (approved):
I see the value here and generally agree with what is outlined in the doc. I would like to maybe see something about how we communicate various levels of "support" for features to users - for example, we could have sections in the `CHANGELOG.md` for this, e.g. RAG could be put under a `Dev Preview` section in the [CHANGELOG.md for the 0.23.0 release of Core](https://github.com/instructlab/instructlab/blob/main/CHANGELOG.md#v023)

anastasds:
Agreed @nathan-weinberg - added that to the `Consequences` section.

RobotSail (commented):
Maybe I'm missing something here from this document, but I'm struggling to understand why OpenShift's feature-gating process is relevant to InstructLab? 

In general I imagine that any product consuming InstructLab should be responsible for its own implementation of a gating system. I'm not sure why the responsibility of this would come down to an upstream project.

I'm not against a system that proposes a methodology for declaring certain features as experimental, but I feel like doing anything more might be adding more complexity than is needed. 

For instance, if a feature is there then it's there, if it's not then it's not. But all of the components present within a codebase must already be in agreement with each other else they wouldn't be there. So I'm not sure why we would want to gate functionality for an upstream project.

anastasds:
@RobotSail as described in the ADR, the motivation was to use pre-existing standard terminology within the larger ecosystem. 

The reason for feature gating is the same as in any application or product - let users opt into new experimental functionality that might introduce unforeseen regressions or bugs simply by being enabled. In a perfect world, such a thing would never happen and everything would be perfectly modularized, architected, etc., but we do not live in such a world.

It's the same concept as alpha functionality.

RobotSail:
Hey @anastasds thank you for the detailed response. I'll start off by saying that I fully support this idea. In general, I have no issue with having functionality for declaring features as experimental in an upstream community.  As a matter of fact, here's a great community project which provides exactly what you're describing here that we could use as inspiration: https://github.com/ipfs/kubo/blob/master/docs/experimental-features.md

Re-reading your ADR, I'm realizing that this is not focused around a product and just using OpenShift as a reference.

RobotSail (approved):

nathan-weinberg:
@anastasds if you can squash commits here I'm happy to merge it in, pending no objections from @instructlab/oversight-committee

anastasds:
@nathan-weinberg done

PR #179, opened 2025-01-16
feat: Create separate repository for our in-house GitHub actions

courtneypacheco:
This doc describes how we can reuse our in-house actions as our CI processes continue to grow.

### DIFF
--- a/docs/ci/ci-universal-actions.md
+++ b/docs/ci/ci-universal-actions.md
@@ -1,0 +1,80 @@
+# Create Separate Repo for Custom GitHub Actions
+
+## Idea Overview
+
+Create a separate repository within the `instructlab` GitHub org to house our custom CI Github Actions.
+
+## Motivation for this Proposal
+
+Within our `instructlab` GitHub org, we have an in-house GitHub action that we use across several repos: `free-disk-space`.
+
+Examples:
+
+- [`free-disk-space` in the `eval` repo](https://github.com/instructlab/eval/blob/main/.github/actions/free-disk-space/action.yml)
+- [`free-disk-space` in the `instructlab` repo](https://github.com/instructlab/instructlab/blob/main/.github/actions/free-disk-space/action.yml)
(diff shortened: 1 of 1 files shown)


nathan-weinberg:
@courtneypacheco can you socialize this on Slack/Discord/the mailing lists/wherever you want so we can get community input?

jjasghar (changes requested):
This seems like a reasonable expectation, but we should probably socialize this more before adoption.

If anything, we should talk about it at the Tuesday Community meeting.

markstur:
ci-actions repo sounds like a good idea to me

nathan-weinberg (approved):
+1 on this idea - I would vote on **not** publishing to GitHub Actions Marketplace - far prefer a single repo for the InstructLab org

bbrowning:
I'm +1 generally to having a central place for all of our in-house GitHub actions. My only question is do we want this to be a repository just for GitHub actions, or a repository for centralized CI more generally which may initially only include some common GitHub actions but that may expand to be other things? Or, is the thought that we can encapsulate the majority of what could be common across CI setups within GitHub actions?

ktdreyer (approved):
Proposal looks great. It will be great to reduce duplication like in that `free-disk-space` example.

The more we keep the focus on GitHub Actions content, the easier it will be to lint etc and guide new contributors.

@bbrowning what other types of things were you picturing hosting in this repository?

bbrowning:
@ktdreyer The other obvious thing we share across repos, besides GitHub actions themselves, are the workflow files. So, if we wanted to reduce duplication across the repositories by creating a couple of [reusable workflows](https://docs.github.com/en/actions/sharing-automations/reusing-workflows), would we also store those in this same place? Or would that be in a different centralized place?

The only reason I bring it up is that this is targeted specifically for our GitHub actions, so I just want to clarify if we mean only GitHub actions or if we mean this to house any shared CI bits, which may start out as only GitHub actions but grow in scope as necessary.

danmcp (commented):
The work being done here is another good example of what could be shared:

https://github.com/instructlab/training/pull/419

jjasghar (approved):

danmcp (approved):

PR #184, opened 2025-01-24
introduce document for instructlab-sdk

cdoern:
this document outlines the structure of the instructlab-sdk, the purpose, and the scope. The implementation specifics will be dealt with on the subsequent implementation PRs.

### DIFF
--- a/.spellcheck-en-custom.txt
+++ b/.spellcheck-en-custom.txt
@@ -6,6 +6,8 @@ agentic
 Akash
 AMDGPU
 Anil
+API
+api
 arge
 args
 arXiv
@@ -215,6 +217,8 @@ Salawu
 scalable
 SDG
 sdg
+SDK
+sdk
 semvar
 sexualized
 SHA
--- a/docs/sdk/instructlab-sdk.md
+++ b/docs/sdk/instructlab-sdk.md
@@ -1,0 +1,153 @@
+# InstructLab Python SDK
+
+## Motivation
+
+Today, the only way to "drive" the InstructLab opinionated workflow is via the `ilab` CLI. While this process provides a succinct way for everyday users to initialize a config, generate synthetic data, train a model, and evaluate it: the guardrails are quite limiting both in what a user can do and what the development team can add as features exposed directly to the user over time.
+
+Additionally, current consumers of InstructLab are finding themselves importing different library private and public APIs in combination with CLI functionality to achieve the workflows they want. While these more advanced usage patterns are not for everyone, providing ways to run bespoke and piecemeal workflows in a standardized and safe way for the community is a necessity.
+
+Unifying these various ranges of advanced workflows under an overarching `InstructLab Python SDK` will allow for new usage patterns and a clearer story on what InstructLab can and should provide as user accessible endpoints.
+
+While each library can and _should_ have their own publicly accessible SDK, not all functionality being added to SDG, Training, and Eval needs to be correlated directly to the "InstructLab workflow". This Python SDK should, as the CLI does, expose an opinionated flow that uses functionality from the various libraries. The InstructLab SDK should be derived from the library APIs, not the other way around. SDG, for example, currently has a `generate_data` method, meant to only be accessed by InstructLab. This method simply calls other publicly available SDG functionality. Orchestration of the InstructLab flow like this should not be of concern to the individual libraries and instead be handled by the overarching InstructLab SDK which will maintain the user contracts. The InstructLab SDK will need to work within the bounds of what the Libraries expose as public APIs.
+
+The benefit of the above is that the opinionated flow can be accessed in a more nuanced and piecemeal way while also gaining the potential for more advanced features. Say a consumer wants to:
+
+1. Setup a custom config file for ilab (optional)
+2. Initialize a taxonomy
+3. Ensure their taxonomy is valid
+4. Ingest some data for RAG and SDG (SDG coming soon)
+5. Generate synthetic data using an InstructLab pipeline
+6. Do some custom handling per their use case
+7. Fine-tune a model using the custom config they initialized for their hardware
+8. Evaluate their model after training using various benchmarks
+
+A user could do this if they had an SDK.
+
+(the structure of the SDK and actual arguments is discussed below)
+
+However today, users are forced to run a sequence of commands tailored to only work with the proper directory structure on the system.
+
+## Major Goals
+
+1. Modularize the InstructLab workflow such that any part can be run independently
+2. Allow users to choose whether or not to take advantage of the config/system-profile method of running InstructLab. Meaning they do not need any pre-existing configuration to run the SDK.
+3. Standardize user contracts for the existing functionality of the InstructLab workflow. Existing CLI commands should be using the SDK once past click parsing, not separate code.
+4. Define Contracts loose enough that functionality can be expanded as more advanced features are released.
+5. Document SDK usage in upcoming InstructLab releases.
+
+## Non-Goals
+
+1. Exposing all library functionality immediately
+2. Replacing CLI
+3. Shipping an SDK that is generally available as opposed to v1alpha1 or v1beta1.
+
+## Design
+
+### Versioning
+
+The SDK would start at version v1alpha1 such that it can change/break at any time for the first few iterations as libraries adjust their API surface.
+
+### Structure
+
+This SDK should live in a net new package inside of `instructlab/instructlab` preferably to limit unnecessary imports in a new repository. The SDK could be imported as `instructlab.core...`
+
+The user surface initially should look like this:
+
+`instructlab.core` contains all SDK definitions. Users can `from instructlab.core import...` to use specific SDK classes
+
+For most of the existing InstructLab command groups, there should be a class:
+
+`from instructlab.core import, Config, Taxonomy, Data, Model, RAG, System`
+
+The full list of classes and their methods for now (subject to change during development process):
+
+```console
+instructlab.core.Config
+instructlab.core.Config.init
+instructlab.core.Config.show (get)
+instructlab.core.Taxonomy
+instructlab.core.Taxonomy.diff
+instructlab.core.System
+instructlab.core.System.info
+instructlab.core.Data
+instructlab.core.Data.ingest
+instructlab.core.Data.generate_data
+instructlab.core.Model
+instructlab.core.Model.serve
+instructlab.core.Model.train_model
+instructlab.core.Model.process_data (calling the training library's data process class in a safe way)
+instructlab.core.Model.evaluate_mt_bench
+instructlab.core.Model.evaluate_dk_bench
+instructlab.core.Model.evaluate_mmlu_bench
+instructlab.core.RAG.ingest
+instructlab.core.RAG.convert
+```
+
+a brief example:
+
+```python
+
+from instructlab.core import Config, Taxonomy, Data, Model
+
+config_object = Config.init(...)
+diff = Taxonomy.diff()
+
+if diff:
+    data_client = Data(data_path="", teacher_model="", num_cpus="", taxonomy_path="",)
+
+    # not in v1alpha1
+    data_path = data_client.ingest()
+    # not in v1alpha1
+
+    openai_compat_client = some_server()
+
+    data_jsonls = data_client.generate_data(client=openai_compat_client, data=data_path)
+
+    some_custom_handling(data_jsonls)
+
+    # you can either use a config obj or pass trainer args
+    model_client = Model(student_model=path_to_student_model, configuration=config_object)
+
+    model_path = model_client.train_model()
+
+    # since we initialized the model client with the config, the training args are passed implicitly
+    eval_output = model_client.mt_bench(model_path=model_path)
+```
+
+The above example utilizes the configuration object to instantiate the `Model` class. However, a user could also pass `training_args=` directly to `model_client.train_model` to override the configuration class defaults. This allows the SDK to utilize the System Profiles of the ilab CLI but not rely on them too much.
+
+Presumably, the distinct methods under each class will grow, which is why I am opting to make very distinct classes per command group. Another benefit to the parent classes is that individual methods can inherit defaults from the instantiation of the object.
+
+These initial exposed functions can expand to include any new functionality that is more SDK oriented from the various libraries. For example, if SDG adds something like subset selection, teacher as annotator, data mixing, etc we could expose an `instructlab.core.Data.annotate` or `instructlab.core.Data.mix` that could be invoked in sequence in a user's script with other parts of the ilab workflow. Some things make _less_ sense to be exposed via a CLI, but still are critical to ensuring users get a good model and properly generated data.
+
+There are certain things that only exist in `ilab` currently and functionality that is going to be moving these such as data ingestion, RAG, etc. Forming an SDK for `instructlab` allows us to capture all of these concerns under one API.
+
+These endpoints in combination with the curated InstructLab Config File will open up these workflows to users and allow InstructLab to be easily incorporated into other projects. Allowing people to run things like data generation, and full fine-tuning via an SDK that pulls in their pre-existing `config.yaml` but also can be run independently will open new avenues for InstructLab adoption and extensibility.
+
+## Changes to the CLI
+
+The `ilab` CLI will need to adapt to this new structure. Commands like `ilab data generate` should, in terms of code, follow this flow:
+
+1. `src/instructlab/cli/data/generate.py`
+2. `src/instructlab/data/generate.py`
+3. `src/instructlab/process.py`
+4. `src/instructlab/core/data/generate.py`
+
+So generally: cli -> process management package to kick off a sub-process -> internal handling package -> core SDK (public definitions) -> library code, is the flow.
+
+The flow of the CLI today is such that the cli package for a command (`src/instructlab/cli/data/generate.py`) parses the command line options, manages creating a sub-process, and passes control to the core code (`/src/instructlab/core/data/generate.py`). This then calls out to the library APIs
+
+The internal handling package is necessary as it allows us to split off a sub-process when it makes the most sense for us before calling the library code directly. This is how the CLI works today.
+
+The difference with an SDK is that we would eventually want to end up executing `core/data/generator.py`, the actual publicly consumable python SDK. This will ensure that the CLI can do whatever custom handling it needs to do on top, but eventually it must boil down to the `core` package which uses publicly available methods from the various libraries.
+
+## Scope of work
+
+In upcoming releases the InstructLab team should aim to:
+
+1. Design the SDK given the structure above
+2. Converse with Library maintainers to negotiate user contracts
+3. Begin work to re-architect how the CLI works using the SDK
+4. Publish an alpha SDK for public consumption
+
+After this initial work, the team can scope adding net new functionality that is not in the CLI to the SDK.


docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
+# InstructLab SDK
+
+## Motivation
+
+Today, the only way to "drive" the InstructLab opinionated workflow is via the `ilab` CLI. While this process provides a succinct way for everyday users to initialize a config, generate synthetic data, train a model, and evaluate it: the guardrails are quite limiting both in what a user can do and what the development team can add as features exposed directly to the user over time.
+
+Additionally, current consumers of InstructLab are finding themselves importing different library private and public APIs in combination with CLI functionality to achieve the workflows they want. While these more advanced usage patterns are not for everyone, providing ways to run bespoke and piecemeal workflows in a standardized and safe way for the community is starting to seem like a necessity.
+
+Unifying these various ranges of advanced workflows under an overarching `InstructLab SDK API` will allow for new usage patterns and a clearer story on what InstructLab can and should provide as user accessible endpoints.
```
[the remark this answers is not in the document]
cdoern: good point, folks have been colloquially calling this "The SDK API for InstructLab", but I will use just SDK to keep this simple

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Major Goals
+
+1. Modularize the InstructLab workflow such that any part can be run independently
+2. Allow users to choose whether or not to take advantage of the config/system-profile method of running InstructLab. Meaning they do not need any pre-existing configuration to run the SDK.
+3. Standardize user contracts for the existing functionality of the InstructLab workflow. Existing CLI commands should be using the SDK once past click parsing, not separate code.
+4. Define Contracts loose enough that functionality can be expanded as more advanced features are released.
+5. Document SDK usage in upcoming InstructLab releases.
+
+## Non-Goals
```
RobotSail, line 45: I love that you've included this as part of the dev-doc.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+The user surface initially should look like this:
+
+```console
+instructlab.api.configuration.init
+instructlab.api.taxonomy.diff
+instructlab.api.data.generate
+instructlab.api.data.list
+instructlab.api.model.train
+instructlab.api.model.evaluate
+instructlab.api.model.chat (one shot request probably)
```
RobotSail, line 70: I feel like `ilab chat` today provides an unmatched experience for having a local chat with models. But we shouldn't expose this through the SDK. The reason being here is that we'd simply be wrapping around LlamaCPP or vLLM, all for a single message request. At that point, it would be faster for them to just load transformers directly and manually generate from the model. 

In particular, whenever I'm testing models programatically, my workflow looks like:

```
1. Spin up a vLLM server with the model I want
2. Run a script with whatever pre-existing messages I've written
```

The script itself is super simple too, like:

```py
import openai

client = openai.Client(
    api_base="http://localhost:8000/v1",
    api_key="empty"
)

messages = [
    {
        "role": "system",
        "content": "You are an AI assistant"
    },
    {
        "role": "user",
        "content": "Why is Linux better than Windows?"
    }
]

response = client.chat.completions.create(messages, model="my-model")
assistant_response = response.choices[0].content
```

So if you're already at the level of consuming an SDK, I feel like it's a very simple task, so exposing it might not give much benefit.

RobotSail (approved):
This is a good document overall and a step in the right direction.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Structure
+
+This SDK should live in a net new package inside of `instructlab/instructlab` preferably to limit unnecessary imports in a new repository. It could be imported as `instructlab.api...`
+
+The user surface initially should look like this:
+
+```console
+instructlab.api.configuration.init
+instructlab.api.taxonomy.diff
```
nimbinatus, line 65: As this structure is what the current CLI has, I may be asking a larger question and won't be offended if you shut this as resolved to handle it somewhere else. Would it make more sense for the `.taxonomy.diff` tree to be under `.data`, to eventually leave room for alternative forms of data sorting and ingestion into the pipeline (i.e., is taxonomy a subfield of data management  overall, with SDG as another subfield)?

nimbinatus (commented):
Just one question on my end :) And it may be too far out of scope; feel free to ignore it!

jwm4 (approved):
This seems like a good idea.  A lot of work though, of course.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
+# InstructLab SDK
+
+## Motivation
+
+Today, the only way to "drive" the InstructLab opinionated workflow is via the `ilab` CLI. While this process provides a succinct way for everyday users to initialize a config, generate synthetic data, train a model, and evaluate it: the guardrails are quite limiting both in what a user can do and what the development team can add as features exposed directly to the user over time.
+
+Additionally, current consumers of InstructLab are finding themselves importing different library private and public APIs in combination with CLI functionality to achieve the workflows they want. While these more advanced usage patterns are not for everyone, providing ways to run bespoke and piecemeal workflows in a standardized and safe way for the community is starting to seem like a necessity.
```
nathan-weinberg, line 7:
```suggestion
Additionally, current consumers of InstructLab are finding themselves importing different library private and public APIs in combination with CLI functionality to achieve the workflows they want. While these more advanced usage patterns are not for everyone, providing ways to run bespoke and piecemeal workflows in a standardized and safe way for the community is a necessity.
```

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+The benefit of the above is that the opinionated flow can be accessed in a more nuanced and piecemeal way while also gaining the potential for more advanced features. Say a consumer wants to:
+
+1. setup a custom config file for ilab (optional)
+2. initialize a taxonomy
+3. ensure their taxonomy is valid
+4. ingest some data for RAG and SDG (SDG coming soon)
+5. generate synthetic data using an InstructLab pipeline
+6. do some custom handling per their use case
+7. fine-tune a model using the custom config they initialized for their hardware
```
nathan-weinberg, line 21: Nit but can you capitalize the beginning of these

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+a user could do the following to accomplish this if they had an SDK:
```
nathan-weinberg, line 23:
```suggestion
A user could do the following to accomplish this if they had an SDK:
```

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+1. Exposing all library functionality immediately
+2. Replacing CLI
+3. Shipping an SDK that is generally available as opposed to v1alpha1 or v1beta1.
+
+## Design
+
+### Versioning
+
+The SDK would start at version v1alpha1 such that it can change/break at any time for the first few iterations as libraries adjust their API surface.
```
nathan-weinberg, line 55: Is there a reason to do `v1alpha1` specifically other than that it's an OpenShift convention?

nathan-weinberg (commented):
Overall looking good to me, a few comments

.spellcheck-en-custom.txt
```diff
@@ -6,6 +6,9 @@ agentic
 Akash
 AMDGPU
 Anil
+annotater
```
anastasds, line 9: `annotator`?

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+a user could do the following to accomplish this if they had an SDK:
+
+```python
+instructlab.api.configuration.init(taxonomy=path_to_taxonomy)
+if instructlab.api.taxonomy.diff():
+    instructlab.api.rag.ingest(...)
+    instructlab.api.data.ingest(...)
+    data_paths = instructlab.api.data.generate(...)
+    some_custom_handling(data_paths)
+    instructlab.api.model.train(data_path=..., strategy=lab_multiphase)
```
anastasds, line 32: Having `*.api.*` submodules makes this seems a bit like defining interfaces, which isn't really a pattern in Python. What does having `.api.` subpackages provide? Would they provide some sort of stability guarantee?

anastasds (commented):
Is this really an SDK? Is this not just systematization of interfaces in Python modules, that is, Python APIs? For example, there is no "SDK" artifact to download.

This is actually about cleanly modularizing the codebase, isn't it?

docs/sdk/instructlab-sdk.md
RobotSail: I thought about this as well, but I like it so I feel like it's fine to keep. This is a Kubernetes thing btw, not OpenShift-specific.

docs/sdk/instructlab-sdk.md
cdoern: I am actually thinking of renaming this `.core....` since this isn't an API.


Grouping the code that we want folks to publicly consume is a common trend

cdoern:
@anastasds:

> Is this really an SDK? Is this not just systematization of interfaces in Python modules, that is, Python APIs? For example, there is no "SDK" artifact to download.
> This is actually about cleanly modularizing the codebase, isn't it?


The goal of this is to have folks who are currently haphazardly consuming different parts of InstructLab to have a centralized "core" of code that is versioned and static. What we call it, I don't feel strongly about, but it is pretty close to an SDK I believe. If there is some sort of standardization we need to fall into, we can. but other projects like [llama stack](https://github.com/meta-llama/llama-stack-client-python) do something similar but they _do_ package it separately. 

We could scope packaging it separately but having it live within the `instructlab` repo.

docs/sdk/instructlab-sdk.md
cdoern: not really. we can use version numbers, I just think the OCP way of doing it is helpful in that `alpha` Is basically throwaway

nathan-weinberg: I'm fine to keep it!

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+```
+
+However today, users are forced to run a sequence of commands tailored to only work with the proper directory structure on the system.
+
+## Major Goals
+
+1. Modularize the InstructLab workflow such that any part can be run independently
+2. Allow users to choose whether or not to take advantage of the config/system-profile method of running InstructLab. Meaning they do not need any pre-existing configuration to run the SDK.
+3. Standardize user contracts for the existing functionality of the InstructLab workflow. Existing CLI commands should be using the SDK once past click parsing, not separate code.
```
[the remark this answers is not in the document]
williamcaban: should we consider `core` be the actual main library instead of having a subcomponent of `core`? Like `from instructlab import InstructLab` and then have things like config init as part of that initialization.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+instructlab.core.model.train
+instructlab.core.model.evaluate
+instructlab.core.model.chat (one shot request probably)
+instructlab.core.model.serve (vLLM and llama-cpp-python)
+instructlab.core.model.download
+instructlab.core.model.list
+instructlab.core.system.info
+instructlab.core.rag.ingest
+instructlab.core.rag.convert
+```
```
williamcaban, line 77: The intent with the Python SDK should be to serve the Data Scientists and AI engineers doing experiments as the primary use. Then the constructs required for embedding or extending InstructLab from other products and platforms. 

For a data science experience I propose to be inspired to what sklearn or keras or pytorch high level structure provide for this persona.

For example, what would it take to have an experience  similar to :
```python


from instructlab import InstructLab, SDG, Train, Eval

ilab = InstructLab(taxonomy=<path_to_taxonomy>) # and on initizalization, the taxonomy is diff & loaded

# like in Pandas DataFrame.describe()
ilab.describe() # prints a summary of the taxonomy diff and attributes
ilab_df = ilab.data.ingest() # ingest the data from the taxonomy and return a dataset as a Pandas DataFrame

oa_client = {
    "url": "https://api.openactive.io/v1",
    "api_key": "YOUR_API_KEY"
}

##
# SDG interactions
##
ilab_sdg=SDG(client=oa_client,
             data=ilab_df,
             teacher_model={<model_and_attributes>},
             ) # ilab SDG class

ilab_sdg.load_pipeline(<path_to_pipeline>) # load a pipeline from a file

for block in ilab_sdg.pipeline.get_blocks():
    # do block customization
    ilab_sdg.pipeline[block].something()
    ilab_sdg.pipeline[block].block.foo = "bar"

ilab_sdg.run(
    callback=<callback_function>, # a function to call after executing each block (e.g. to report progress, or save intermediate results)
) # generate the SDG data

##
# Training interactions
##
ilab_train = Train(dataset=ilab_sdg.data, 
                   student_model={<model_and_attributes>},
                     ) # ilab Train class

ilab_train.run(
    callback=<callback_function>, # a function to call after each cycle/teration/epoch loop is run (e.g. to report progress, or save intermediate results, or to stop the training if certain stopping criteria are met)
) # execute the Training loop

ilab_train.model_save(<path_to_save>, resolution={}, model_format={gguf|safe_tensors|onnx|etc}) # save the model to a file in the specified format

##
# Evaluation
###

ilab_eval = Eval(dataset=[<path_to_eval_dataset>], 
                 endpoints=[<path_to_endpoints_for_eval>], 
                 ) # ilab Eval class

ilab_eval.evals(
    [list_of_evaluations], # a list of evaluations to run (e.g. accuracy, precision, dk-bench, etc.)
)

ilab_eval.run(
    callback=<callback_function>, # a function to call after each evaluation is run (e.g. to report progress, or save intermediate results)
)

ilab_eval.summary() # print a summary of the evaluation results

ilab_eval.save(<path_to_save>, format={jsonl|parquet|csv|xls}) # save the evaluation results to a file

ilab_eval.export_report(<path_to_save>, format={html|pdf}) # report with  the evaluation results
```

I would propose to focus on accelerating the goal of the persona (the data scientists), and not on mapping or exposing the InstructLab internal architecture.

cdoern: I think this is possibly a good end goal, but would require significant  re-architecture of InstructLab, and the libraries (what they expose).

For the purposes of this dev-doc, I can incorporate some of this into what I am proposing, but for an alpha SDK, keeping the interactions as simple yet expandable as possible is my goal. 

So we should aim to not require library adjustments/changes at first, and then we can add functionality once the structure is in place.

Let me incorporate some of this and I will update the PR

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
+# InstructLab SDK
```
booxter, line 1: probably obvious to you but consumers would be better served if we refer and publish the SDK as "Python SDK" (to contrast with whatever other bindings we could have in the future).

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Motivation
+
+Today, the only way to "drive" the InstructLab opinionated workflow is via the `ilab` CLI. While this process provides a succinct way for everyday users to initialize a config, generate synthetic data, train a model, and evaluate it: the guardrails are quite limiting both in what a user can do and what the development team can add as features exposed directly to the user over time.
+
+Additionally, current consumers of InstructLab are finding themselves importing different library private and public APIs in combination with CLI functionality to achieve the workflows they want. While these more advanced usage patterns are not for everyone, providing ways to run bespoke and piecemeal workflows in a standardized and safe way for the community is a necessity.
```
booxter, line 7: What's the mechanism to collect these real use patterns from external consumers, to make sure their needs are properly served?

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+Unifying these various ranges of advanced workflows under an overarching `InstructLab SDK API` will allow for new usage patterns and a clearer story on what InstructLab can and should provide as user accessible endpoints.
+
+While each library can and _should_ have their own publicly accessible API, not all functionality being added to SDG, Training, and Eval needs to be correlated directly to the "InstructLab workflow". This Python SDK should, as the CLI does, expose an opinionated flow that uses functionality from the various libraries. The InstructLab SDK should be derived from the library APIs, not the other way around. SDG, for example, currently has a `generate_data` method, meant to only be accessed by InstructLab. This method simply calls other publicly available SDG functionality. Orchestration of the InstructLab flow like this should not be of concern to the individual libraries and instead be handled by the overarching InstructLab SDK which will maintain the user contracts. The InstructLab SDK will need to work within the bounds of what the Libraries expose as public APIs.
```
booxter, line 11: (Thinking while reading) "an opinionated" (but perhaps LESS opinionated?) flow.

booxter: > within the bounds of what the Libraries expose as public APIs

What's the current mechanism in which libraries communicate which APIs are public and which are not? Or is the definition of a "public API" here, loosely, "whatever ilab already pulls from the library"? Or, alternatively, that will be the result of "contract negotiation" that you mention below?

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+```python
+instructlab.core.configuration.init(taxonomy=path_to_taxonomy)
+if instructlab.core.taxonomy.diff():
+    instructlab.core.rag.ingest(...)
+    instructlab.core.data.ingest(...)
+    data_paths = instructlab.core.data.generate(...)
+    some_custom_handling(data_paths)
+    instructlab.core.model.train(data_path=..., strategy=lab_multiphase)
+```
```
booxter, line 33: Not an issue with the proposal itself since the snippet is only illustrative, but I'd consider whether ilab SDK should promote using file paths for any artifacts. I'd rather have ilab encapsulate the knowledge about storage options (for models and datasets) and then allow SDK users to refer to the artifacts by their names. (This would be in line with the spirit of https://github.com/instructlab/instructlab/issues/1832 and some other issues under https://github.com/instructlab/instructlab/issues/1871)

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+## Changes to the CLI
+
+The `ilab` CLI will need to adapt to this new structure. Commands like `ilab data generate` should, in terms of code, follow this flow:
+
+1. `src/instructlab/cli/data/generate.py`
+2. `src/instructlab/data/generate.py`
+3. `src/instructlab/process.py`
+4. `src/instructlab/api/data/generate.py`
+
+So generally: cli -> internal handling package -> process management package -> api -> library code, is the flow.
```
booxter, line 94: api -> core? (though perhaps "api" is a better name for a module that is meant to be consumed externally. That said, API may also refer to upcoming REST API initiative, so perhaps it's better to avoid this word either. `public` could `sdk` would be another option with clear intent.) Whatever the final name, these should be unified to use just one across the doc.

booxter (commented):
This is a great idea. We should incorporate real experience from current consumers into the design of the Python public API.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+The flow of the CLI today is such that the cli package for a command (`src/instructlab/cli/data/generate.py`) parses the command line options, and passes control to the core code (`/src/instructlab/data/generate.py`). This then (if applicable), yields control to the `process` package which kicks off necessary processes for the given command.
+
+The difference with an SDK is that we would eventually want to end up executing `api/data/generate.py`, the actual publicly consumable python SDK. This will ensure that the CLI can do whatever custom handling it needs to do on top, but eventually it must boil down to the `api` package which uses publicly available methods from the various libraries.
```
alimaredia, line 98: If I understand this right `src/instructlab/data/generate.py` would just kick off processes using the code in `api/data/generate.py`?  Is there a reason that couldn't be done in `cli/data/generate.py`? Is there a reason we need to go from 3 files -> 4 to do the same thing?

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,109 @@
...
+
+The benefit of the above is that the opinionated flow can be accessed in a more nuanced and piecemeal way while also gaining the potential for more advanced features. Say a consumer wants to:
+
+1. Setup a custom config file for ilab (optional)
+2. Initialize a taxonomy
+3. Ensure their taxonomy is valid
+4. Ingest some data for RAG and SDG (SDG coming soon)
+5. Generate synthetic data using an InstructLab pipeline
+6. Do some custom handling per their use case
+7. Fine-tune a model using the custom config they initialized for their hardware
```
alimaredia, line 21: Evaluation, both quantitative (with benchmarks) and qualitative (just doing ilab chat) needs to be added to this list.

docs/sdk/instructlab-sdk.md
alimaredia: I think @williamcaban has a point here about the persona we're building this SDK for.  Who are we making this SDK for? Datascientist? People trying to build REST-API based instructlab services? The design of the SDK might vary based on who our target user of the SDK would be.

bbrowning: It's definitely important to understand the persona and goal of the SDK. William is proposing a fairly large re-architecting of the APIs, while this dev doc seems mostly focused on exposing existing CLI flows and functionality via a Python SDK. Is our goal to expose the existing end-to-end flow via a Python SDK? Or to provide a new way to interact with various InstructLab components that's more granular and flexible in how you compose your own end-to-end workflow from the pieces we're exposing?

cdoern: I am making some updates to the PR (some are already up), aiming to meld the two approaches a bit, lmk what you think

docs/sdk/instructlab-sdk.md
cdoern: thanks!

docs/sdk/instructlab-sdk.md
cdoern: good question. Most of our usage information is coming from internal use cases combined with reports of what people are asking for colloquially. We have no request gathering system as of yet

docs/sdk/instructlab-sdk.md
cdoern: most libraries have made APIs over time and are just now standardizing what they want public vs private. Part of the process of building this SDK will be making final decisions on that front.

In regard to opinionated, yes this will be less opinionated and more user driven

anastasds:
@cdoern I want to revisit and re-emphasize the point I raised earlier.

I really do not see how this is an SDK. This is about Python API design and modularization.

The act of logically grouping operations, giving them good names, and understandable function signatures - of defining the "nouns" and "verbs" in the application - is an act of... wait for it... [domain modeling](https://docs.google.com/document/d/11fKIVlL2jpSASHSIMFlM_rqw5lLlSoidtVVg6tHbuw8/edit?tab=t.0#heading=h.6plu6egletqh)!

The work being proposed here is fundamentally a modeling and modularization effort - it is not about creating an SDK.

docs/sdk/instructlab-sdk.md
bbrowning: I think the updates look like a reasonable enough place for work to start.

The actual Python code and list of APIs is only illustrative, right? From reading this, it's my understanding that the actual parameters and methods on individual classes shown here are just an example since you call out future work is to design the actual SDK based on the structure above and negotiate user contracts with library maintainers. For example, if SDG points out that the `Data.Generator` constructor needs different params or that we have to actually use the taxonomy diff results as input to data generation as opposed to just checking if there is a diff, this kind of thing can be ironed out later?

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Design
+
+### Versioning
+
+The SDK would start at version v1alpha1 such that it can change/break at any time for the first few iterations as libraries adjust their API surface.
+
+### Structure
+
+This SDK should live in a net new package inside of `instructlab/instructlab` preferably to limit unnecessary imports in a new repository. The SDK could be imported as `instructlab.core...`
```
nathan-weinberg, line 76: What does this look like practically? I'm not sure I agree with the idea of have a second package contained inside of this repository

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,171 @@
...
+
+The user surface initially should look like this:
+
+`instructlab.core` contains all SDK definitions. Users can `from instructlab.core import...` to use specific SDK classes
+
+For most of the existing InstructLab command groups, there should be a class:
+
+`from instructlab.core import, InstructLab, Eval, Data, Model`
+
+However, things like configuration initialization, taxonomy management, system info are utilities. So they can live in a centralized `InstructLab` client class.
```
nathan-weinberg, line 86: I disagree with this point - things like `Config`, `Taxonomy`, `System` should have their own client classes

Few reasons for this:
1. More extendable in the future
2. Clearer line drawn from CLI functionality 
3. Less confusing - importing `instructlab.core.InstructLab` does tell me as a developer what I'm getting in terms of functionality, especially given the mapping of commands to classes for everything else

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+So the full list:
+
+```console
+instructlab.core.InstructLab
+instructlab.core.InstructLab.taxonomy.diff
+instructlab.core.InstructLab.system.info
+instructlab.core.InstructLab.config.show
+instructlab.core.InstructLab.config.get
```
nathan-weinberg, line 122: To illustrate my point above, I would prefer 
```suggestion
instructlab.core.Taxonomy.diff
instructlab.core.System.info
instructlab.core.Config.show
instructlab.core.Config.get
```

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,171 @@
...
+
+The SDK would start at version v1alpha1 such that it can change/break at any time for the first few iterations as libraries adjust their API surface.
+
+### Structure
+
+This SDK should live in a net new package inside of `instructlab/instructlab` preferably to limit unnecessary imports in a new repository. The SDK could be imported as `instructlab.core...`
+
+The user surface initially should look like this:
+
+`instructlab.core` contains all SDK definitions. Users can `from instructlab.core import...` to use specific SDK classes
```
nathan-weinberg, line 80: Why not just use `sdk` instead of `core`?

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,171 @@
...
+1. `src/instructlab/cli/data/generate.py`
+2. `src/instructlab/data/generate.py`
+3. `src/instructlab/process.py`
+4. `src/instructlab/core/data/generator.py`
+
+So generally: cli -> internal handling package -> process management package -> core SDK -> library code, is the flow.
+
+The flow of the CLI today is such that the cli package for a command (`src/instructlab/cli/data/generate.py`) parses the command line options, and passes control to the core code (`/src/instructlab/data/generate.py`). This then (if applicable), yields control to the `process` package which kicks off necessary processes for the given command.
+
+The difference with an SDK is that we would eventually want to end up executing `core/data/generator.py`, the actual publicly consumable python SDK. This will ensure that the CLI can do whatever custom handling it needs to do on top, but eventually it must boil down to the `core` package which uses publicly available methods from the various libraries.
```
nathan-weinberg, line 160: Would it not make more sense for both the CLI and "Core" to be consuming from the generation and process logic separately? Why couple the CLI with the SDK at all?

docs/sdk/instructlab-sdk.md
JamesKunstle: I agree w/ this naming convention.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,171 @@
...
+
+diff = ilab_client.taxonomy.diff()
+
+if diff:
+    # not in v1alpha1
+    data_ingestor = Data.Ingestor(data_path="",...)
+    data_path = data_ingestor.ingest()
+    # not in v1alpha1
+
+    sdg_client = Data.Generator(client=openai_compat_client, data=data_path, teacher_model=path_to_model)
```
JamesKunstle, line 100: It'd be really cool if we could use the vLLM offline inference features in an SDK scenario like this. My understanding is that that's the highest throughput calling pattern if requests are well-organized to begin with.

docs/sdk/instructlab-sdk.md
JamesKunstle: This could be accomplished w/ `__init__.py` module visibility too.

docs/sdk/instructlab-sdk.md
JamesKunstle: It's possible we'd want this in some situations, but I think it's a good design decision to plop the CLI on top of the SDK as much as possible. For one thing, it minimizes duplicate code doing similar things. Another good thing is that we'd be testing our own SDK by wrapping it- testing the CLI would also be (only slightly indirectly) testing the SDK.

JamesKunstle (commented):
I really like the design changes represented in this document.
One consideration I'd raise as we proceed is that we may want to avoid the object-oriented pattern that the HF Trainer object is representative of, and prefer a functional programming convention. 
The ravioli-code "Trainer" / "Generator" classes are a problematic design pattern elsewhere in this problem space. `instructlab/training` has avoided duplicating this convention because it's harder to reason about and encourages large procedural instance methods.

Instead, it's more readable, hackable, and composable to have an SDK expose functional building blocks w/ attention to the interfaces informing the call.

For illustration, why ever use:
```python
trainer = Trainer(
    data=data,
    model=model,
    epochs=5,
    lr=1e-4,
    ...
    )

trainer.train()
```

when one could just run:
```python
train(
    data=data,
    model=model,
    epochs=5,
    lr=1e-4,
    ...
    )
```

docs/sdk/instructlab-sdk.md
cdoern: Nathan and I discussed this, and will be adding classes for these operations

docs/sdk/instructlab-sdk.md
cdoern: just as a note, these classes are departing from the structure of the other ones where its `Model.Trainer.train_model` for example, is this something we are ok with?

docs/sdk/instructlab-sdk.md
cdoern: Nathan and I talked and going w/ core makes more sense given that we aren't following and of the norms of packaging or shipping an SDK with this Python SDK so keeping the naming as loose as possible is probably better. Other projects tend to use `core` for their main functionality they intend to be consumed I have found.

docs/sdk/instructlab-sdk.md
booxter: Understood. I'd advise then to at least have several interview sessions with potential consumers, collecting requests / feedback / pain points.

booxter:
Agreed with @JamesKunstle on NOT adopting object oriented primitives to describe process flow (ilab methodology being a pipeline of processes with - hopefully pure - inputs and outputs). (As an anecdote: We have Java-style RAG implementation in tree with factories and all, which is very hard to follow and unravel due to all the layers of indirection, and I don't think this is a good pattern to continue in this code base, esp. for "SDK".)

---

@anastasds I wonder what makes something an SDK (SDK being Software Development Kit) if not a "pack of compute primitives meant to be consumed externally", which this proposal is (?) (Obviously, proper modeling is part of any good API and SDK design; but this seems wider.)

AFAIU a primary problem being solved here "current consumers of InstructLab are finding themselves importing different library private and public APIs", which is being solved by giving these consumers something to consume that is **meant** to be consumed. This is SDK.

anastasds:
@booxter in that general of a sense, every source codebase is an SDK.

An `sdk` package that contains all functions, classes, constants, etc meant to be used for custom development would reasonably be called an SDK. It would unarguably be an SDK if it were its own additional artifact that could be optionally downloaded by those intending to develop custom functionality, e.g. `pip install instructlab[sdk]`.

Having `*.api.*` or `*.core.*` submodules interspersed throughout is just Python APIs.

docs/sdk/instructlab-sdk.md
booxter: I don't think core should consume process management at all. Vice versa, CLI - through process management layer, if needed - will consume SDK. The user of the SDK should not be required to use our particular process management model. (Nor I think process management module is interesting to general consumers per se as a separate library. Prefer keeping it private to ilab CLI flow.)

cdoern:
@booxter @JamesKunstle given the consumption  patterns of _who_ is going to use this while also trying to the best of our ability to align with InstructLab's core values here is what I propose:

rather than a "spaghetti" structure like:

```python
from instructlab.core import Data

sdg_client = Data.Generator(<args>)
sdg_client.generate()

```
I propose instead to at least do

```python
from instructlab.core import Data, Config


config_object = Config.init(auto_detect=True)
data_client = Data(taxonomy_path="", cfg=config_object)
data_client.generate_data()
```

I think this structure at a minimum is what we need for the following reasons:

1. "parent" classes align well with the ethos of the command line structure. a Data class can be instantiated with common values meant to be inherited by other commands so args are not necessary each time
2. current consumers expect a hierarchical structure of InstructLab, doing something like `from instructlab.core import generate_data`
3. adding a level of abstraction allows the codebase to meld better with this new "core" package. Alternatively moving key functions into a core package at a top level would make the codebase confusing to understand as opposed to being grouped under a class like `Data`

In general usability, searchability (is that a word?), and our inheritance structure would suffer from not having at least top level cmd classes. 

@nathan-weinberg do you agree with the above?

docs/sdk/instructlab-sdk.md
nathan-weinberg: Yup, that makes sense and I had a convo with @cdoern that confirms!

docs/sdk/instructlab-sdk.md
nathan-weinberg: I think the CLI will always be more "opinionated" than the SDK - in terms of functionality, we're exposing the same functionality to both "consumers" but the nature of the different interfaces will likely always allow for more flexibility with the SDK (since it's easier to orchestrate Python classes than CLI commands)

docs/sdk/instructlab-sdk.md
nathan-weinberg: `core` is my preference, personally

docs/sdk/instructlab-sdk.md
nathan-weinberg: @cdoern talked about this in-person - basically this decision boils down to whether or not we need process management or not for the SDK

docs/sdk/instructlab-sdk.md
nathan-weinberg: Is this a departure from that? I don't see how

docs/sdk/instructlab-sdk.md
nathan-weinberg: Yes, I've come around on `core` :smile:

docs/sdk/instructlab-sdk.md
nathan-weinberg: Hmmm it's an interesting point @JamesKunstle @booxter - but isn't the difference here that the CLI and the SDK are consuming the same logic but exposing it in different ways, i.e. via the command line versus importable classes? I suppose the CLI could utilize the SDK classes, but at that point it seems to me that the logic itself could simply expose importable classes and that would cover the usecase? 

Indeed the process management piece shapes a lot of this conversation

nathan-weinberg:
> @booxter @JamesKunstle given the consumption patterns of _who_ is going to use this while also trying to the best of our ability to align with InstructLab's core values here is what I propose:
> 
> rather than a "spaghetti" structure like:
> 
> ```python
> from instructlab.core import Data
> 
> sdg_client = Data.Generator(<args>)
> sdg_client.generate()
> ```
> 
> I propose instead to at least do
> 
> ```python
> from instructlab.core import Data, Config
> 
> 
> config_object = Config.init(auto_detect=True)
> data_client = Data(taxonomy_path="", cfg=config_object)
> data_client.generate_data()
> ```
> 
> I think this structure at a minimum is what we need for the following reasons:
> 
>     1. "parent" classes align well with the ethos of the command line structure. a Data class can be instantiated with common values meant to be inherited by other commands so args are not necessary each time
> 
>     2. current consumers expect a hierarchical structure of InstructLab, rather than doing something like `from instructlab.core import generate_data`
> 
>     3. adding a level of abstraction allows the codebase to meld better with this new "core" package. Alternatively moving key functions into a core package at a top level would make the codebase confusing to understand as opposed to being grouped under a class like `Data`
> 
> 
> In general usability, searchability (is that a word?), and our inheritance structure would suffer from not having at least top level cmd classes.
> 
> @nathan-weinberg do you agree with the above?

Indeed - I think from a usability perspective it makes sense that the SDK classes would align with the CLI command groups - ultimately we're exposing the same (if not more) functionality through a different interface, but keeping the naming conventions of our original interface (the CLI) is a good thing IMO.

I also echo what @JamesKunstle says above about having sub-classes like `trainers` and `generators` etc. etc. - it's overengineering in my opinion, though I could be convinced otherwise given a realistic example.

booxter:
+1. Keeping the existing click command group hierarchy reflected in "SDK" while avoiding popping up classes is a good balance. Represent entities with objects and processes with methods / functions (Not with a "factory" to execute processes or to produce objects).

---

On the question of whether something is SDK or not - the main point is the intent. If the intent is for external consumers to take it and integrate with it, then it's SDK. Whether it's shipped in the same artifact is an implementation detail.

That said, as long as the name for this "thing" is not just "API" (which will be ambiguous for upcoming REST API), but e.g. "Python API" then I am ok with renaming too.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,153 @@
...
+
+### Structure
+
+This SDK should live in a net new package inside of `instructlab/instructlab` preferably to limit unnecessary imports in a new repository. The SDK could be imported as `instructlab.core...`
+
+The user surface initially should look like this:
+
+`instructlab.core` contains all SDK definitions. Users can `from instructlab.core import...` to use specific SDK classes
+
+For most of the existing InstructLab command groups, there should be a class:
```
booxter, line 58: The actual groups are

```
Commands:
  config    Command group for interacting with the config of InstructLab.
  data      Command group for interacting with the data generated by...
  model     Command group for interacting with the models in InstructLab.
  process   Command group for interacting with the processes run by...
  rag       Command group for interacting with the RAG for InstructLab.
  system    Command group for all system-related command calls.
  taxonomy  Command group for interacting with the taxonomy of InstructLab.
```
I don't think Eval (process) makes sense a class. What makes sense is Model.eval() though.

Note each click command group is tied to a "thing" (object) (except RAG? Which perhaps should've moved under `data`?), not a verb.

---

This is probably just an oversight though since I don't see Eval listed in the examples below, only in the line 60.

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,153 @@
...
+
+if diff:
+    # not in v1alpha1
+    data_client = Data(data_path="",..., teacher_model="", num_cpus="", taxonomy_path="",)
+    data_path = data_client.ingest()
+    # not in v1alpha1
+
+    openai_compat_client = some_server()
+
+    data_jsonls = data_client.generate_data(client=openai_compat_client, data=data_path)
```
booxter, line 106: I wonder how you distinguish between parameters passed into verb methods (`client`, `data` here) and parameters passed to `class.__init__` (line 100).

docs/sdk/instructlab-sdk.md
```diff
@@ -0,0 +1,153 @@
...
+3. `src/instructlab/process.py`
+4. `src/instructlab/core/data/generate.py`
+
+So generally: cli -> process management package to kick off a sub-process -> internal handling package -> core SDK (public definitions) -> library code, is the flow.
+
+The flow of the CLI today is such that the cli package for a command (`src/instructlab/cli/data/generate.py`) parses the command line options, manages creating a sub-process, and passes control to the core code (`/src/instructlab/core/data/generate.py`). This then calls out to the library APIs
+
+The internal handling package is necessary as it allows us to split off a sub-process when it makes the most sense for us before calling the library code directly. This is how the CLI works today.
+
+The difference with an SDK is that we would eventually want to end up executing `core/data/generator.py`, the actual publicly consumable python SDK. This will ensure that the CLI can do whatever custom handling it needs to do on top, but eventually it must boil down to the `core` package which uses publicly available methods from the various libraries.
```
booxter, line 142: Good goal. Eventually we can programmatically enforce that no imports for libraries are happening except under the sdk package.

booxter (approved):
This is correctly directed. We can flush out details (verbiage etc.) during development phase. What I'd encourage is to always refer to this as **Python** API/SDK going forward.

docs/sdk/instructlab-sdk.md
booxter: As far as I understood so far, process management is applicable where we anticipate long running processes (generate, train etc.) Some commands are like that and will go through the chain of calls as:

```
click -> (cli handler) - ...schedules process... -> process management -> ...spawns process... -> (process handler) -> sdk -> libraries
```

while for processes that don't need background execution, it would be:


```
click -> (cli handler) -> sdk -> libraries
```

docs/sdk/instructlab-sdk.md
cdoern: @bbrowning yep, the code is pretty much illustrative. The structure of classes I think is what I am focusing on the most. The actual arguments, functions within the `Data` class, etc can be ironed out later on. If we were to figure that out now, it would take forever!

docs/sdk/instructlab-sdk.md
cdoern: this is outdated, the new structure aligns with this comment

docs/sdk/instructlab-sdk.md
cdoern: yes, eval was a typo, it should live under model.


RAG is a new cmd group and yeah it should ideally live somewhere else but for simplicity it should probably have its own definition since it does in the CLI

docs/sdk/instructlab-sdk.md
cdoern: the SDK should be separate from the process logic, having bespoke state and process tracking through and SDK is out of the ordinary because a user might want to coordinate the processes themselves in a script. The current state of the document represents how I believe the flow should work

docs/sdk/instructlab-sdk.md
cdoern: new update (sorry), Taxonomy is going to get its own class after speaking with folks in the PR here and offline

docs/sdk/instructlab-sdk.md
cdoern: we can leave chat out for now

docs/sdk/instructlab-sdk.md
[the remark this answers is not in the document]
cdoern: structurally, this would be difficult. I have updated the document with a new structure I think fits both the current shape of the CLI and a usable scheme for an SDK

docs/sdk/instructlab-sdk.md
cdoern: I would probably do something like:

```python
def generate_data(self, data=None...):
data = data if data is not None else self.data
...

```

something like "hey, use the class defaults if the value we are given is None. Also, I think we can hash things like this out during development as its hard to figure out these patterns until writing the classes I find

nathan-weinberg (approved):

booxter (approved):

jwm4 (approved):
This seems like a fine start to me and I am fine for this to merge as is and drive the initial alpha API.  However, I think there is a big gap here because the proposed API doesn't include something that let's you send in a query and get back a response (both with and without RAG).  This would fill the role of `ilab model chat` but in an API-user friendly way.  Hopefully that can be added in a fast follow.

mairin (approved):
Here is further development needed:
- As @jwm4 pointed out, there's no calls for chat / rag chat
- As @booxter pointed out, we should refer to it as "Python API/SDK"

Question:

If I was working on an API/SDK for say SDG instead of something like

`instructlab.core.Data.generate_data`

Where the namespace is 

`instructlab.core.*`

Would the namespace be

`instructlab.data.*`

eg
`
instructlab.data.generate_data`?

I am going to approve this because it seems we've gotten a lot of reviewers, thoughtful comments, edits and updates and the basics seem like what is needed for a good start. I would like to see the two points above addressed in a subsequent draft.

cdoern:
thanks @mairin!

a few notes on the feedback:

1. I initially had `chat` as an endpoint but both William and Oleg wanted to table that for now since users might initially opt to just use chat completion requests manually from their server: https://github.com/instructlab/dev-docs/pull/184#discussion_r1929249682
2. Yeah, I can do a follow up switching all instances of SDK to Python SDK.

as for the question:

I think if I am understanding correctly, if users want to use the SDG library python API directly, they would instead use `instructlab-sdg.utils.taxonomy_to_leaf_nodes` for example. the library APIs are meant to be the smaller pieces of what something like `generate_data` would orchestrate. This fits the overall design shift some of the libraries are going for: They want to deal with functionality only pertaining to the inputs/outputs of their library rather than orchestration. Since today, generate_data is basically an orchestration of multiple SDG functions, I think core (`instructlab/instructlab`) would be the only owner of something like that


Do you want me to open the draft before merging this or can we do that as a follow up?

jwm4:
> since users might initially opt to just use chat completion requests manually from their server

FWIW, we're planning to also wrap the RAG pipeline in a typical chat completion API too (for a variety of reasons, e.g., ease of integration with the existing eval code), so this might also fill that need.  That might become part of this SDK, but it seems debatable since it is using a pre-existing API.  Something to consider for round 2 of this work....

PR #186, opened 2025-01-29
Custom Exceptions Patterns

anastasds:


docs/adr-custom-exceptions.md
```diff
@@ -0,0 +1,23 @@
+# InstructLab Custom Exceptions Pattern
+
+## Context
+
+The historically pervasive pattern for exception handling during application runtime is to catch internally raised exceptions in the CLI layer and use `click.secho` followed by `click.exceptions.Exit` to display a useful error message to the user before exiting the application. This leaves a risk of intermediate calls between the site of the exception and the user-facing layer changing, leading to missed new exceptions and outdated caught exceptions. A second issue is that of discoverability and verification: given a `click.exceptions.Exit` exception handling, it is not clear from the code where the caught exception originates from in the call stack, and, similarly, given a piece of code that can raise an exception, it is not clear from the local code whether that exception is properly handled in the CLI layer without investigation.
+
+These issues will compound whenever REST APIs begin development.
```
jwm4, line 7: and/or whenever the [SDK APIs](https://github.com/instructlab/dev-docs/pull/184) begin development.

docs/adr-custom-exceptions.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Decision
+
+InstructLab will adopt custom exceptions to handle specific faults which are then caught in the CLI layer.
```
jwm4, line 11: And separately caught in the REST layer I assume, since the handling would be different there.  For the SDK layer, I guess we want to just pass them along?  Maybe that decision is out of scope for this document.

jwm4 (approved):
This seems like a sensible idea.

docs/adr-custom-exceptions.md
nathan-weinberg: I would say they should be caught by whatever the "caller" of the logic is - whether it's the CLI, a REST API, another Python application importing `instructlab`, etc.

nathan-weinberg (approved):
Generally I agree with this, worth calling out that some of this work is ongoing in the Core repo as part of https://github.com/instructlab/instructlab/issues/2325

docs/adr-custom-exceptions.md
```diff
@@ -0,0 +1,23 @@
...
+
+Accepted
+
+## Consequences
+
+* Verification of exhaustive error handling will be clear within a code section that can raise.
+* CLI layer error handling will be easy to understand and trace.
+* Whenever REST APIs are developed, HTTP error codes should be easier to be associated with specific exceptions.
+* It should be easier to compose useful error messages for the user.
+* It should be easier to correctly scope exception handling (consider a `URLError` raised about SSL verification, for example, versus a custom `SSlVerificationException`).
```
booxter, line 23: I'm with you in general but not for this specific example (as I expressed in the patch), for the reason that we cannot and should not enumerate all the possible ways a request may fail, so letting URLError bubble up is fine here. (Caught further up the call stack and transformed into ilab specific exception as needed.)

PR #190, opened 2025-02-12
ci: Automated Y-Stream Releases

courtneypacheco:
This dev-doc describes a design we can utilize to automate y-stream release processes for each library within the `instructlab` GitHub org.

Z-stream releases can use the base logic defined in this dev-doc, but will require additional logic to handle backporting, manual rebasing, etc.

### DIFF
--- a/docs/ci/ci-automated-release-workflow.md
+++ b/docs/ci/ci-automated-release-workflow.md
@@ -1,0 +1,134 @@
+# Creating Automated Releases: Design Document
+
+## Motivation & Overview
+
+Presently, the release processes for every library within this `instructlab` GitHub organization is entirely manual:
+
+For example, for a typical y-stream release, a maintainer has to:
+
+1. Manually create a new release branch -- e.g., `release-0.y.0`,
+2. Manually create a pull request against `release-0.y.0` to cap the versions on some of the dependencies defined in their library's `requirements.txt` file,
+3. Optionally trigger an E2E test against that pull request,
+4. Wait for all pull request CI checks to complete,
+5. Manually request two maintainers to approve the pull request, and
+6. Manually create a release from the GitHub UI using that new branch
+
+This entire process takes at least 10 minutes of manual work, plus however long it takes for the pull request's checks to complete. (In some repositories, like the core repo, this can take 2+ hours.)
+
+Going forward, we should automate these release processes so that contributors and maintainers can focus more on development work and less on creating actual releases.
+
+## Generic Automation Workflow: Major Releases, Minor Releases, and Z-Stream Releases
+
+### Brief Overview of the Automation
+
+The automation logic described in this dev-doc will be published in the form of an in-house GitHub action called `create-automated-release` and it will therefore be callable from any workflow file. Thus, it is strongly recommended that each repository maintainer creates a `.github/workflows/automated-release.yml` workflow file to call this `create-automated-release` GitHub action from.
+
+### Goals of the Automation
+
+The automated process should be:
+
+1. Configurable so that any library maintainer can configure the automation to meet their specific repository's needs
+2. Schedulable so that releases are generated according to a specific cadence
+
+Scheduled releases are generally important for setting expectations around release cadences, but they are by no means required for every library and may not even be applicable to some in this GitHub organization.
+
+## Y-Stream Automation Workflow
+
+### Overview of Y-Stream (Minor) Release Automation
+
+Y-stream (minor) releases have historically been handled differently from Z-stream releases. Z-stream releases oftentimes involve backports for bugfixes and may require manual code rebasing to get those backports merged into the appropriate existing release branch. Therefore, we can think of Y-stream release logic as the "basis" for Z-stream release logic, which takes the Y-stream logic and builds upon it to account for backporting and other desirable actions.
+
+### Configurable Components
+
+As mentioned above, there are configurable components within this release process automation. The diagram in the next section references two configurable components:
+
+#### Trigger Schedule
+
+The trigger schedule defines the day and time (in UTC) when the release process will kick off.  This schedule can be disabled if desired, and maintainers can trigger the release process manually when needed instead.
+
+As mentioned in the brief overview of the automation, this in-house `create-automated-release` action will be callable from any workflow file. Thus, each library maintainer who wants to create a scheduled release should first create a  `.github/workflows/automated-release.yml` workflow file in their repository and define a schedule for it. For example:
+
+```yaml
+name: Create Y-Stream Release
+
+on:
+  schedule:
+    - cron: '30 1 1,15 * *' # Triggers at 1:30am UTC every 2 weeks on the 1st and the 15th day of each month
+```
+
+#### Custom List of Dependencies to Cap
+
+With each release, some library maintainers may want to cap the version of certain dependencies within their `requirements.txt` file as well as specify the desired upper cap for each one.
+
+The list of dependencies to cap should be provided in a file called `automated-release-config.yml`. This configuration file may be expanded in the future to accommodate more configurations as needed. Example:
+
+```yaml
+y_stream_release:
+
+  dependency_caps:
+    enable: true # optional parameter. If set to false, then none of the dependencies in `requirements.txt` will be capped.
+    packages:
+      instructlab-sdg: "+0.1.0"
+      instructlab-eval: "+0.2.0"
+      instructlab-training: "+1.0.0"
+```
+
+The keys specify the dependencies to cap. (In this case, we only have three: `instructlab-sdg`, `instructlab-eval`, and `instructlab-training`. The other dependencies in the `requirements.txt` file will be ignored and left untouched.)
+
+The value for each key specifies the cap relative to the current lower bound. For example, let's say the current `requirements.txt` file looks like this:
+
+```bash
+instructlab-sdg>0.20.0
+instructlab-eval>0.1.0
+instructlab-training
+```
+
+In this case, the automated logic will create a pull request that updates the `requirements.txt` file like so (ignoring the inline explanations I provided as comments):
+
+```bash
+instructlab-sdg>0.20.0,<=0.21.0 # increment by 0.1.0 because of the '+0.1.0' in the `cap-deps.cfg` file
+instructlab-eval>0.1.0,<=0.3.0  # increment by 0.2.0 because of the '+0.2.0' in the `cap-deps.cfg` file
+instructlab-training            # do nothing because there was no lower bound set.
+```
+
+If `dependency_caps` is not defined or `enable` is set to `False` under the `dependency_caps` key, then none of the dependencies defined in `requirements.txt` will be capped. If a dependency to cap is defined in `dependency_caps` and that dependency doesn't exist, then it will be ignored.
+
+### Example Workflow File
+
+Below is an example workflow file used to call this in-house GitHub action:
+
+```yaml
+name: Create Y-Stream Release
+
+on:
+  # Run every Monday at 1AM UTC
+  schedule:
+    - cron: '0 7 * * 1 ' 
+
+  # Allow manual dispatch, too
+  workflow_dispatch:
+    inputs:
+      pr_or_branch:
+        description: 'pull request number or branch name'
+        required: true
+        default: 'main'
+
+jobs:
+  create-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "create-automated-release"
+        uses: instructlab/ci-actions/create-automated-release@v1
+        with:
+          release_config: ".github/automated-release-config.yml" # points to where the library's release config is located in its repository
+```
+
+### Y-Stream Release Flow Diagram
+
+![Automated workflow for creating new GitHub releases](../images/design-diagram-for-automated-releases.png)
+
+## Z-Stream Release Automation Workflow
+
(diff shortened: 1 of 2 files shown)


docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,120 @@
+# Creating Automated Releases: Design Document
+
+## Motivation & Overview
+
+Presently, the release processes for every library within this `instructlab` GitHub organization is entirely manual:
+
+A maintainer has to:
+
+1. Manually create a new release branch -- e.g., `release-0.y.0`,
```
danmcp, line 9: Is this document/automation only meant to cover the y stream releases?

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,120 @@
...
+### Brief Overview of the Automation
+
+The automation logic described in this dev-doc will be published in the form of an in-house GitHub action called `create-automated-release` and it will therefore be callable from any workflow file. Thus, it is strongly recommended that each repository maintainer creates a `.github/workflows/automated-release.yml` workflow file to call this `create-automated-release` GitHub action from.
+
+### Goals of the Automation
+
+The automated process should be:
+
+1. Configurable so that any library maintainer can configure the automation to meet their specific repository's needs
+2. Schedulable so that releases are generated according to a specific cadence
```
danmcp, line 31: No objection to this being an option but it's unclear (to me at least) whether we want to use this.

docs/ci/ci-automated-release-workflow.md
courtneypacheco: Yes, y-stream only. Sorry, I should clarify that!

At this point, I don't think it's worth automating z-stream releases. We can certainly consider automating them in the future, but they can be complicated due to the fact sometimes backporting and manual rebasing is needed.

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,120 @@
...
+on:
+  schedule:
+    - cron: '30 1 1,15 * *' # Triggers at 1:30am UTC every 2 weeks on the 1st and the 15th day of each month
+```
+
+#### Custom List of Dependencies to Cap
+
+With each release, some library maintainers may want to cap the version of certain dependencies within their `requirements.txt` file as well as specify the desired upper cap for each one.
+
+The list of dependencies to cap should be provided in a file called `automated-release-config.yml`. This configuration file may be expanded in the future to accommodate more configurations as needed. Example:
```
kami619, line 57: @courtneypacheco Would this `automated-release-config.yml` file be per `y` stream release ? Is there a way we could add a checkpoint to verify in the beginning of the workflow to make sure this file exists and if it does, make sure it exists with the necessary details and format ? We could skip the secondary validation, if the file doesn't exist, which seems to be a possibility based on your workflow diagram.

kami619: Also would it be a good idea to version this file within the `y` stream release, if things change on us ? In other words, do we want this file to be immutable once the release process kicks off.

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,120 @@
...
+#### Trigger Schedule
+
+The trigger schedule defines the day and time (in UTC) when the release process will kick off.  This schedule can be disabled if desired, and maintainers can trigger the release process manually when needed instead.
+
+As mentioned in the brief overview of the automation, this in-house `create-automated-release` action will be callable from any workflow file. Thus, each library maintainer who wants to create a scheduled release should first create a  `.github/workflows/automated-release.yml` workflow file in their repository and define a schedule for it. For example:
+
+```yaml
+name: Create Y-Stream Release
+
+on:
```
kami619, line 48: Would we want to make this also to be available to run outside of this cron schedule for any reason, by exposing the inputs for the workflow ?

docs/ci/ci-automated-release-workflow.md
courtneypacheco: Sometimes maintainers want to create a new y-stream release automatically every X days. For example, the core repo, `instructlab/instructlab`, has a release cadence of 14 days. 

Realistically though, the scheduling portion of this design utilizes the built-in GitHub workflow trigger conditions and is therefore independent of the design itself. Thus, if a maintainer wants to release on a cadence, they can certainly define a schedule in their workflow file, but if they  don't want to do a schedule, then they can just omit the `schedule` trigger condition.

kami619:
@courtneypacheco Thank you for the PR, I think overall intent is on point and the workflow diagram really lays it out to understand it better. Just few comments, but I get those are nit picks.

docs/ci/ci-automated-release-workflow.md
courtneypacheco: Yes, as mentioned near the bottom of the document, there is an option to add your own trigger conditions, like `workflow_dispatch`. When using `workflow_dispatch`, users can manually kick off a release process.

The trigger conditions is defined at the Git workflow level, but not at the `create-automated-release` level.

docs/ci/ci-automated-release-workflow.md
courtneypacheco: I just updated the document to specify y-stream. Can you take a look?

ktdreyer:
Looks good to me.

This is not a blocker, just a thought. `enabled: true` looks superfluous to me - like if a developer wants to disable `dependency_caps`, they could specify an empty dict, or delete the `dependency_caps` section altogether. Or if you want to keep it that feature, maybe we could default `enabled: true` so users don't have to copy that line around to all repos.

Either way, great feature.

ktdreyer (approved):

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
+# Creating Automated Releases: Design Document
+
+## Motivation & Overview
+
+Presently, the release processes for every library within this `instructlab` GitHub organization is entirely manual:
+
+For example, for a typical y-stream release, a maintainer has to:
+
+1. Manually create a new release branch -- e.g., `release-0.y.0`,
```
nathan-weinberg, line 9:
```suggestion
1. Manually create a new release branch -- e.g., `release-0.y`
```

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+2. Manually create a pull request against `release-0.y.0` to cap the versions on some of the dependencies defined in their library's `requirements.txt` file,
```
nathan-weinberg, line 10:
```suggestion
2. Manually create a pull request against the `release-0.y` branch to cap the versions on some of the dependencies defined in their library's `requirements.txt` file
```

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+3. Optionally trigger an E2E test against that pull request,
+4. Wait for all pull request CI checks to complete,
+5. Manually request two maintainers to approve the pull request, and
+6. Manually create a release from the GitHub UI using that new branch
```
nathan-weinberg, line 14: I feel the commas and "and" aren't really necessary in a numbered list

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+1. Configurable so that any library maintainer can configure the automation to meet their specific repository's needs
+2. Schedulable so that releases are generated according to a specific cadence
+
+Scheduled releases are generally important for setting expectations around release cadences, but they are by no means required for every library and may not even be applicable to some in this GitHub organization.
+
+## Y-Stream Automation Workflow
+
+### Overview of Y-Stream (Minor) Release Automation
+
+Y-stream (minor) releases have historically been handled differently from Z-stream releases. Z-stream releases oftentimes involve backports for bugfixes and may require manual code rebasing to get those backports merged into the appropriate existing release branch. Therefore, we can think of Y-stream release logic as the "basis" for Z-stream release logic, which takes the Y-stream logic and builds upon it to account for backporting and other desirable actions.
```
nathan-weinberg, line 39:
```suggestion
Y-stream (minor) releases have historically been handled differently from Z-stream releases. Z-stream releases often times involve backports for bugfixes and may require manual code rebasing to get those backports merged into the appropriate existing release branch. Therefore, we can think of Y-stream release logic as the "basis" for Z-stream release logic, which takes the Y-stream logic and builds upon it to account for backporting and other desirable actions.
```

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+
+### Configurable Components
+
+As mentioned above, there are configurable components within this release process automation. The diagram in the next section references two configurable components:
```
nathan-weinberg, line 43:
```suggestion
As mentioned above, there are configurable components within this release process automation. The diagram in the next section references two configurable components
```

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+y_stream_release:
+
+  dependency_caps:
+    enable: true # optional parameter. If set to false, then none of the dependencies in `requirements.txt` will be capped.
+    packages:
+      instructlab-sdg: "+0.1.0"
+      instructlab-eval: "+0.2.0"
+      instructlab-training: "+1.0.0"
```
nathan-weinberg, line 73: Suggestion here, lmk what you think
```suggestion
    packages:
      - name: instructlab-sdg
        greater_than: 0.1.0
      - name: instructlab-eval
        greater_than: 0.2.0
      - name: instructlab-training
        greater_than: 1.0.0
```

You could have similar fields `equals` or `less_than` when applicable

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+name: Create Y-Stream Release
+
+on:
+  # Run every Monday at 1AM UTC
+  schedule:
+    - cron: '0 7 * * 1 ' 
+
+  # Allow manual dispatch, too
```
nathan-weinberg, line 108:
```suggestion
  # Allow manual dispatch too
```

docs/ci/ci-automated-release-workflow.md
booxter, line 10: Where is this requirement to cap coming from? Aren't we asked to [not](https://github.com/instructlab/dev-docs/blob/main/docs/dependency-management.md) cap?

```
Only apply "caps" to dependencies (using <) when that dependency has established a pattern of producing new releases with breaking changes.
```

Without this, steps 3-5 are no longer necessary.

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+
+```yaml
+name: Create Y-Stream Release
+
+on:
+  schedule:
+    - cron: '30 1 1,15 * *' # Triggers at 1:30am UTC every 2 weeks on the 1st and the 15th day of each month
+```
+
+#### Custom List of Dependencies to Cap
```
booxter, line 59: See above. Capping is not something we should automate. It's an exceptional situation, not a matter of course.

booxter (changes requested):
I have my reservations about automation of capping. We should just stop capping.

docs/ci/ci-automated-release-workflow.md
courtneypacheco: We have been capping certain dependencies in the core repo since October 2024: https://github.com/instructlab/instructlab/pulls?q=is%3Apr+%22deps%3A+cap%22+is%3Aclosed+

To elaborate, we have been capping our own InstructLab library dependencies to ensure that if we create a new release from the core repo, that new release won't automatically consume potential breaking changes from one of our own libraries.

courtneypacheco: Also, to be clear, the "automatic capping" feature is not required to be used by anybody. Maintainers can ignore the feature if desired. And since the automation described here will create a pull request that someone has to manually approve and review, the dependency cappping changes not be merged automatically. 😃

docs/ci/ci-automated-release-workflow.md
courtneypacheco: Good catch on the typo! Will update!

docs/ci/ci-automated-release-workflow.md
courtneypacheco: Thanks for catching that again will update.

docs/ci/ci-automated-release-workflow.md
```diff
@@ -0,0 +1,134 @@
...
+instructlab-sdg>0.20.0
+instructlab-eval>0.1.0
+instructlab-training
+```
+
+In this case, the automated logic will create a pull request that updates the `requirements.txt` file like so (ignoring the inline explanations I provided as comments):
+
+```bash
+instructlab-sdg>0.20.0,<=0.21.0 # increment by 0.1.0 because of the '+0.1.0' in the `cap-deps.cfg` file
+instructlab-eval>0.1.0,<=0.3.0  # increment by 0.2.0 because of the '+0.2.0' in the `cap-deps.cfg` file
```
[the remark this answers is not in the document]
courtneypacheco: Yes it was. Thanks for catching! Will remove that reference. 👍

PR #194, opened 2025-03-25
Create a proposal for a `notebooks` repository

alimaredia:
See `docs/notebooks/notebooks-repo.md` in this pull request for more details.


cdoern:
I think a decent counterproposal to this is that each component would house its own notebooks given that it's likely they will become upstream projects in their own right housed under the instructlab org. WDYT about that?

RobotSail (approved):
LGTM

kelbrown20 (commented):
I had a similar comment with what @cdoern was saying. I think this makes sense for this repo to contain the combined workflow notebooks. Where the stages would go in the library repos.

alimaredia:
I've taken the feedback on this PR and removed the `stages` directory. The reason I included that was so that users had a centralized place to discover all of the libraries notebooks but there are other ways to do this, like on the community's website or on google colab that accomplish the same goal.

cdoern (approved):

nimbinatus:
A bit of bikeshedding on my part, but perhaps `examples` is better than `notebooks` so it can be more of a catchall for any and all types of examples, like a one-stop shop? Notebooks are one type of artifact, but what about if we have code examples once the SDK from https://github.com/instructlab/dev-docs/blob/main/docs/sdk/instructlab-sdk.md is built? I would think it would be good to have all of these things in one place for every user to use.

alimaredia:
@nimbinatus great idea. I changed the repo name and dev doc a bit for flexibility sake.

PR #198, opened 2025-05-14
Document pip constraints CI enforcement for all dependencies

booxter:
This work is being implemented in the main (instructlab/instructlab)
repository. The same approach should be applied across all the repos.

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

### DIFF
--- a/docs/ci/dependency-constraints.md
+++ b/docs/ci/dependency-constraints.md
@@ -1,0 +1,66 @@
+# Dependency Constraints in InstructLab CI
+
+This document describes how Python dependencies should be managed in
+InstructLab CI.
+
+## Goals
+
+1. Ensure that the CI environment is consistent and reproducible.
+2. Ensure that new dependency releases do not break the CI environment.
+3. Ensure that we adopt new dependency releases in the CI environment in a
+   timely manner.
+
+## Approach
+
+### Pin all versions with constraints files
+
+Each repository should contain a [pip
+constraints](https://pip.pypa.io/en/stable/user_guide/#constraints-files) file
+that lists the pinned versions of all dependencies used in the CI environment.
+
+In case a repository supports multiple platforms (`linux` vs `darwin`) or
+accelerators (`cuda`, `cpu`, `hpu`), multiple constraints files may be
+generated.
+
+The constraints files should be used in all CI jobs that install Python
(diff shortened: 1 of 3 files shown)


docs/ci-dependency-constraints.md
```diff
@@ -0,0 +1,54 @@
...
+re-generate all the constraints files in the project.
+
+Updates to these files should be generated automatically by the CI system and
+should not be modified manually (subject to rare exceptions). A periodic
+CI job should be added to ensure this happens on a regular basis (at least once
+a week).
+
+The job should run `tox -e constraints` and post the result as a PR for review.
+The PR will be validated by all the relevant CI jobs. Project core team is
+expected to review these PRs in a timely manner. The team should make sure that
```
courtneypacheco, line 41: I think we should define "timely manner" because every contributor has their own idea of what "timely manner" means. For some, that means under 1 day. For others, that could mean a few weeks.

I say we put a number -- for example, "review the PR within 3 days of creation". That doesn't necessarily mean we _merge_ the PR within 3 days, but it means we at least acknowledge it and look at it within 3 days to decide if it's appropriate to merge.

Btw, I'm not set on any specific amount of days. I just gave 3 days as an example.

booxter: 3 business days seem fair for a team of people. I made it business days.

docs/ci/dependency-constraints.md
```diff
@@ -0,0 +1,55 @@
...
+
+This document describes how Python dependencies should be managed in
+InstructLab CI.
+
+## Goals
+
+1. Ensure that the CI environment is consistent and reproducible.
+2. Ensure that new upstream releases do not break the CI environment.
+3. Ensure that we adopt new upstream releases in the CI environment in a timely
+   manner.
```
ktdreyer, line 11: Let's clarify which upstream we're talking about here:
```suggestion
2. Ensure that new dependency releases do not break the CI environment.
3. Ensure that we adopt new dependency releases in the CI environment in a timely
   manner.
```

ktdreyer (approved):

booxter:
Updated a bit to reflect that we now have a new `ci-actions` action for periodic updates.

booxter:
I'm self-merging because the document is effectively implemented across repos and no longer in the decision making phase.

PR #201, opened 2025-05-15
Require launch-ec2-runner-with-fallback use for all ec2 runners

booxter:
Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>


ktdreyer (approved):

nathan-weinberg (approved):

booxter:
I do not think this is controversial - it's already used across the board, so this doc just captures the reality. I will assume lazy consensus here and merge it.

